PR #275, opened 2014-04-16
Permalink support for Git and HG based indexes

jamonation:
Ok this is **just** a demo but it gives an idea of what a permalink might look like for a Mercurial based project. There are a couple steps involved to make it work:

1. Edit dxr.config.in and add a line: "vcs_uri = http://hg.mozilla.org/mozilla-central"
2. Browse to mozilla-central and choose a file, say: http://hg.mozilla.org/mozilla-central/file/e7ab61c6eb1d/db/sqlite3/src/sqlite3.h
3. Grab the part of the url after 'file' e.g. /e7ab61c6eb1d/db/sqlite3/src/sqlite3.h
4. Now fire up DXR and visit: http://33.33.33.77:8000/mozilla-central/rev/e7ab61c6eb1d/db/sqlite3/src/sqlite3.h
5. Try http://33.33.33.77:8000/mozilla-central/rev/e7ab61c6eb1d/build/application.ini etc.

You should see a highlighted version of the file in question, with a working context menu and highlighting. Of course this is **not** working with indexed code and the search won't return, but you get the idea.

It's also really slow since it has to run pygment's guess_lexer, and on really long files it just hangs. I don't know what to do about that except perhaps try using hightlights.js instead, since that is client side and will save on DXR having to do a bunch of guess work tying up requests and giving the user a poor experience.

Any thoughts though? On the right track? Worth pursuing?


erikrose:
A version of this has since landed. Woot. :-)

PR #336, opened 2014-09-20
Alpha Dockerfile to create docker based dev container

jamonation:
This is an alpha Dockerfile that seems to work. I can run make test, and dxr-serve.py on a test_basic directory.

To run on Windows and OSX will take some extra steps that I'll document once this build runs successfully.

Here goes!


jamonation:
Closing this, work is in the es branch.

PR #367, opened 2014-11-25
Bug 1104302 - dxr-build fails when the git's origin has been cloned from an http(s) address 

rricard:
See bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1104302

> Note: let's discuss this PR, I'm sure there are better things to do than what I just did !


rricard:
FIrst, let's fix this one

rricard:
```
Stderr from the command:

stdin: is not a tty
+ apt-key add -
+ wget -qO - http://packages.elasticsearch.org/GPG-KEY-elasticsearch
+ echo deb http://packages.elasticsearch.org/elasticsearch/1.1/debian stable main
+ apt-get update
W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/trusty-updates/main/i18n/Translation-en  Hash Sum mismatch

E: Some index files failed to download. They have been ignored, or old ones used instead.
Pseudo-terminal will not be allocated because stdin is not a terminal.
make: Nothing to be done for `clean'.
make -C dxr/plugins/clang build
make[1]: Nothing to be done for `build'.
make[1]: Leaving directory `/home/vagrant/dxr/dxr/plugins/clang'
npm install
make[1]: llvm-config: Command not found
make[1]: Entering directory `/home/vagrant/dxr/dxr/plugins/clang'
make[1]: llvm-config: Command not found
make: npm: Command not found
make: *** [node] Error 127
[default] Forcing shutdown of VM...
[default] Destroying VM and associated drives...
[default] Running cleanup tasks for 'shell' provisioner...
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

erikrose:
Wow! My guess is it's a transient network failure. Wait a few minutes and try again?

rricard:
I can't retrigger the build by myself

erikrose:
Oh, sorry. I didn't scroll all the way up and assumed this way on your local machine. :-)

dxr/plugins/omniglot.py
```diff
@@ -192,12 +192,16 @@ def synth_web_url(self, repo):
...
+        elif repo.startswith(("git://github.com/", "https://github.com/")):
             self._is_github = True
             if repo.endswith(".git"):
                 repo = repo[:-len(".git")]
-            return "https" + repo[len("git"):]
-        raise Exception("I don't know what's going on")
+            if repo.startswith("git:"):
+                repo = "https" + repo[len("git"):]
+            return repo
+        else:
```
erikrose: The fix looks good to me. And I see where you were going with the `else` here; it's very noble. :-) But the rest of omniglot seems to assume something useful will be in `self.upstream`, and, if we just make up a URL, it won't be. So let's keep raising an exception in that case, with a message apologizing that we only support GitHub-derived repos right now. Ultimately, we should provide a config value so you can specify a URL manually, but first I want to refactor the config system.

rricard: Ok, no problem, I just found that an onboarding with an error saying: "I don't know what's going on" was a bit weird. Maybe we could change it with an error saying something like this: "This remote is not supported for now, please use a remote Github repository. You are welcome to fix it, see bug XXX."

erikrose: Sounds like a perfect approach.

erikrose:
Good first bug! Thank you! :-D

PR #369, opened 2014-11-29
Bug 938222 - Show hash or other VCS identifier of the indexed source

rricard:
See the discussion on bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=938222


rricard:
Still working on it ... Been greatly delayed by the holiday season and some other side projects ! Happy new year to all of you by the way !

rricard:
Ok, really back this time ! I think I finally worked out what I had to do on Fennec and now comes DXR !

rricard:
Ok, so @erikrose I'm still quite confused here on the approach to take ... Do we make a general hash for every possible things (hash, authors, ...) or should we just add fields in the plugin API ?

dxr/static/templates/folder.html
```diff
@@ -31,14 +31,18 @@
         <th scope="col">Name</th>
         <th scope="col">Modified</th>
         <th scope="col">Size</th>
+        <th scope="col">Author</th>
```
erikrose: There's no need to add these columns here. Author isn't that important, and the VCS hash will be the same for every file under most VCSs.

rricard: I was about to get rid of that anyway

erikrose:
Okay, how about this? Let's add a `revision_id()` method to `FileToSkim`. Then omniglot can implement it in its `FileToIndex` (which is a subclass of `FileToSkim`). Then, in build.py, we can take the first `revision_id()` that returns an nonempty string (which is what the default implementation would do) and plop that into the footer of the template.

Making `revision_id()` per-file rather than per-tree lets us support Perforce and CVS, whose gettable identifiers are per-file (P4 has tree identifiers, but they aren't preserved in checkouts). For git and hg, we can cache the revision in the TreeToIndex so we don't continually re-fetch it. Does that sound sane?

rricard:
That's fine to me ! Working on it right away !

rricard:
Ok, I'm testing this locally, I'll ping you when it's ready !

rricard:
@erikrose: I'm struggling with my vm right now and I have to stop for tonight. Getting back to it by Sunday I think. Thanks for your help today !

rricard:
One more thing: the current code is definitely wrong as it messes with es indexing and file generation. So it'll definitely need some more work !

erikrose:
Are you waiting for feedback, or is there more you'd like to do first?

rricard:
No thanks,  I still need some work on this

rricard:
@erikrose: found the problem after rebuilding the vm like 6 times: a missing indent was changing the whole logic and preventing the building of the index ... I rebuild and test it again. I'll ping you if it works as expected ! :)

rricard:
With little luck, the CI will be green for once ! :)

rricard:
@erikrose So I finally managed to make it work locally ! :) You can review this but I don't know why those tests are still failing though.

erikrose:
It's not normal, but it's started happening for everyone yesterday. It's not your fault.

PR #370, opened 2014-12-02
Image preview (Bug 1052217)

pelmers:
Show some preview images for image files, and list binary files in folder views.
Not merge-ready yet, since right now it just copies image files to another directory and serves from there.
Instead it should put them in ES (separate index from the code?) and query that on request.
See Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1052217

edit: now things do go into ES


dxr/app.py
```diff
@@ -162,6 +164,30 @@ def _tree_tuples(trees, tree, query_text, is_case_sensitive):
              description)
             for t, description in trees.iteritems()]
 
+@dxr_blueprint.route('/<tree>/images/<path:path>')
+def send_image(tree, path):
+    '''
```
erikrose: Just a nit, but we'll obviously want to stay consistent with the docstring quoting in the rest of the project.

dxr/build.py
```diff
@@ -539,8 +538,26 @@ def index_file(tree, tree_indexers, path, es, index, jinja_env):
...
+    '''
+    Index the image located at path relative to given tree using es in index.
+    Return the URL to get the image.
+    '''
+    # define the image source
+    image_src = "%s/images/%s" % (tree.name, rel_path)
+    # encode the image file as base64 string and index it
+    with open(join(tree.source_folder, rel_path), "rb") as img:
+        imgdata = img.read().encode("base64")
+        es.index(index, IMAGE, {
```
erikrose: You've got the right idea here, but I think it would be more consistent with the existing ES schema if we put the image in a field of FILE rather than introducing a new IMAGE doctype. We'd then tweak app.browse() to pull only the fields it needs when building dir listings (so it doesn't pull down huge images all the time). This would also simplify a future feature: thumbnails. It would become trivial to store a thumbnail-sized image in a FILE field as well, and that could be fetched in the same request as the rest of the file data in browse().

dxr/plugins/core.py
```diff
@@ -81,6 +81,22 @@
...
+        '_all': {
+            'enabled': False
+        },
+        'properties': {
+            # IMAGE filters query this (like FILE)
+            'path': PATH_MAPPING,
+            # actual image data here
+            'data': {
+                'type': 'binary',
+                'index': 'no'
```
erikrose: Very good. :-)

dxr/static/js/dxr.js
```diff
@@ -461,7 +461,7 @@ $(function() {
     // Thanks to bug 63040 in Chrome, onpopstate is fired when the page reloads.
     // That means that if we naively set onpopstate, we would get into an
     // infinite loop of reloading whenever onpopstate is triggered. Therefore,
-    // we have to only add out onpopstate handler once the page has loaded.
+    // we have to only add our onpopstate handler once the page has loaded.
```
erikrose: Thanks for the typo fixes!

dxr/build.py
erikrose: Best of all, we wouldn't need this special-purpose stuff in build.py anymore; the image data would become merely a file-wide needle, returned from the core plugin's needles() method!

dxr/app.py
```diff
@@ -162,6 +164,30 @@ def _tree_tuples(trees, tree, query_text, is_case_sensitive):
              description)
             for t, description in trees.iteritems()]
 
+@dxr_blueprint.route('/<tree>/images/<path:path>')
+def send_image(tree, path):
```
erikrose: I'd like to see this merged into browse() somehow. We shouldn't need a separate kind of URL for images (or other things we can display). Just think: what would actually visiting the image at its browse()-style URL do otherwise?

dxr/static/js/image_preview.js
```diff
@@ -0,0 +1,19 @@
...
+    'use strict';
+
+    /**
+     * Replace 'source' with 'images' in href, and set that to the background-image
+     */
+    function setBackgroundImageFromLink(anchorElement) {
+        var href = anchorElement.getAttribute('href');
+        // note: breaks if the tree's name is "source"
+        var bg_src = href.replace('source', 'images');
+        anchorElement.style.backgroundImage = 'url(' + bg_src + ')';
```
erikrose: The entire, full-size image? What if it's humongous?

erikrose:
I pulled it down and tried it. Looks really good! To finish up, let's…
* reduce the default max thumbnail size to 20K,
* fix the lack of margins or whatever is causing the image_file template to render the image over the breadcrumbs and to kick the "generated by" footer over to the right

screen shot 2015-01-09 at 13 02 26

docs/source/configuration.rst
```diff
@@ -100,6 +100,10 @@ Here are the options that can live in the ``[DXR]`` section:
     will be substituted, and their meanings are as in ``es_index``. Default:
     ``dxr_{format}_{tree}``.
 
+``icon_size_cutoff``
```
erikrose: Let's call this `max_thumbnail_size`.

dxr/app.py
```diff
@@ -13,7 +15,7 @@
 
 from dxr.build import linked_pathname
 from dxr.exceptions import BadTerm
-from dxr.filters import FILE
+from dxr.filters import FILE, IMAGE
```
erikrose: We don't need IMAGE to exist anymore.

dxr/app.py
```diff
@@ -162,6 +164,27 @@ def _tree_tuples(trees, tree, query_text, is_case_sensitive):
              description)
             for t, description in trees.iteritems()]
 
+@dxr_blueprint.route('/<tree>/raw/<path:path>')
+def send_raw(tree, path):
```
erikrose: Let's just call this `raw`. The rest of the routes are pretty much just named after their URL segments.

dxr/app.py
```diff
@@ -162,6 +164,27 @@ def _tree_tuples(trees, tree, query_text, is_case_sensitive):
...
+    query = {
+            'filter': {
+                'term': {'path': path}
+                }
+            }
+    index = current_app.config['ES_ALIASES'][tree]
+    results = current_app.es.search(
+            query,
+            index=index,
+            doc_type=FILE, size=1)
```
erikrose: When we go with one arg per line, let's do it with all of them.

dxr/app.py
```diff
@@ -162,6 +164,27 @@ def _tree_tuples(trees, tree, query_text, is_case_sensitive):
              description)
             for t, description in trees.iteritems()]
 
+@dxr_blueprint.route('/<tree>/raw/<path:path>')
+def send_raw(tree, path):
+    """Send raw data at path from tree, for binary things like images."""
+    query = {
+            'filter': {
```
erikrose: We can outdent all these lines so the final } is aligned with "query". See some of the other JSON-ish places in query.py for examples. (Now I'm nitpicking, so you know your PR is in good shape! :-))

dxr/app.py
```diff
@@ -162,6 +164,27 @@ def _tree_tuples(trees, tree, query_text, is_case_sensitive):
...
+            }
+    index = current_app.config['ES_ALIASES'][tree]
+    results = current_app.es.search(
+            query,
+            index=index,
+            doc_type=FILE, size=1)
+    try:
+        # we explicitly get index 0 because there should be exactly 1 result
+        data = results['hits']['hits'][0]['_source']['raw_data'][0]
+        dataFile = StringIO(data.decode("base64"))
```
erikrose: Nits! :-) `dataFile` should be `data_file` for conformance with PEP 8's naming conventions. And let's go with single quotes, as we typically do when we're not trying to enclose other single quotes.

More substantially, we should keep the body of `try` statements minimal so we don't catch errors we weren't expecting. So let's move the second two lines of this one to after the `except`. You can put it in an `else` clause if you like, or just leave that out, since your `except` exits the function anyway.

dxr/app.py
```diff
@@ -162,6 +164,27 @@ def _tree_tuples(trees, tree, query_text, is_case_sensitive):
...
+        return send_file(dataFile, mimetype=guess_type(path)[0])
+    except IndexError: # couldn't find the image
+        raise IndexError("Couldn't find image for %s.\nSearch: %s\nResults: %s"
```
erikrose: The end user shouldn't be able to get DXR to throw exceptions by passing in bad data. So let's make this a 404 instead.

dxr/app.py
```diff
@@ -263,10 +286,20 @@ def parallel(tree, path=''):
 
 
 def _tree_folder(tree):
-    """Return the on-disk path to the root of the given tree's folder in the
-    instance."""
+    """Return the on-disk path to the root of the given tree's folder in
+    the instance."""
     return join(current_app.instance_path, 'trees', tree)
 
+def _icon_class_name(file_doc):
```
erikrose: Good factoring-out

dxr/build.py
```diff
@@ -507,40 +506,50 @@ def index_file(tree, tree_indexers, path, es, index, jinja_env):
...
+            'name': os.path.basename(rel_path),
+        }
+
+        if is_image(rel_path):
+            _fill_and_write_template(
+                jinja_env,
+                'image_file.html',
+                join(tree.target_folder, rel_path + '.html'),
+                merge(common, file_vars, {
+                'image_src': raw_url(tree.name, rel_path) }))
```
erikrose: Ugh, I can't wait for request-time rendering so we can stop introducing our own URL-building machinery.

dxr/build.py
```diff
@@ -507,40 +506,50 @@ def index_file(tree, tree_indexers, path, es, index, jinja_env):
...
+
+        else:
```
erikrose: You didn't put the `is_text` test back in here. I'd prefer not to show binary files until we indicate they are binary, so you don't bother clicking into them only to find zero information.

erikrose: Greying them out in directory listings might be one approach. (We shouldn't do that now; it'd just hold up merging this useful stuff in.)

dxr/build.py
```diff
@@ -651,6 +660,9 @@ def build_sections(links):
     # Return list of section and items (without importance)
     return [(section, list(items)) for importance, section, items in links]
 
+def raw_url(treename, path):
```
erikrose: Might as well put this in utils.py with the rest of the URL builders

dxr/plugins/core.py
```diff
@@ -306,6 +311,8 @@ def needles(self):
         extension = splitext(self.path)[1]
         if extension:
             yield 'ext', extension[1:]  # skip the period
+        if is_image(self.path):
```
erikrose: This should be `if self.contains_text() and is_image(self.path)`. Otherwise, we could end up base64ing a unicode string, and that might land us in weird places.

dxr/build.py
```diff
@@ -485,17 +485,16 @@ def index_file(tree, tree_indexers, path, es, index, jinja_env):
...
-                 # Hard-code the keys that are hard-coded in the browse()
-                 # controller. Merge with the pluggable ones from needles:
-                 dict(folder=folder_name,
-                      name=file_name,
-                      size=file_info.st_size,
-                      modified=datetime.fromtimestamp(file_info.st_mtime),
-                      is_folder=False,
-                      **needles))
+    es.index(index,
+                FILE,
```
erikrose: Somehow the indentation got messed up here.

dxr/plugins/core.py
erikrose: I feel like the "is it an image" test should be factored up to create a single point of truth—sometimes we test the file extension, other times the type of `file_to_index.contents` (which is currently decided by a technically incorrect but remarkable reliable heuristic: does the file contain \0?). I'm tempted to add it to FileToIndex, but that's kind of ugly—which plugin's FileToIndex would win if they disagreed? Would we just spec it to be unoverridable? Ick. Needs more thought. Do you have any good ideas?

dxr/plugins/core.py
```diff
@@ -313,6 +320,13 @@ def needles_by_line(self):
             yield [('number', number),
                    ('content', text)]
 
+    def is_interesting(self):
+        """
+        Core is responsible for putting files in the index,
+        so everything is interesting
+        """
+        return True
```
erikrose: Since we don't want to show all binary files yet, only images and text files are interesting.

dxr/static/js/image_preview.js
```diff
@@ -0,0 +1,19 @@
+$(function() {
```
erikrose: Why not merge this into dxr.js and save a request? (Someday we'll minify our JS and this won't matter.)

dxr/static/templates/image_file.html
```diff
@@ -0,0 +1,38 @@
...
+        <span class="path-separator">/</span><a href="{{ wwwroot }}{{ path }}">{{ name }}</a>
+      {%- else -%}
+        <a href="{{ wwwroot }}{{ path }}">{{ name }}</a>
+      {%- endif -%}
+    {%- endfor -%}
+  </div>
+
+  {{ tree_menu(tree_tuples, tree) }}
+
+  <img src="{{wwwroot}}/{{ image_src }}">
```
erikrose: inconsistent spacing

dxr/build.py
pelmers: I'd been basing this off of [bug 1052217](https://bugzilla.mozilla.org/show_bug.cgi?id=1052217), where you mentioned it would be not bad to acknowledge existence of other binary files, so I removed the is_text check here. The template already has is_text check and shows (binary file) if the file is not text. Actually there is a test for this at [test_binary_files.py](http://git.io/mJA1BA).

erikrose: Yes, that "(binary file)" thing was just a stub, meant to be replaced by, for example, your nice image display. It's a bad UI pattern to dangle something useful in front of a user (e.g. a file icon) but then say "Nope, there's nothing here" when they drill in. So I'd want to at least expose a clue in the dir listing before we start showing all binary files.=

dxr/plugins/core.py
pelmers: Doing a contains_text() check will filter out .svg images since they are just XML files (right now they are previewed just like binary image types).
Currently the "is it an image" check is always done through is_image, which queries the extension.

erikrose: Good point!

dxr/static/templates/image_file.html
pelmers: turns out I'm missing the clossing </div> from line 17.

erikrose:
Ah, just saw this!

dxr/app.py
```diff
@@ -165,6 +167,29 @@ def _tree_tuples(trees, tree, query_text, is_case_sensitive):
...
+            query,
+            index=index,
+            doc_type=FILE,
+            size=1)
+    try:
+        # we explicitly get index 0 because there should be exactly 1 result
+        data = results['hits']['hits'][0]['_source']['raw_data'][0]
+    except IndexError: # couldn't find the image
+        raise NotFound
+    data_file = StringIO(data.decode('base64'))
```
erikrose: Might as well use cStringIO for speed.

dxr/app.py
```diff
@@ -165,6 +167,29 @@ def _tree_tuples(trees, tree, query_text, is_case_sensitive):
...
+        'filter': {
+            'term': {
+                'path': path
+            }
+        }
+    }
+    index = current_app.config['ES_ALIASES'][tree]
+    results = current_app.es.search(
+            query,
+            index=index,
```
erikrose: I hate to say it, but now that we've made FILE docs potentially large, we should avoid needlessly bringing the image over the wire in browse(), likely using http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-fields.html or http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-source-filtering.html.

dxr/build.py
```diff
@@ -342,7 +342,7 @@ def ensure_folder(folder, clean=False):
     if clean and os.path.isdir(folder):
         shutil.rmtree(folder, False)
     if not os.path.isdir(folder):
-        mkdir(folder)
+        os.makedirs(folder)
```
erikrose: Nice catch! DId you actually run into this?

dxr/build.py
```diff
@@ -487,17 +487,18 @@ def index_file(tree, tree_indexers, path, es, index, jinja_env):
...
-                 FILE,
-                 # Hard-code the keys that are hard-coded in the browse()
-                 # controller. Merge with the pluggable ones from needles:
-                 dict(folder=folder_name,
-                      name=file_name,
-                      size=file_info.st_size,
-                      modified=datetime.fromtimestamp(file_info.st_mtime),
-                      is_folder=False,
-                      **needles))
+                FILE,
```
erikrose: Oops, indent got damaged.

dxr/static/js/dxr.js
```diff
@@ -476,4 +476,21 @@ $(function() {
...
+     */
+    function setBackgroundImageFromLink(anchorElement) {
+        var href = anchorElement.getAttribute('href');
+        // note: breaks if the tree's name is "source"
+        var bg_src = href.replace('source', 'raw');
+        anchorElement.style.backgroundImage = 'url(' + bg_src + ')';
+    }
+
+    window.addEventListener('load', function() {
+        $(".image").not('.too_fat').each(function() {
```
erikrose: Hmm, what would happen if we later added some other image to the page, like a logo up in the header?

dxr/static/templates/image_file.html
```diff
@@ -0,0 +1,40 @@
...
+
+{% block site_css %}
+  {{ super() }}
+  <link href="{{ wwwroot }}/static/css/tree-selector.css" rel="stylesheet" type="text/css" media="screen" />
+  <link href="{{ wwwroot }}/static/css/image_file.css" rel="stylesheet" type="text/css" media="screen" />
+{% endblock %}
+
+{% block content %}
+<div id="content" class="content" data-no-results="{{ no_results() }}">
+  <div class="breadcrumbs">
```
erikrose: I notice this is repeated from file.html. Do you think we could factor it up into a macro?

erikrose:
This has turned into a real epic bug. But it is a major feature, so that's not unexpected. Thanks for sticking with it! :-D

dxr/app.py
pelmers: Actually this is cStringIO: `from cStringIO import StringIO` on line 6

dxr/app.py
pelmers: Sure, looking at that second link it seems something like `"_source": { "exclude": ["raw_data"] }` would be sufficient. I'll test it out and see what works.

dxr/build.py
pelmers: I think I ran into this back when I was copying binary files/images to a separate target directory. It might not be necessary anymore.

dxr/build.py
pelmers: Sharp catch!

dxr/static/js/dxr.js
pelmers: We probably wouldn't add the "image" class to a logo image, since that would give it an image mimetype icon as the background. Perhaps it's clearer to rename ".image" to something like ".preview_image".

dxr/static/templates/image_file.html
pelmers: Yeah, that's definitely a good idea. I wasn't very familiar with Jinja templating at the time, but I've learned more about it working with it on some other projects recently.

dxr/app.py
erikrose: Haha, I'm an idiot.

dxr/build.py
erikrose: The GitHub diff view makes it obvious. :-) I often catch unintended changes by doing a local git diff before I push. (And then I hook it up to FileMerge for a nice side-by-side view. GitX is pretty good, too.)=

dxr/static/js/dxr.js
erikrose: My fault; I saw ".image" and thought "img". No need to change anything.

dxr/static/templates/image_file.html
erikrose: It's a great piece of kit.

erikrose:
In response to your IRC'd question, you don't need to squash. You can if you want, but I figure it's just an opportunity to introduce errors. Definitely don't rebase it on the tip of es or anything; then you have no recourse if you make a mistake resolving conflicts.

erikrose:
Otherwise, I'll merge it in a little bit.

erikrose:
A logical next step might be to dig into indexing for all binary files, like you started on initially. https://bugzilla.mozilla.org/show_bug.cgi?id=1052217#c6

erikrose:
Great feature! Thanks so much for all your hard work on it!

PR #373, opened 2014-12-11
Only show and use filters from enabled plugins

nick29581:


nick29581:
@erikrose ping - any update here?

erikrose:
Sorry, trying to get the ES branch polished for staging. I'll have a look tomorrow!

nick29581:
@erikrose ping - changes made, see last commit

erikrose:
Having a look now.

dxr/query.py
```diff
@@ -309,17 +320,51 @@ def generic_visit(self, node, visited_children):
         """
         return visited_children or node
 
+def filters_per_plugin(enabled_plugins):
```
erikrose: It's not clear from the docstrings what this and the next 2 functions do. Trying to figure it out....

dxr/query.py
```diff
@@ -309,17 +320,51 @@ def generic_visit(self, node, visited_children):
...
+    """Return a list of filters for each plugin in enabled_plugins"""
+    # A list of plugins for all enabled plugins
+    plugins = [p for (n, p) in all_plugins().iteritems() if n in enabled_plugins]
+    # A list of filters (with a non-null description) for each enabled plugin 
+    filters = [[f for f in p.filters if f.description] for p in plugins]
```
erikrose: Testing for "if f.description" here breaks plain-text searching.

dxr/query.py
```diff
@@ -179,52 +172,70 @@ def direct_result(self):
...
-        # regexp, function, etc. No filter is a prefix of a later one. This
-        # avoids premature matches.
-        '|'.join(sorted((re.escape(filter_name) for
-                                filter_name, filters in
-                                FILTERS_NAMED.iteritems() if
-                                filters[0].description),
-                        key=len,
-                        reverse=True)) + ur'''"
+query_grammar_cache = {}
+@cached(query_grammar_cache)
```
erikrose: Much better. I did another round of simplification so you no longer have to pass in the cache from outside as well.

nick29581:
updated with (hopefully) better docstrings and moved the valid description check to fix the plain text search bug

erikrose:
What part of "merged" don't you understand, GitHub?

PR #379, opened 2014-12-29
Fix #280: Add help icon to header with links to Github and docs.

### LINKED ISSUE #280: Link to DXR project info in HTML output
I really like it when projects contain a link to the source code or a "getting involved" page in their output. This both lets people know they can scratch itches and also tells them how to do so. AFAICT dxr.mozilla.org doesn't have such an advertisement. I think we should put one in.

Osmose:
Just an idea I had while looking for easy bugs to mess around with. Adds a help icon to the header that, when hovered over, shows a popup below the header with a small paragraph describing DXR and two links to the Github page and the docs.

screen shot 2014-12-28 at 11 02 37 pm

It uses CSS for the hover and has some padding around the box to allow for inaccurate mousing around the box.

I wasn't able to test how this looks / works with the navigation box as I couldn't find any tests that had files with the navigation box, nor could I figure out how to get it to appear myself. It shouldn't get in the way but I'm not 100% sure about that.

Lemme know what you think!


erikrose:
I love it. I've been meaning to do something like that. I haven't played with it yet, but Kneejerk Me wonders if we could make it consistent with the Filters menu by having it clickable rather than hoverable. That would also dodge any accidental opens if someone were to brush over it on the way to, say, the nav panel. What do you think?

erikrose:
Incidentally, I've also been meaning to expose recent release notes somewhere within the app. This might be a good place. :-)

Osmose:
Could do it on click, would just require some JavaScript (on hover allowed me to cheat by using CSS only). Will look at adding that sometime today or this weekend.

erikrose:
Cool. You can take a look at tree-selector.js, near "contentContainer.on('click'", to see how we do the other menus.=

Osmose:
Updated. The help box now appears when you click the help icon, and disappears when you click outside of the box or the icon.

dxr/static/js/dxr.js
```diff
@@ -428,6 +428,22 @@ $(function() {
     }
 
 
+    // Show the help box when the help icon is clicked, and hide it when
+    // anything outside of the box is clicked.
+    var helpIcon = $('.help-icon'),
+        helpMsg = helpIcon.find('.help-msg');
+
+    helpIcon.on('click', function() {
+        helpMsg.fadeIn(300);
```
erikrose: Let's not fade it in; we don't do that anywhere else (for the perception of speed).

erikrose:
And if the help "menu" is still disembodied (I have yet to init a new vagrant box locally to try it out), can we "attach" it to the help icon or something?

Osmose:
Updated:
screen shot 2015-01-15 at 10 14 38 am

erikrose:
It's a beaut!

erikrose:
Couple things I notice on actually trying it on my machine:
1. The icon is showing up as a "broken char" glyph.
2. Clicking the icon a second time doesn't close the menu. That's the convention the Filters menu follows, borrowed from desktop environments.

erikrose:
nm about number 1; a super-reload fixed the glyph.

erikrose:
And there are some visual ickies in Chrome:
screen shot 2015-01-29 at 17 51 03

Osmose:
Updated.

erikrose:
Pretty doggone good! I like that the icon depresses slightly on click—nice touch.

erikrose:
I do notice it breaks when using text-only zooming, whereas the rest of the UI is pretty okay. That's not necessarily a blocker, but if it's an easy fix, hey.

dxr/static/templates/layout.html
```diff
@@ -20,54 +20,65 @@
   </head>
   <body>
 
-    <form method="get" action="{{ wwwroot }}/{{ tree }}/search" id="basic_search" class="search-box">
```
erikrose: Thanks for getting rid of that thing where there was a search-box class *and* a search-box ID. That was confusing.

erikrose:
A good next bug might be https://bugzilla.mozilla.org/show_bug.cgi?id=965666 or https://bugzilla.mozilla.org/show_bug.cgi?id=965647, as long as you have your head in the menu routines. Or literally anything else you're interested in. :-) Say, do you know anything about JS test harnesses? We're been meaning to do that, and a lot of the design work is already done: https://bugzilla.mozilla.org/show_bug.cgi?id=937568. There's even a js-tests branch in the main repo.

erikrose:
I hope you don't mind; I took the liberty of standardizing the look of the menus across all 3. You inspired me to give it some thought, and I added drop shadows for all of them, which makes the limits of the menus easier to see, especially over a page that has a lot of ink on it, like a directory listing. So glad you opened this!

erikrose:
Another idea I've been playing with is to expose the last release note or two someplace in the app. The Help menu might be a good place. For example, if I fix the quotes parser, the help menu could offer, “You can put single quotes around things now: 'say "what"?”

PR #380, opened 2015-01-05
Avoid installing development-only dependencies during "npm install".

abbeyj:


erikrose:
AFAIK, we have yet to confirm that this installs everything we need—that is, to try it starting from a blank node_modules folder. (See https://bugzilla.mozilla.org/show_bug.cgi?id=1118889.)

erikrose:
We've pared down the npm dependencies a LOT since this was opened. The only remaining dependencies is nunjucks, currently listed under devDependencies. It's a questionable fit there: it's more of a build dependency than a development one, but we don't commit the built artifacts, so it's needed even in production—at least on the box that builds the deployed artifact. Please reopen if you can still think of a way to improve things, but I think the spirit of this is fixed by the lighterweight dependencies and by the new, smarter makefile that doesn't `npm install` unless things actually changed.

PR #381, opened 2015-01-05
Fix tests under clang 3.5

abbeyj:


erikrose:
If the tests pass, looks good to me. r+

PR #382, opened 2015-01-06
969775 add margin to bottom of search results

rachelhathaway:


erikrose:
Hmm. Infinite scrolling seems to be broken right now. Ouch. Once we get that fixed, I can review your patch. :-(

erikrose:
It occurs to me I could try this on master while the es branch is hung up on infinite scroll being broken...

dxr/static/css/dxr.css
```diff
@@ -332,7 +332,7 @@ mark {
     transform: rotate3d(0, 0, 1, 270deg);
 }
 .content {
-    margin-top: 55px;
+    margin: 55px 0 25px;
```
erikrose: Did you really mean to change the right and left margins to 0? That's what happened here. (See http://www.htmldog.com/reference/cssproperties/margin/ and the meaning of 3-value margin arguments.)

rachelhathaway: Yes, I meant to do that. Left and right margins were already 0 due to the body's margins being explicitly set to 0. If you'd rather it not be explicit on this element, I'd be happy to add the margin-top declaration back in and change my added line to just be margin-bottom.

PR #383, opened 2015-01-14
Bugfix3

anmoljagetia:
Updated so that the checkbox matches the current value of variable isCaseSensitive.


erikrose:
`isCaseSensitive` isn't in scope there, so it always contains `undefined`. (It's unfortunate that JS fails so quietly, but we're stuck with it.) I think you might benefit from reading up on JavaScript and doing a few smaller projects, and then come back to something complex like DXR.

PR #384, opened 2015-01-18
Fixes Bug 1119862, get infinite scroll to work on ES branch

ghost:
@erikrose This fixes the infinite scroll problem but, now w have the resultHead repeating, which is probably a bug all it's own.

screen shot 2015-01-18 at 18 03 35

### DIFF
--- a/dxr/static/js/dxr.js
+++ b/dxr/static/js/dxr.js
@@ -264,7 +264,7 @@ $(function() {
                         var state = {};
 
                         // Update result count
-                        resultCount = data.results.length;
+                        resultCount = data.results[0].lines.length;
                         // Use the results.html partial so we do not inject the entire container again.
                         populateResults('partial/results.html', data, true);
                         // update URL with new offset
@@ -330,7 +330,7 @@ $(function() {
         } else {
 
             var results = data.results;
-            resultCount = results.length;
+            resultCount = results[0].lines.length;
 
             for (var result in results) {
                 var icon = results[result].icon;


ossreleasefeed:
Or perhaps the header repeating is a good thing, as it gives a marker of where the next set of results started. Might just look like a bug because all of these results are from a single file.

ossreleasefeed:
Oh wait, was just thinking about this. Does each object inside results, contain the results from a single file? i.e. if results were found in more than one file, there will be more than 1 object, containing lines with matches, inside the returned data.results.

ossreleasefeed:
If so, then my fix here will not solve the problem and just happen to solve it here, because all the results was in one file.

ossreleasefeed:
If my comment before is correct, that means the data.results.length will indicate the number of files that contain results but previousDataLimit actually refers to the maximum number of lines that should be returned.

So, we will need a counter that indicated the total number of lines across all files in data.results and use that count for resultCount

erikrose:
Sorry I've let this fester for so long! Let me have a look at the server side and remind myself what's going on.

erikrose:
> Does each object inside results, contain the results from a single file?

Yes. This snippet from app.py, which creates the `results` key of the object sent over the network, says it best:

```python
        results = [{'icon': icon,
                    'path': path,
                    'lines': [{'line_number': nb, 'line': l} for nb, l in lines]}
                   for icon, path, lines in query.results(offset, limit)]
```

So there's an `{icon: …, path: …, lines: …}` object for each file. Those may be split in half, resulting in more than 1 object for each file, if a file spans the fetch-limit boundary.

erikrose:
I am open to whichever approach makes the cleanest code:

1. We can either continue using the current results format, in which lines are grouped by file. In this case, we'll have to make a note in the DOM of the last file fetched and, if the first one of the next fetch is the same, merge the two under the same header.
2. Alternately, we could just change to a line-based transfer format, with no nesting. It would be bigger, but the gzipping of the response should knock it back down.

schalkneethling:
Thanks for the feedback @erikrose

erikrose:
Fixed by #410.

PR #386, opened 2015-02-07
Bug 1052217: index binary files

pelmers:
https://bugzilla.mozilla.org/show_bug.cgi?id=1052217

Now index all binary files, not just images. Arbitrary binary files are shown in faded italics on the browse view and are not clickable.
I add an is_binary needle to the core plugin which is True when the file is binary but not an image that we can preview (and left out otherwise). The browse template then uses the flag to determine whether to fade out a line and invalidates its hrefs.


erikrose:
Just wanted to let you know I haven't forgotten about this. I'm hung up fixing some ES stuff but shall return.

dxr/static/templates/folder.html
```diff
@@ -33,11 +33,11 @@
...
-      {% for icon, name, modified, size, relative_url in files_and_folders %}
-        <tr>
-          <td><a href="{{ relative_url }}" class="icon {{ icon }}">{{ name }}</a></td>
-          <td><a href="{{ relative_url }}">{{ '' if modified is none else modified.strftime("%Y %b %d %H:%m") }}</a></td>
-          <td><a href="{{ relative_url }}">{{ '' if size is none else size|filesizeformat }}</a></td>
+      {% for icon, name, modified, size, relative_url, is_binary in files_and_folders %}
+        <tr class={{ "binary_row" if is_binary else "" }}>
+          <td><a href="{{ relative_url if not is_binary else "#" }}" class="icon {{ icon }}">{{ name }}</a></td>
+          <td><a href="{{ relative_url if not is_binary else "#" }}">{{ '' if modified is none else modified.strftime("%Y %b %d %H:%m") }}</a></td>
+          <td><a href="{{ relative_url if not is_binary else "#" }}">{{ '' if size is none else size|filesizeformat }}</a></td>
```
erikrose: Wouldn't it be better to not have these be links at all?

erikrose:
I think we're going to need to ship some default `ignore_patterns` with this so, for example, `.DS_Store` doesn't show up on Macs. :-)

dxr/build.py
```diff
@@ -487,18 +487,16 @@ def index_file(tree, tree_indexers, path, es, index, jinja_env):
...
-                     size=file_info.st_size,
-                     modified=datetime.fromtimestamp(file_info.st_mtime),
-                     is_folder=False,
-                     **needles))
+    es.index(index,
+             FILE,
+             # Hard-code the keys that are hard-coded in the browse()
+             # controller. Merge with the pluggable ones from needles:
+             dict(folder=folder_name,
+                 name=file_name,
```
erikrose: indentation

erikrose:
Before I get too engrosed in picking nits, let me say I like it. I really like it. :-) I keep on getting requests for this; it's going to make a lot of people happy.

dxr/static/css/dxr.css
```diff
@@ -148,6 +148,10 @@ td#line-numbers span {
     text-align: right;
     padding: 0 0.5rem;
 }
+tr.binary_row {
+    font-style: italic;
```
erikrose: I'm leaning toward losing the italics; I think the opacity (which I love) and removing the hover state (which we should do) are hints enough.

dxr/build.py
```diff
@@ -487,18 +487,16 @@ def index_file(tree, tree_indexers, path, es, index, jinja_env):
...
-                 FILE,
-                 # Hard-code the keys that are hard-coded in the browse()
-                 # controller. Merge with the pluggable ones from needles:
-                 dict(folder=folder_name,
-                     name=file_name,
-                     size=file_info.st_size,
-                     modified=datetime.fromtimestamp(file_info.st_mtime),
-                     is_folder=False,
-                     **needles))
+    es.index(index,
```
erikrose: Btw, you might want to rebase; I rewrote this section yesterday.

erikrose:
Other than the things I mentioned, it looks good. Have you tried fooling it on any real-world project? Showing all this stuff makes bad calls by our (crazy-simple, likely naive) is_text() routine much more visible.

pelmers:
Hey Erik, thanks for the comments. I've rebased and removed italics and the <a> on binary file rows. I haven't tested it on a real-world project yet, any suggestions for something that has a lot of binary files?

erikrose:
Good question! Maybe something with a GUI or something with a test suite that involves sound?

pelmers:
At some point I completely forgot that I was still working on this... now it's rebased so I'll bump.

I agree we should add ignore_patterns -- shall this be exposed in dxr.config? Perhaps by default we ignore all hidden files.
Interestingly, the null byte check is exactly what GNU diff does (was it your inspiration?): http://www.gnu.org/software/diffutils/manual/html_mono/diff.html#Binary

erikrose:
Not sure who introduced the null-byte check; it was there when I started working on the project.

dxr/app.py
```diff
@@ -360,7 +361,6 @@ def _linked_pathname(path, tree_name):
 
     return components
 
-
```
erikrose: Let's keep the double line skip here. PEP 8 says 2 blank lines between top-level constructs.

erikrose:
Looks good! All that remains before merging is to add some sensible defaults to the ignore directives, like maybe…

`*~ *.o *.out .nfs* CVS .DS_Store ._* *.pyc .git .svn .hg`

erikrose:
Ah, I see many of them are there already. Let's just add `.DS_Store .nfs* *~ ._*` and call it done.

dxr/plugins/core.py
```diff
@@ -381,6 +381,9 @@ def needles(self):
             bytestring = (self.contents.encode('utf-8') if self.contains_text()
                           else self.contents)
             yield 'raw_data', b64encode(bytestring)
+        # binary, but not an image
+        elif not self.contains_text():
+            yield 'is_binary', True
```
erikrose: We should add this to the mapping, above in core.py.

pelmers: Added^

PR #387, opened 2015-02-11
Add a test directory for the python plugin.

Osmose:
Adds a test directory for manual testing of the python plugin. Automated tests may come later, but at least this will make it easy to test improvements manually.

The Readme was copied from test_basic, I'm not sure if it's useful as a skipped file for indexing but whatever.

### DIFF
--- a/tests/test_python/code/README.mkd
+++ b/tests/test_python/code/README.mkd
@@ -1,0 +1,12 @@
+Hello World
+===========
+_The most annoying sources in the world._
+
+These sources are designed to be as annoying as possible,
+this achieved by doing stuff such as having really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really long lines :)
+Or I would include UTF-8 charaters, to see if something crashes, that would be really annoying as I think I've already fixed that bug...
+Well here we go... I'll just write something in Danish, then we'll have a little easter egg too :)
+Hej Verden,
+Håber godt nok aldrig jeg kommer til rette den der bug med UTF-8 fejl, for skal jeg være helt ærligt, så er det lavet så det virker...
+Jeg ved faktisk ikke rigtigt om det er korrekt lavet, mest af alt fordi jeg ikke finder encodings særligt spændende.
+Når, hvis jeg da så lige skal være et rigtigt røvhul kan jeg slutte denne tekst af med uden et linjeskift!!! Det er godt nok træls
--- a/tests/test_python/code/code.py
+++ b/tests/test_python/code/code.py
@@ -1,0 +1,41 @@
+from collections import namedtuple
+
+
+Point = namedtuple('Point', 'x y')
+
+
+class Graph(object):
+    def __init__(self, width, height):
+        self.width = width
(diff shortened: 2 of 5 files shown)


Osmose:
@erikrose r?

erikrose:
1. This is more of a demo or an example than a test (since there are no tests), so it doesn't belong in the tests dir. I'd be open to including some examples with the project, maybe some things that check out a real-world project and run analysis against them. In fact, I've been meaning to add an appendix to the docs wherein we lay out how to configure DXR again mozilla-central.
 2. If it did belong with the tests, language-specific tests should be inside language-specific plugins, like in plugins/python/tests.

So I think you should totally use this yourself to try things, but it's not necessary to merge it in.

Btw, did you know you can go into any folder-based test, `make` it, and fire it dxr-serve and look around?

Also, did you know you can turn any SingleFileTestCase into a folder-based test by adding `should_delete_instance = False` to its class temporarily?

Osmose:
> This is more of a demo or an example than a test (since there are no tests), so it doesn't belong in the tests dir. I'd be open to including some examples with the project, maybe some things that check out a real-world project and run analysis against them. In fact, I've been meaning to add an appendix to the docs wherein we lay out how to configure DXR again mozilla-central.

The problem I'm attempting to solve here is "I would like to work on the python plugin. In order to check out any changes I make, I need some sample data to run against." The use of `test_basic` in the docs as what you run for your dev instance of DXR implied to me that the test directories are for all sorts of testing, rather than just automated testing.

I guess I just feel weird leaving this big gap between "I want to work on X" and "I have a test instance of DXR running with my changes to X" in the case of the Python. But whatevs, _I_ at least have the test directory. :P

erikrose:
What you say is interesting, and I wish to subscribe to your newsletter. I have to admit I'm a little puzzled; maybe I'm inured and don't feel the pain anymore. Please help me understand.

For me, the first step in working on a new analysis is to make a new test for it: usually a fresh SingleFileTestCase. I don't have a "dev instance" of DXR as such: I have a bunch of test cases, any of which I can turn interactive at the flip of a switch if I need to poke around manually. In the docs, I seized upon `test_basic` only because it has more than one file in it and is already a folder, so you don't have to edit any code to make it interactive. I didn't mean to recommend it as something for constant use; it was just an expedient demo.

After reading that, do you still see a gap? If so, I think we should contrive a way to ship the vagrant box with some kind of example: maybe a script that pulls down redis and indexes it or something. Or (why not—might as well be useful, and it saves on downloads) we could have DXR index itself.

Osmose:
I still do. I think it's not uncommon to develop the way I do, which is to (at least in some cases) throw down code as a way to figure out the shape of what you're actually trying to make. Sometimes, as in the case of adding a new way to index code, it's not too hard to come up with a test case matching your final output, but often, I find it hard to reason about the interface that I'd be testing until I've at least written part of it.

It's like how development instances of Django sites will often have a command to generate some test data to populate your dev site with. It lets you check out if the site works and lets you play with a feature in-development without having to write an automated test for the feature you've got in mind. That's manual testing, and I feel like it's actually a good fit to use the tests directory for things like this. Also, eventually `test_python` would have some actual automated tests in it anyway.

erikrose:
Makes sense. I probably do that less because I'm used to the system now and pretty much know where I want to put things.

What if we put an "examples" folder at the root level and put manual-test thingies in there? I'm loath to mix big, thick, full-featured manual harnesses with automated tests, because it'll never be clear what you're allowed to edit without invalidating the tests. (Invalidations don't throw errors.) Once we get a good example in there, we can change the docs to stop referring to test_basic, plus we can move all the non-tested stuff out of test_basic. How's that strike you?

Osmose:
It makes everyone happy, let's do it!

Issue #388, opened 2015-02-21
Errors running tests on es branch


Osmose:
I'm running into some issues running the tests via `make test` on the `es` branch in Vagrant:

- With no modifications, I get this a bunch:

  ```
ERROR
Warning: $jobs is not used in build_command for 'clang'
Build command for 'code' failed, exited non-zero.
Log follows:
    | /bin/sh: 1: clang++: not found
  ```

  Which leads to a bunch of BuildErrors when the actual testcases are run.

- If I add a symlink from `clang++` to `clang++-3.5`, I get segfaults when attempting to build the code.

Any tips on how to debug this issue?

erikrose:
I'm glad you get that error locally. We started getting it on CI a few days ago, so at least it's reproducible on a dev machine. So much for LTS releases, eh? Were we testing against clang 3.5 before? If not, maybe drop down to whatever version we were. I suspect we were on 3.5, though. Can I see the full error?

Osmose:
Between the time I posted this and now I had already recreated my vagrant box, and now I only get 2 test failures and none of the issues I was seeing before.

Output of `make test` on my new box: https://gist.github.com/Osmose/de7609305c3f42e887f6

erikrose:
Hmm. You shouldn't get any test failures. I guess I'll go clear the Jenkins env and see if that helps things. We need CI working.

PR #389, opened 2015-02-21
Refactor Python Plugin

Osmose:
This adds 4 basic unit tests for the Python plugin. I feel like I should maybe be testing more but it's not immediately obvious to me what some other interesting test cases would be.

This includes the test directory from https://github.com/mozilla/dxr/pull/387 because I based it off of that, but I can rebase it to not include that if you want (my argument for keeping it is that it's a good temporary measure before making a full examples directory like we talked about, but I'm not particularly tied to it, just figured I'd check before rebasing).

This also includes a refactoring of the python plugin. Originally this PR was going to add a new `derived:` filter, which fit in with the refactoring, but I've since found that adding that filter will be harder than I thought, so I decided to submit what I have now for review. I think the code is more straightforward and easier to add to, but I'm also happy to remove that if it's considered out of scope or unnecessary.

As per #388 I'm having trouble running the entire test suite, but the new tests added seem to pass, so I figure this is good enough for at least some feedback and review.

### DIFF
--- a/dxr/plugins/python/indexers.py
+++ b/dxr/plugins/python/indexers.py
@@ -45,8 +45,8 @@ mappings = {
 
 class FileToIndex(FileToIndexBase):
     needle_types = {
-        'ClassDef': 'py_type',
-        'FunctionDef': 'py_function',
+        ast.ClassDef: 'py_type',
+        ast.FunctionDef: 'py_function',
     }
 
     def is_interesting(self):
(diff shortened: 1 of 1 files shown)


dxr/plugins/python/indexers.py
```diff
@@ -45,8 +45,8 @@
 
 class FileToIndex(FileToIndexBase):
     needle_types = {
-        'ClassDef': 'py_type',
-        'FunctionDef': 'py_function',
+        ast.ClassDef: 'py_type',
+        ast.FunctionDef: 'py_function',
```
erikrose: Hooray for this!

tests/test_python/code/README.mkd
```diff
@@ -0,0 +1,12 @@
+Hello World
```
erikrose: Is this one of the unnecessary files you were talking about? I agree: it's unnecessary (unless I'm missing something).

Osmose: I'm not quite sure which unnecessary files you're referring to, but this file is useful mainly for ensuring that only Python files are analyzed. On some of my work branching off of this I ran into SyntaxErrors because I wasn't properly filtering, and this helped me catch that during manual testing.

erikrose: Great. Let's replace the file contents with a message explaining that, then.

tests/test_python/test_python.py
```diff
@@ -0,0 +1,42 @@
+from textwrap import dedent
+
+from dxr.testing import SingleFileTestCase
+
+
+class PythonSingleFileTestCase(SingleFileTestCase):
+    """Test case suited to testing the Python plugin."""
+    enabled_plugins = ('pygmentize', 'python')
```
erikrose: It seems like the common pattern is to use lists for homogenous series like this and tuples for anonymous structs. What do you think?

tests/test_python/test_python.py
```diff
@@ -0,0 +1,42 @@
...
+class FunctionTests(PythonSingleFileTestCase):
+    source = dedent("""
+    def foo():
+        return 1
+
+    def   baz  (blargh):
+        return blargh
+    """)
+
+    def test_normal_class(self):
```
erikrose: normal_function, you mean?

dxr/testing.py
```diff
@@ -207,30 +208,45 @@ class SingleFileTestCase(TestCase):
     # print its path so you can examine it:
     should_delete_instance = True
 
+    # Override this in a subclass to change the plugins used.
+    enabled_plugins = ('pygmentize', 'clang')
+
+    # Override this in a subclass to change the build command.
+    # $CXX gets injected by the clang DXR plugin:
+    build_command = '$CXX -o main main.cpp'
```
erikrose: Seems like SingleFileTestCase should be an abstract class, and there should be concrete C and Python subclasses that fill these out. Does have to happen in this PR; could follow up later.

dxr/plugins/python/indexers.py
```diff
@@ -63,58 +63,43 @@ def needles_by_line(self):
...
 
-            cur_token = {'name': tok_name, 'start': start, 'end': end}
-            if relocate_nodes:
-                cur_token['nodes'] = relocate_nodes
-            else:
-                cur_token['nodes'] = node_location.get(start, [])
-
-            relocate_nodes = []
+            if tok_name in ('def', 'class'):
+                previous_start = start
```
erikrose: Interesting trick. :-)

dxr/plugins/python/indexers.py
```diff
@@ -63,58 +63,43 @@ def needles_by_line(self):
...
+            elif previous_start:
```
erikrose: Could this fail if previous_start ends up being 0, like if a "def" is the first thing in a file?

Osmose: `bool((0,0))` appears to be true, so I don't thing so, but that's grounds for using `is not None` to be explicit about it.

erikrose: Ah, I didn't realize those were tuple keys because I hadn't read far enough down yet.

tests/test_python/test_python.py
Osmose: I usually go with mutability to determine what to use, rather than the contents; since I don't ever expect to change this value I went with tuple. I'm not necessarily _trying_ to stop people from changing it, but it feels like something that shouldn't change.

The pattern you mention is new to me, and interesting, but I feel like mutability is an easier rule. Not sure. (I mean it's moot for this PR anyway so yay).

dxr/plugins/python/indexers.py
```diff
@@ -63,58 +63,43 @@ def needles_by_line(self):
...
-            )
+        # Run through the AST looking for things to index!
+        syntax_tree = ast.parse(self.contents.encode('utf-8'))
+        for node in ast.walk(syntax_tree):
+            if isinstance(node, ast.ClassDef) or isinstance(node, ast.FunctionDef):
+                node.start = (node.lineno, node.col_offset)
+                if node.start in node_start_table:
+                    node.start = node_start_table[node.start]
+
+                node.end = (node.start[0], node.start[1] + len(node.name))
```
erikrose: I take it tokens never span lines? What about a multi-line string?

erikrose: nm

erikrose:
If the tests pass (locally, since CI is busted), r+.

PR #390, opened 2015-02-25
Add bases: and derived: filters to the Python plugin.

Osmose:
Add a post_build handler that parses each Python file and stores the
class inheritance tree as well as import paths. The indexing step then
uses this information to figure out where the base classes and derived
classes of each class it finds live, and generates needles for each of
them.

Currently figure out module names using a new configuration value
called `python_path`. In the future we probably want to support multiple
paths or find a smarter way of figuring this out.


Osmose:
@erikrose r?

erikrose:
Following http://dxr.readthedocs.org/en/es/configuration.html#plugin-specific-options, the option should really be called plugin_python_python_path. Of course, that's all going to change and get a lot less ugly when this config rewrite lands.

dxr/plugins/python/indexers.py
```diff
@@ -38,19 +45,114 @@
...
         },
     },
 }
 
 
+class TreeToIndex(TreeToIndexBase):
+    def __init__(self, *args, **kwargs):
+        super(TreeToIndex, self).__init__(*args, **kwargs)
+
+        self.python_path = getattr(self.tree, 'python_path', self.tree.source_folder)
```
erikrose: Then you wouldn't have to edit this at all.

erikrose:
Maybe just `plugin_python_path` would be a good compromise for the time being.

Osmose:
Updated the option to `plugin_python_path` for the time being.

Osmose:
Rebased on the latest `es` with the config file changes, and updated to work with them and use a dict for the test config instead of a string.

dxr/plugins/python/__init__.py
```diff
@@ -1,4 +1,16 @@
...
-from .indexers import FileToIndex, mappings
-from .filters import TypeFilter, FunctionFilter
+from dxr.plugins import Plugin
+
+from .indexers import mappings, TreeToIndex
+from .filters import BasesFilter, DerivedFilter, FunctionFilter, TypeFilter
+
+
+plugin = Plugin(
+    filters=[BasesFilter, DerivedFilter, FunctionFilter, TypeFilter],
```
erikrose: Why not use `filters_from_namespace` here so you never have to update this again?

dxr/plugins/python/filters.py
```diff
@@ -15,3 +15,13 @@ class TypeFilter(_PyFilter):
...
+
+
+class DerivedFilter(_PyFilter):
+    name = 'derived'
+    description = Markup("Subclasses of a class: <code>derived:SomeSuperclass</code>")
+
+
+class BasesFilter(_PyFilter):
+    name = 'bases'
+    description = Markup("Superclasses of a class: <code>derived:SomeSubclass</code>")
```
erikrose: "derived" → "bases"

dxr/plugins/python/indexers.py
```diff
@@ -35,19 +42,115 @@
...
     },
 }
 
 
+class TreeToIndex(TreeToIndexBase):
+    def __init__(self, *args, **kwargs):
+        super(TreeToIndex, self).__init__(*args, **kwargs)
+
+        self.python_path = getattr(self.tree, 'plugin_python_path', self.tree.source_folder)
+        self.class_bases = {}
```
erikrose: Maybe call it `base_classes` for symmetry with `derived_classes`.

dxr/plugins/python/indexers.py
```diff
@@ -35,19 +42,115 @@
...
+        self.derived_classes = defaultdict(list)
+        self.names = {}
+
+    @property
+    def unignored_files(self):
```
erikrose: I like this. I like this so much, it's making me think that perhaps build.unignored_files() should become a method on TreeConfig. What do you think?

Osmose: Yes please!

dxr/plugins/python/indexers.py
```diff
@@ -35,19 +42,115 @@
...
+                base_name = normalize_name(self.names, base_name)
+                self.derived_classes[base_name].append(class_name)
+
+    def analyze(self, path):
+        """Parse a Python file and analyze it for import and class data
+        that we need to find the inheritance tree for classes later on.
+
+        """
+        module_path = path_to_module(self.python_path, path)
+        syntax_tree = ast.parse(file_contents(path, 'utf-8'))
```
erikrose: We can use `self.tree.source_encoding` instead of 'utf-8'. (Someday, we can let any encoding declaration in the Python file itself override this.)

dxr/plugins/python/indexers.py
```diff
@@ -35,19 +42,115 @@
...
+                         self.tree.ignore_filenames)
+
+    def post_build(self):
+        for path in self.unignored_files:
+            if is_interesting(path):
+                self.analyze(path)
+
+        # Once we've got all the base classes, store the opposite
+        # direction: derived classes!
+        for class_name, bases in self.class_bases.items():
```
erikrose: Might as well use iteritems()

dxr/plugins/python/indexers.py
```diff
@@ -35,19 +42,115 @@
...
+            for base_name in bases:
+                base_name = normalize_name(self.names, base_name)
+                self.derived_classes[base_name].append(class_name)
+
+    def analyze(self, path):
+        """Parse a Python file and analyze it for import and class data
```
erikrose: Can you mention that this stows the results in instance vars? It'll make a little faster (for me) to reason about.

dxr/plugins/python/indexers.py
erikrose: Cool. After this lands.

dxr/plugins/python/indexers.py
```diff
@@ -100,3 +233,95 @@ def _needle(self, needle_type, name, start, end):
                    Position(row=end[0],
                             col=end[1]))
         )
+
+    def get_bases_for_class(self, absolute_class_name):
+        """Return a list of all the classes that the given class
```
erikrose: Is there any reason this can't be a generator? It's be more RAM- and cache-efficient and thus probably faster, and its caller is already a generator anyway.

dxr/plugins/python/indexers.py
```diff
@@ -100,3 +233,95 @@ def _needle(self, needle_type, name, start, end):
...
+        bases = []
+        for base in self.class_bases.get(absolute_class_name, []):
+            base = normalize_name(self.names, base)
+            bases.append(base)
+            if base in self.class_bases:
+                bases += self.get_bases_for_class(base)
+        return bases
+
+    def get_derived_for_class(self, absolute_class_name):
+        """Return a list of all the classes that derive from the given
```
erikrose: Same question about being a generator

dxr/plugins/python/indexers.py
```diff
@@ -80,6 +188,8 @@ def _all_needles(self):
         # Run through the AST looking for things to index!
         syntax_tree = ast.parse(self.contents.encode('utf-8'))
         for node in ast.walk(syntax_tree):
+            # Index classes and functions for the type: and function:
```
erikrose: I love me some block comments! :-D

dxr/testing.py
```diff
@@ -305,6 +319,16 @@ def _make_file(path, filename, contents):
         file.write(contents.encode('utf-8'))
 
 
+def _format_dict_values(dictionary, *args, **kwargs):
```
erikrose: It might be simpler to just have a SingleFileTestCase.get_config() method that instantiates a dict inline and returns it. Subclasses could customize the dict by calling the superclass method and messing with the result. Plus, get_config() could take required args (like `config_dir_path`) explicitly (and document them) rather than just hoping the caller passes them in. We wouldn't need to use voodoo like deepcopy() or recursively descend through all the dict keys.

```python
    def get_config(config_dir_path):
        return {
            'DXR': {
                ...
                'temp_folder': '{config_dir_path}/temp'.format(config_dir_path),
            },
            ...
        }
```

dxr/plugins/python/tests/test_imports/code/child.py
```diff
@@ -0,0 +1,25 @@
+import parent
+import parent as blarent
+from . import parent as carent
```
erikrose: lol, blarent, carent :-D

dxr/testing.py
erikrose: I see you're doing some of that already, as I read PythonSingleFileTestCase.

Osmose: I like that a lot better, I wasn't quite happy with what I had after the rebase but didn't give it enough thought before pushing. :D

dxr/plugins/python/indexers.py
```diff
@@ -89,6 +199,29 @@ def _all_needles(self):
...
+            # bases: filters.
+            if isinstance(node, ast.ClassDef):
+                class_name = module_name + '.' + node.name
+
+                # Index both the full absolute base name and the
+                # local name for conveniece.
+                bases = self.get_bases_for_class(class_name)
+                for base in bases:
+                    base_name = base.split('.')[-1]
+                    yield self._needle('py_derived', base, node.start,
```
erikrose: Rather than filing 2 separate needles, have you considered using (and promoting up to dxr.filters) clang.filters._QualifiedNameFilter and using a needle like QUALIFIED_NEEDLE? That way, the `+` syntax for fully qualified names would work, so you could search for `+foo.bar` and find just that rather than also finding "foo.bar.foo.bar".

dxr/plugins/python/indexers.py
```diff
@@ -100,3 +233,95 @@ def _needle(self, needle_type, name, start, end):
...
+                bases += self.get_bases_for_class(base)
+        return bases
+
+    def get_derived_for_class(self, absolute_class_name):
+        """Return a list of all the classes that derive from the given
+        class in their canonical form.
+
+        """
+        derived_class_names = []
+        for derived in self.derived_classes.get(absolute_class_name, []):
```
erikrose: This is already a defaultdict, so why the `, []`?

dxr/plugins/python/indexers.py
erikrose: Unfortunately, we lack `yield from`, so you'd need to make use of chain.from_iterable().

dxr/plugins/python/indexers.py
```diff
@@ -35,19 +42,115 @@
...
+        for node in ast.walk(syntax_tree):
+            # Save the base classes of any class we find.
+            if isinstance(node, ast.ClassDef):
+                class_path = module_path + '.' + node.name
+                bases = []
+                for base in node.bases:
+                    base_name = convert_node_to_name(base)
+                    if base_name:
+                        bases.append(module_path + '.' + base_name)
+                self.class_bases[class_path] = bases
```
erikrose: Later, if this ends up running out of RAM on large projects, you might find needles_from_graph() educational. There we compress a map much like this one by expressing it as a graph. That way, you don't store "C→B→A" reduntantly if you store both "D→C→B→A" and "E→C→B→A" (where "→" means "is a subclass of").

dxr/testing.py
erikrose: Yeah, and I wrote the new config system 3 times. :-)

dxr/plugins/python/__init__.py
```diff
@@ -1,4 +1,16 @@
...
+    tree_to_index=TreeToIndex,
+    mappings=mappings,
+    config_schema = {
+        'python_path': str
```
erikrose: You should probably use the AbsPath thing from config.py rather than str, so you'll be immune to working directory changes (and support unicode paths, and SPOT, and all that).

dxr/plugins/python/tests/test_imports/test_imports.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        self.found_line_eq('bases:child.RelativeImportChildClass',
+                           'class <b>ParentClass</b>(object):', 1)
+
+    def test_derived(self):
+        """Make sure the derived: filter matches child classes that
+        import from another file in a variety of ways.
+
+        """
+        self.found_lines_eq('derived:parent.ParentClass', [
```
erikrose: Deliciously pathological. I ♥ it.

dxr/plugins/python/tests/test_imports/test_imports.py
```diff
@@ -0,0 +1,56 @@
...
+            ('class <b>FromImportChildClass</b>(ParentClass):', 8),
+            ('class <b>ImportChildClass</b>(parent.ParentClass):', 12),
+            ('class <b>FromImportAsChildClass</b>(PClass):', 16),
+            ('class <b>ImportAsChildClass</b>(blarent.ParentClass):', 20),
```
erikrose: Yeah, having line numbers change is another reason to keep too many tests from sharing a source file. :-)

erikrose:
That's all I have to say about that! It looks really good. Let's just hit that handful of comments, and I'll have one more quick look, and then we can merge.

Osmose:
Phew! Finally got through all the feedback. Major new additions are fixes to the failing test cases and moving `QualifiedNameFilter` and `QUALIFIED_NEEDLE` up into `dxr.filters` and `dxr.indexers` respectively.

Only piece of feedback I ignored was https://github.com/mozilla/dxr/pull/390#discussion_r25814825 as this PR is getting gigantic and I'd rather refactor that in a separate patch after landing some of the other branches I have in waiting.

erikrose:
It says "later" on it, so ignoring for now was my intent for you. :-)

dxr/testing.py
```diff
@@ -1,5 +1,7 @@
 import cgi
+import collections
 from commands import getstatusoutput
+from copy import deepcopy
```
erikrose: No longer needed

dxr/testing.py
```diff
@@ -1,5 +1,7 @@
 import cgi
+import collections
```
erikrose: Not used

dxr/plugins/python/tests/__init__.py
```diff
@@ -0,0 +1,18 @@
+from dxr.testing import SingleFileTestCase
+
+
+class PythonSingleFileTestCase(SingleFileTestCase):
+    """Test case suited to testing the Python plugin."""
+    source_filename = 'main.py'
+
+    @classmethod
+    def get_config(cls, config_dir_path):
+        config = super(PythonSingleFileTestCase, cls).get_config(config_dir_path)
```
erikrose: Here you need to say PythonSingleFileTestCase instead of cls. It's akin to what you do in an instance method's super call. Yeah, I know it sucks. But otherwise, this won't work correctly if somebody subclasses PythonSingleFileTestCase.

erikrose: Never mind; I'm an idiot and read it backwards.

erikrose:
Remove those 2 imports and merge! r+! (Bonus point if you can figure out why Leeroy didn't run the tests)

Osmose:
giphy

PR #399, opened 2015-03-08
Add callers: and called-by: filters for the Python plugin.

Osmose:
Because the return order of `ast.walk` is not guaranteed to mean anything, I needed to be able to control the order that we walk through the tree if I wanted to correctly assign function calls to the right caller. The first commit refactors the Python plugin to move all the searching for needles to a NodeVisitor subclass. The second commit adds the new filters and some tests for them.

I spent some time trying to see if I could make a NodeVisitor subclass that returned a generator of needles, but I couldn't figure anything out, and I didn't want to copy over the NodeVisitor code and tweak it just for that. Oh well.

### DIFF
--- a/dxr/plugins/python/indexers.py
+++ b/dxr/plugins/python/indexers.py
@@ -21,6 +21,8 @@ mappings = {
             'py_function': QUALIFIED_NEEDLE,
             'py_derived': QUALIFIED_NEEDLE,
             'py_bases': QUALIFIED_NEEDLE,
+            'py_callers': QUALIFIED_NEEDLE,
+            'py_called_by': QUALIFIED_NEEDLE,
         },
     },
 }
@@ -123,13 +125,47 @@ class TreeToIndex(TreeToIndexBase):
 
 
 class IndexingNodeVisitor(ast.NodeVisitor):
+    """Node visitor that walks through the nodes in an abstract syntax
+    tree and finds interesting things to index.
+
+    """
+
     def __init__(self, file_to_index):
         self.file_to_index = file_to_index
+        self.function_call_stack = []  # List of lists of function names.
+        self.needles = None
+
+    def get_needles(self, node):
+        """Walk through the given AST and return needles to index from
+        the tree.
+
+        """
         self.needles = []
+        self.visit(node)
+        return self.needles
 
     def visit_FunctionDef(self, node):
+        # Index the function itself for the function: filter.
         start, end = self.file_to_index.get_node_start_end(node)
         self.yield_needle('py_function', node.name, start, end)
+
+        # Index function calls within this function for the callers: and
+        # called-by filters.
+        self.function_call_stack.append([])
+        self.generic_visit(node)
+        call_needles = self.function_call_stack.pop()
+        for name, call_start, call_end in call_needles:
+            self.yield_needle('py_callers', name, start, end)
+            self.yield_needle('py_called_by', node.name, call_start, call_end)
+
+    def visit_Call(self, node):
+        # Save this call if we're currently tracking function calls.
+        if self.function_call_stack:
+            call_needles = self.function_call_stack[-1]
+            name = convert_node_to_name(node.func)
+            if name:
+                start, end = self.file_to_index.get_node_start_end(node)
+                call_needles.append((name, start, end))
 
         self.generic_visit(node)
 
@@ -197,16 +233,15 @@ class FileToIndex(FileToIndexBase):
         return is_interesting(self.path)
 
     def needles_by_line(self):
-        self.analyze_tokens()
+        self.node_start_table = self.analyze_tokens()
 
         visitor = IndexingNodeVisitor(self)
-        syntax_tree = ast.parse(self.contents.encode('utf-8'))
-        visitor.visit(syntax_tree)
+        syntax_tree = ast.parse(self.contents)
 
         return iterable_per_line(
             with_start_and_end(
                 split_into_lines(
-                    visitor.needles
+                    visitor.get_needles(syntax_tree)
                 )
             )
         )
@@ -220,7 +255,7 @@ class FileToIndex(FileToIndexBase):
         # their 'def' and 'class' tokens. To get the position of their
         # names, we look for 'def' and 'class' tokens and store the
         # position of the token immediately following them.
-        self.node_start_table = {}
+        node_start_table = {}
         previous_start = None
         token_gen = tokenize.generate_tokens(StringIO(self.contents).readline)
 
@@ -231,23 +266,27 @@ class FileToIndex(FileToIndexBase):
             if tok_name in ('def', 'class'):
                 previous_start = start
             elif previous_start is not None:
-                self.node_start_table[previous_start] = start
+                node_start_table[previous_start] = start
                 previous_start = None
 
+        return node_start_table
 
     def get_node_start_end(self, node):
         """Return start and end positions within the file for the given
         AST Node.
 
         """
-        start = (node.lineno, node.col_offset)
+        start = node.lineno, node.col_offset
         if start in self.node_start_table:
             start = self.node_start_table[start]
 
+        end = None
         if isinstance(node, ast.ClassDef) or isinstance(node, ast.FunctionDef):
-            end = (start[0], start[1] + len(node.name))
-        else:
-            end = None
+            end = start[0], start[1] + len(node.name)
+        elif isinstance(node, ast.Call):
+            name = convert_node_to_name(node.func)
+            if name:
+                end = start[0], start[1] + len(name)
 
         return start, end
 
@@ -274,12 +313,18 @@ class FileToIndex(FileToIndexBase):
 
 
 def line_needle(needle_type, name, start, end, qualname=None):
+    data = {
+        'name': name,
+        'start': start[1],
+        'end': end[1]
+    }
+
+    if qualname:
+        data['qualname'] = qualname
+
     return (
         needle_type,
-        {'name': name,
-         'qualname': qualname,
-         'start': start[1],
-         'end': end[1]},
+        data,
         Extent(Position(row=start[0],
                         col=start[1]),
                Position(row=end[0],
--- a/dxr/plugins/python/indexers.py
+++ b/dxr/plugins/python/indexers.py
@@ -122,12 +122,50 @@ class TreeToIndex(TreeToIndexBase):
                                self.derived_classes, self.names)
 
 
-class FileToIndex(FileToIndexBase):
-    needle_types = {
-        ast.ClassDef: 'py_type',
-        ast.FunctionDef: 'py_function',
-    }
+class IndexingNodeVisitor(ast.NodeVisitor):
+    def __init__(self, file_to_index):
+        self.file_to_index = file_to_index
+        self.needles = []
 
+    def visit_FunctionDef(self, node):
+        start, end = self.file_to_index.get_node_start_end(node)
+        self.yield_needle('py_function', node.name, start, end)
+
+        self.generic_visit(node)
+
+    def visit_ClassDef(self, node):
+        # Index the class itself for the type: filter.
+        start, end = self.file_to_index.get_node_start_end(node)
+        self.yield_needle('py_type', node.name, start, end)
+
+        # Index the class hierarchy for classes for the derived: and
+        # bases: filters.
+        class_name = self.file_to_index.module_name + '.' + node.name
+
+        # Index both the full absolute base name and the
+        # local name for conveniece.
+        bases = self.file_to_index.get_bases_for_class(class_name)
+        for qualname in bases:
+            name = qualname.split('.')[-1]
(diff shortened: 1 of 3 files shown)


Osmose:
@erikrose r?

erikrose:
Sorry this took so long.

So could we later use your import mapping to support fully qualified caller searches, like `+function:urllib.parse()`?

dxr/plugins/python/indexers.py
```diff
@@ -153,30 +227,35 @@ def __init__(self, path, contents, plugin_name, tree, python_path,
         self.derived_classes = derived_classes
         self.names = names
 
+        self.module_name = path_to_module(self.python_path, self.path)
+
     def is_interesting(self):
         return is_interesting(self.path)
 
     def needles_by_line(self):
+        self.analyze_tokens()
```
erikrose: What do you think of making analyze_tokens() return its results and changing this statement to assign them to self.node_start_table? I generally like to bubble instance attr assignment up as far as possible to make clear the side effects. Passing the entire file_to_index into the visitor smells a little to me, too, but I can't think of a better idea right now.

Osmose: Good idea!

dxr/plugins/python/indexers.py
```diff
@@ -122,12 +124,84 @@ def file_to_index(self, path, contents):
...
+        start, end = self.file_to_index.get_node_start_end(node)
+        self.yield_needle('py_function', node.name, start, end)
+
+        # Index function calls within this function for the callers: and
+        # called-by filters.
+        self.function_call_stack.append([])
+        self.generic_visit(node)
+        call_needles = self.function_call_stack.pop()
+        for name, call_start, call_end in call_needles:
+            self.yield_needle('py_callers', name, start, end)
```
erikrose: We shouldn't yield `'qualname': None` keys for these; it just wastes space and makes the query parts QualifiedNameFilterBase emits for fully qualified queries useless busy-work. If we're not ready to do the qualnames yet, we should use NameFilterBase and omit those keys. I don't think it's important to change the needle mapping; I don't think the unused fields cost anything.

dxr/plugins/python/indexers.py
```diff
@@ -153,30 +227,35 @@ def __init__(self, path, contents, plugin_name, tree, python_path,
...
+        self.module_name = path_to_module(self.python_path, self.path)
+
     def is_interesting(self):
         return is_interesting(self.path)
 
     def needles_by_line(self):
+        self.analyze_tokens()
+
+        visitor = IndexingNodeVisitor(self)
+        syntax_tree = ast.parse(self.contents.encode('utf-8'))
```
erikrose: Orly? It wants a str? How about that.

dxr/plugins/python/indexers.py
```diff
@@ -187,54 +266,28 @@ def _all_needles(self):
...
-                   Position(row=end[0],
-                            col=end[1]))
-        )
+
+    def get_node_start_end(self, node):
+        """Return start and end positions within the file for the given
+        AST Node.
+
+        """
+        start = (node.lineno, node.col_offset)
```
erikrose: You don't need the parens if you don't want. But if they make you happy, keep them. :-)

dxr/plugins/python/indexers.py
```diff
@@ -153,30 +227,35 @@ def __init__(self, path, contents, plugin_name, tree, python_path,
...
-        module_name = path_to_module(self.python_path, self.path)
-
         # AST nodes for classes and functions point to the position of
         # their 'def' and 'class' tokens. To get the position of their
         # names, we look for 'def' and 'class' tokens and store the
         # position of the token immediately following them.
-        node_start_table = {}
+        self.node_start_table = {}
         previous_start = None
         token_gen = tokenize.generate_tokens(StringIO(self.contents).readline)
```
erikrose: We're passing this a unicode and the visitor a str. Is that going to cause inconsistencies?

dxr/plugins/python/indexers.py
```diff
@@ -122,12 +124,84 @@ def file_to_index(self, path, contents):
...
+class IndexingNodeVisitor(ast.NodeVisitor):
+    """Node visitor that walks through the nodes in an abstract syntax
+    tree and finds interesting things to index.
+
+    """
+
+    def __init__(self, file_to_index):
+        self.file_to_index = file_to_index
+        self.function_call_stack = []
+        self._needles = None
```
erikrose: This doesn't strike me as any more private that the above 2. Let's be consistent, one way or the other.

dxr/plugins/python/indexers.py
```diff
@@ -122,12 +124,84 @@ def file_to_index(self, path, contents):
...
-    }
+class IndexingNodeVisitor(ast.NodeVisitor):
+    """Node visitor that walks through the nodes in an abstract syntax
+    tree and finds interesting things to index.
+
+    """
+
+    def __init__(self, file_to_index):
+        self.file_to_index = file_to_index
+        self.function_call_stack = []
```
erikrose: It'd be nice to note what this data structure is like, so my addled brain can get oriented.

dxr/plugins/python/indexers.py
```diff
@@ -122,12 +124,84 @@ def file_to_index(self, path, contents):
...
+        self.function_call_stack.append([])
+        self.generic_visit(node)
+        call_needles = self.function_call_stack.pop()
+        for name, call_start, call_end in call_needles:
+            self.yield_needle('py_callers', name, start, end)
+            self.yield_needle('py_called_by', node.name, call_start, call_end)
+
+    def visit_Call(self, node):
+        # Save this call if we're currently tracking function calls.
+        if len(self.function_call_stack) > 0:
```
erikrose: Or just `if self.function_call_stack`

dxr/plugins/python/indexers.py
```diff
@@ -21,6 +21,8 @@
             'py_function': QUALIFIED_NEEDLE,
             'py_derived': QUALIFIED_NEEDLE,
             'py_bases': QUALIFIED_NEEDLE,
+            'py_callers': QUALIFIED_NEEDLE,
+            'py_called_by': QUALIFIED_NEEDLE,
```
erikrose: It looks like I quit supporting called-by queries in clang awhile back. There was some implementation motivation, and I couldn't think of a use case for them. Can you?

dxr/plugins/python/indexers.py
erikrose: Do I reason correctly that this would test false only if there were a call in global module space or maybe in a class def?

dxr/plugins/python/indexers.py
```diff
@@ -122,12 +124,84 @@ def file_to_index(self, path, contents):
...
+
+    def visit_FunctionDef(self, node):
+        # Index the function itself for the function: filter.
+        start, end = self.file_to_index.get_node_start_end(node)
+        self.yield_needle('py_function', node.name, start, end)
+
+        # Index function calls within this function for the callers: and
+        # called-by filters.
+        self.function_call_stack.append([])
+        self.generic_visit(node)
```
erikrose: This append/visit/pop thing is elegant and charming. :-) You get to collect the results of visit_Call without having to rewrite the whole visitation framework to preserve its results. Nice work.

dxr/plugins/python/indexers.py
```diff
@@ -187,54 +266,28 @@ def _all_needles(self):
...
-             'qualname': qualname,
-             'start': start[1],
-             'end': end[1]},
-            Extent(Position(row=start[0],
-                            col=start[1]),
-                   Position(row=end[0],
-                            col=end[1]))
-        )
+
+    def get_node_start_end(self, node):
```
erikrose: I wonder if it would be possible to get the start/end positions by walking the ast itself, rather than having to tokenize each file twice. Did you already look into that? (Even if it is possible, it's not necessary to merge this.)

erikrose:
That's all I have to say about that. All in all, looks good to me!

Osmose:
> So could we later use your import mapping to support fully qualified caller searches, like +function:urllib.parse()?

I believe so. You know how it traces the imports back until they're not imported anymore? I could have a function that yields each of those jumps and then index each of them as needles, similar to how I was doing multiple needles for qualnames before switching to use `QUALIFIED_NEEDLE`.

dxr/plugins/python/indexers.py
Osmose: I wish I wrote a comment or test about this, but I seem to remember running into some issue where self.contents caused ast.parse to fail. But I cannot for the life of me remember what it was.

Maybe I'll remove it and see if I can come up with anything? `test_unicode.py` maybe?

Osmose: Upon further research this was encoded like this before I started working on this plugin, so I have no idea.

dxr/plugins/python/indexers.py
Osmose: I imagine it could. I think I'll remove that encode call and see what happens.

dxr/plugins/python/indexers.py
Osmose: You reason correctly.

dxr/plugins/python/indexers.py
Osmose: I can't think of a very strong one. Maybe as part of some larger analysis rather than a user-facing filter. Should I remove it?

dxr/plugins/python/indexers.py
Osmose: So the root of the problem is that `ClassDef` and `FunctionDef` nodes give their `col_offset` as the column where the `class`/`def` keyword appears, whereas we want to highlight just the class or function name. Another thing compounds this problem: we don't know where to stop highlighting except by finding the length of the class/function name, so even if we said "Well fine, start highlighting at the def/class token" the highlighting would stop at the wrong place.

Hell, there's even evil code like this:

```python
class \
\
   LineContinuationsHateYou    \
 (object):
    pass # Yeah this is valid wtf
```

As far as I can tell, the AST doesn't have any extra info we can use here. So we have to do something special.

We were discussing in channel yesterday that maybe we could do something like a substring search in the file content for the first occurrence of the class name after the location given by the AST. That'd avoid tokenizing the entire file, although I'm not sure if it'd be clean/performant.

Osmose:
Updated. Kept `called-by` for now until you reply.

dxr/plugins/python/indexers.py
erikrose: I lean that way.

dxr/plugins/python/indexers.py
erikrose: Yeah, any file for which contains_text() is true is going to be Unicode. In fact, that's the definition of that function.

erikrose: Just talked to Jeff. He does seem to think it exists for a reason. Can't recall the details, though. Let's put a frightening comment on it and leave it there for now.

erikrose: Maybe something to do with explicit encoding specifications?

erikrose: OTOH, I just rediscovered my comment below about inconsistencies, which frightens me more than any alleged reason of Jeff's, so yeah: let's be consistent.

erikrose: Okay, here's the skinny: the ast module works in terms of UTF-8 offsets. From the ast docs…

> the col_offset is the UTF-8 byte offset of the first token that generated the node

It'll crash if fed a unicode with certain explicit encoding declarations up top. If we pass it a utf-8 bytestring, it doesn't try parsing the declaration. Se we should keep this as is, and perhaps we should change the tokenize call in some way. We'll also need to convert back from utf-8 offsets to unicode ones to conform to DXR's API, but that can wait if we ticket it. That shouldn't be too labor-intense if we make use of FileToSkim._line_offsets().

erikrose: Specifically, this crashes:

```python
>>> parse(u'# -*- coding: utf-8 -*-\n')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 37, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 0
SyntaxError: encoding declaration in Unicode string
```

This doesn't:

```python
>>> parse(u'# -*- coding: utf-8 -*-\n'.encode('utf-8'))
<_ast.Module object at 0x10a6d9f90>
```

erikrose:
So let's (1) kill off called-by and then merge. Then the next thing should be to force the utf-8 offsets back into Unicode and write some tests to make sure the offsets from tokenize and ast agree in the face of unicode. That doesn't have to be part of this PR since, if it's broken here, it was already broken before.

Osmose:
Removed and merged. (Sorry if you wanted to take a last look at this, I interpreted you as giving this an r+wc)

erikrose:
Nope, you did the right thing!

PR #400, opened 2015-03-08
Fix bug 1137356: Warn on invalid Python files and do not index.

Osmose:
Catches errors during the pre-indexing analysis and flags those files to be ignored, warning the user of the issue.

### DIFF
--- a/dxr/plugins/python/indexers.py
+++ b/dxr/plugins/python/indexers.py
@@ -4,6 +4,7 @@ import token
 import tokenize
 from collections import defaultdict
 from StringIO import StringIO
+from warnings import warn
 
 from dxr.build import file_contents, unignored
 from dxr.indexers import (Extent, FileToIndex as FileToIndexBase,
@@ -25,14 +26,27 @@ mappings = {
 }
 
 
+class _FileToIgnore(object):
+    """A file that we don't want to bother indexing, usually due to
+    syntax errors.
+
+    """
+    def is_interesting(self):
+        return False
+FILE_TO_IGNORE = _FileToIgnore()
+
+
 class TreeToIndex(TreeToIndexBase):
     def __init__(self, *args, **kwargs):
         super(TreeToIndex, self).__init__(*args, **kwargs)
 
-        self.python_path = getattr(self.tree, 'plugin_python_path', self.tree.source_folder)
+        self.python_path = self.plugin_config.python_path
+
+        # Post-build analysis results.
         self.base_classes = defaultdict(list)
         self.derived_classes = defaultdict(list)
         self.names = {}
+        self.ignore_paths = set()
 
     @property
     def unignored_files(self):
@@ -59,7 +73,16 @@ class TreeToIndex(TreeToIndexBase):
 
         """
         module_path = path_to_module(self.python_path, path)
-        syntax_tree = ast.parse(file_contents(path, self.tree.source_encoding))
+
+        try:
+            syntax_tree = ast.parse(file_contents(path, self.tree.source_encoding))
+        except SyntaxError as err:
+            rel_path = os.path.relpath(path, self.tree.source_folder)
+            warn('Failed to analyze {filename} due to error "{error}".'.format(
+                 filename=rel_path, error=err))
+            self.ignore_paths.add(rel_path)
+            return
+
         for node in ast.walk(syntax_tree):
             # Save the base classes of any class we find.
             if isinstance(node, ast.ClassDef):
@@ -91,9 +114,12 @@ class TreeToIndex(TreeToIndexBase):
                     self.names[absolute_local_name] = import_name
 
     def file_to_index(self, path, contents):
-        return FileToIndex(path, contents, self.plugin_name, self.tree,
-                           self.python_path, self.base_classes,
-                           self.derived_classes, self.names)
+        if path in self.ignore_paths:
+            return FILE_TO_IGNORE
+        else:
+            return FileToIndex(path, contents, self.plugin_name, self.tree,
+                               self.python_path, self.base_classes,
+                               self.derived_classes, self.names)
 
 
 class FileToIndex(FileToIndexBase):
@@ -153,6 +179,7 @@ class FileToIndex(FileToIndexBase):
         node_start_table = {}
         previous_start = None
         token_gen = tokenize.generate_tokens(StringIO(self.contents).readline)
+
         for tok_type, tok_name, start, end, _ in token_gen:
             if tok_type != token.NAME:
                 continue
--- a/dxr/plugins/python/tests/test_parse_errors.py
+++ b/dxr/plugins/python/tests/test_parse_errors.py
@@ -1,0 +1,64 @@
+import warnings
+from textwrap import dedent
+
+from nose.tools import eq_, ok_
+
+from dxr.plugins.python.tests import PythonSingleFileTestCase
+
+
+class WarningTestCase(PythonSingleFileTestCase):
+    @classmethod
+    def get_config(cls, config_dir_path):
+        # We need to disable workers because catch_warnings does not
+        # support multi-threaded or multi-process capturing.
+        config = super(WarningTestCase, cls).get_config(config_dir_path)
+        config['DXR']['workers'] = 0
+        return config
+
+    @classmethod
+    def setup_class(cls):
+        # Wrap the normal setup to catch warnings.
+        with warnings.catch_warnings(record=True) as caught_warnings:
+            warnings.simplefilter('always')
+            super(WarningTestCase, cls).setup_class()
+        cls.warnings = caught_warnings
+
+
+# Store common test method on a mixin so nose doesn't try to run it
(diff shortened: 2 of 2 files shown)


Osmose:
@erikrose r?

dxr/plugins/python/tests/test_parse_errors.py
```diff
@@ -0,0 +1,70 @@
...
+        # Wrap the normal setup to catch warnings.
+        with warnings.catch_warnings(record=True) as caught_warnings:
+            warnings.simplefilter('always')
+            super(WarningTestCase, cls).setup_class()
+            cls.warnings = caught_warnings
+
+
+class InvalidPythonTest(WarningTestCase):
+    source = dedent("""
+    from foo import bar
```
erikrose: Let's indent these strings one more notch so other class vars, if they existed, would stick out. That's the convention we were using.

dxr/plugins/python/tests/test_parse_errors.py
```diff
@@ -0,0 +1,70 @@
...
+
+    biff = -8bazinvalid  # That ain't right
```
erikrose: Yes, I believe it should be `-7bazinvalid`.

dxr/plugins/python/indexers.py
```diff
@@ -25,14 +26,28 @@
...
+class _FileToIgnore(object):
+    """A file that we don't want to bother indexing, usually due to
+    syntax errors.
+
+    """
+    def is_interesting(self):
+        return False
+FILE_TO_IGNORE = _FileToIgnore()
+
+
```
erikrose: 3 skipped lines? What is this madness?!

dxr/plugins/python/indexers.py
```diff
@@ -25,14 +26,28 @@
...
+
 class TreeToIndex(TreeToIndexBase):
     def __init__(self, *args, **kwargs):
         super(TreeToIndex, self).__init__(*args, **kwargs)
 
         self.python_path = getattr(self.tree, 'plugin_python_path', self.tree.source_folder)
```
erikrose: We should change this to use the new config system (config_schema, self.plugin_config(), etc.) at some point. I thought we had....

dxr/plugins/python/tests/test_parse_errors.py
```diff
@@ -0,0 +1,70 @@
...
+        # We need to disable workers because catch_warnings does not
+        # support multi-threaded or multi-process capturing.
+        config = super(WarningTestCase, cls).get_config(config_dir_path)
+        config['DXR']['workers'] = 0
+        return config
+
+    @classmethod
+    def setup_class(cls):
+        # Wrap the normal setup to catch warnings.
+        with warnings.catch_warnings(record=True) as caught_warnings:
```
erikrose: Kudos for busting out catch_warnings. Very thorough.

dxr/plugins/python/tests/test_parse_errors.py
```diff
@@ -0,0 +1,70 @@
...
+            warnings.simplefilter('always')
+            super(WarningTestCase, cls).setup_class()
```
erikrose: I suppose this is really the only line that needs to be inside the `with` block.

dxr/plugins/python/tests/test_parse_errors.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_parse_error(self):
+        """Make sure nothing is indexed for a file with invalid Python
+        code, and that the build process raised warnings about it.
+
+        """
+        self.found_nothing('class:TestClass')
+
+        # One warning for the pre-index analyzing, and one for indexing.
+        self.assertEqual(len(self.warnings), 1)
```
erikrose: Just a consistency nit: we've been using nose.tools's ok_ and eq_ throughout the codebase.

dxr/plugins/python/tests/test_parse_errors.py
```diff
@@ -0,0 +1,70 @@
...
+    """)
+
+    def test_parse_error(self):
+        """Make sure nothing is indexed for a file with invalid Python
+        code, and that the build process raised warnings about it.
+
+        """
+        self.found_nothing('class:TestClass')
+
+        # One warning for the pre-index analyzing, and one for indexing.
```
erikrose: I'm confused. Can you help me understand what this means?

erikrose: Oh, the global analysis stuff.

erikrose: Still, this sounds like it's testing for 2 warnings, and there's only 1, so I'm still confused.

dxr/plugins/python/tests/test_parse_errors.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class InvalidTokenTest(WarningTestCase):
+    source = dedent("""
+    from foo import bar
+
+    class TestClass(object):
+        pass
+
+    '''Tokenizer fails on un-closed strings. Let's make sure if we do
```
erikrose: Is it even possible for your tokenization code to receive an untokenizable string anymore, now that the ignore-on-syntax-error stuff is in place? AFAICT, both this and the program in the test above fail with SyntaxErrors, making these two tests equivalent.

dxr/plugins/python/tests/test_parse_errors.py
Osmose: Leftover comment by accident, sorry!

dxr/plugins/python/tests/test_parse_errors.py
Osmose: It isn't, but it used to be, and I figure since the code will move around a bunch in the future too it's reasonable to test this case in case someday we run the tokenizer first after a refactor.

erikrose: Okay, that makes some sense. Can you expand the "if we do" bit to clarify this?

dxr/plugins/python/tests/test_parse_errors.py
```diff
@@ -0,0 +1,70 @@
...
+    pass this to the tokenizer that it doesn't fail before the ast
+    parser does.
+    """)
+
+    def test_parse_error(self):
```
erikrose: And also, we should be able to factor this test up, in that case.

dxr/plugins/python/tests/test_parse_errors.py
Osmose: The simplefilter needs to be in there too because `catch_warnings` resets the filter on exit to whatever it previously was.

The `cls.warnings` part could be outside if you want that moved out. I figure it looks cleaner and is relatively harmless to keep inside, but I don't really care that much either way.

dxr/plugins/python/tests/test_parse_errors.py
Osmose: I don't quite get what you mean by that.

Osmose:
Updated based on feedback. :D

dxr/plugins/python/tests/test_parse_errors.py
erikrose: I mean test_parse_error is repeated here and in the test above, so we might as well spell it out only once. I'd probably factor it up to a common superclass.

dxr/plugins/python/tests/test_parse_errors.py
erikrose: I have a slight preference for moving it outside; generally I try to keep `with` blocks as small as possible so it's clear what the context manager's effects are important for. (Doubtless this grew out of the more important practice of keeping `try` blocks minimal.)

dxr/plugins/python/indexers.py
```diff
@@ -25,14 +26,27 @@
...
+        return False
+FILE_TO_IGNORE = _FileToIgnore()
+
+
 class TreeToIndex(TreeToIndexBase):
     def __init__(self, *args, **kwargs):
         super(TreeToIndex, self).__init__(*args, **kwargs)
 
-        self.python_path = getattr(self.tree, 'plugin_python_path', self.tree.source_folder)
+        self.python_path = self.plugin_config.python_path
```
erikrose: Unfortunately, this loses the defaulting to tree.source_folder. I'd like the config system to provide a mechanism for dependent defaults ultimately, but, for now, the best solution is probably to make it `Optional` in the config_schema, test for its presence here, and fall back to the default if absent.

Osmose: Removing that default was intentional; I think it's a good thing to force projects to be explicit about where the modules start in their project. 

Happy to change that if you disagree, I just wanted to make sure you knew it was intentional.

erikrose: Oh, then go right ahead! You're closer to that code than me.

Osmose:
Updated!

erikrose:
r+ assuming tests pass. I guess we should stick a sleep or something in jenkins.sh.

PR #402, opened 2015-03-10
On Jenkins, choose a random box name so test runs don't collide if they ...

erikrose:
...get assigned to the same worker.


erikrose:
Rebased and landed in db3ef213a500343043deb31a121da2735c01fb41.

PR #403, opened 2015-03-11
Accept http and https github repo

tojocky:
Hello,

First, thanks for this project. I really missed this.
This PR improve omniglot plugin by accepting http and https protocol github repository.
Also in my case I have repositories from other domains, so I removed the error.
In the end it contains a small doc typo correction.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -53,7 +53,7 @@ Here are the options that can live in the ``[DXR]`` section:
 ``enabled_plugins``
     Names of plugins to enable. Default: ``*``
 
-``filter_lang``
+``filter_language``
     The default programming language for this instance. Only filters registered
     for this language will be used. Default: ``C``
 
--- a/dxr/plugins/omniglot/htmlifier.py
+++ b/dxr/plugins/omniglot/htmlifier.py
@@ -177,7 +177,12 @@ class Git(VCS):
             if repo.endswith(".git"):
                 repo = repo[:-len(".git")]
             return "https" + repo[len("git"):]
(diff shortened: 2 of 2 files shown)


erikrose:
Hi, and thanks for the patch! Most development is currently happening on the `es` branch, which ports the entire back end to elasticsearch, support multiple languages, and makes many other changes. I would ask you to retarget your patch, except that most of your changes have already been done there. :-D The only thing that hasn't is to support non-GitHub repos. Though your solution doesn't raise an exception anymore, I don't think it returns a usable URL. The real solution is probably to add a configuration option to omniglot to manually specify a template string for building web URLs. Let me know if you're interested in doing that, and I'll be happy to point you in the right direction!

tojocky:
Hi, the configuration option to omniglot to manually specify a template string for building web URLs would be better idea.
Let me know the expected option name, expected format value and target configuration file.
I'm open to implement this and to target on the `es` branch.

erikrose:
There would be a little design work for you to do: do the various web front-ends for git expose a common set of URLs, or does each have its own? If the former, it could be enough to have one config value, but I suspect the latter is the case. If the latter, we might need a config value for each kind of page: diff, log, etc., with placeholders for the specific hash or file or whatever. http://dxr.readthedocs.org/en/es/configuration.html talks about the config file a bit. You'd want to add a plugin-specific option to omniglot (http://dxr.readthedocs.org/en/es/configuration.html#omniglot), and you'd probably want to format it something like the es_index option, with {placeholders_like_this}. We're in the #static channel on irc.mozilla.org if you want to chat some more. Thanks for helping out!

PR #405, opened 2015-03-20
Add module: and package: filters for the Python plugin.

Osmose:
The module filter actually finds modules _and_ packages. Perhaps the package filter isn't even necessary, but I dunno if we want to keep the module filter named "module" if it matches both and there's no package-only filter. Got any better names?

### DIFF
--- a/dxr/indexers.py
+++ b/dxr/indexers.py
@@ -9,32 +9,34 @@ from funcy import group_by, decorator, imapcat
 from os.path import join
 
 
+STRING_PROPERTY = {
+    'type': 'string',
+    'index': 'not_analyzed',
+    'fields': {
+        'lower': {  # for qualified_type direct searcher
+            'type': 'string',
+            'analyzer': 'lowercase'
+        }
+    }
+}
+
+
 # An unanlyzed string property that points to a value that can be exact- or
 # prefix-matched against and carries start/end bounds for highlighting. Has
 # both a name and a qualname.
-QUALIFIED_NEEDLE = {
+QUALIFIED_FILE_NEEDLE = {
     'type': 'object',
     'properties': {
-        'name': {
-            'type': 'string',
-            'index': 'not_analyzed',
-            'fields': {
-                'lower': {
-                    'type': 'string',
-                    'analyzer': 'lowercase'
-                }
-            }
-        },
-        'qualname': {
-            'type': 'string',
-            'index': 'not_analyzed',
-            'fields': {
-                'lower': {  # for qualified_type direct searcher
-                    'type': 'string',
-                    'analyzer': 'lowercase'
-                }
-            }
-        },
+        'name': STRING_PROPERTY,
+        'qualname': STRING_PROPERTY,
+    }
+}
+
+QUALIFIED_LINE_NEEDLE = {
+    'type': 'object',
+    'properties': {
+        'name': STRING_PROPERTY,
+        'qualname': STRING_PROPERTY,
         'start': {
             'type': 'integer',
             'index': 'no'  # just for highlighting
--- a/dxr/plugins/python/indexers.py
+++ b/dxr/plugins/python/indexers.py
@@ -4,28 +4,37 @@ import tokenize
 from StringIO import StringIO
 
 from dxr.build import unignored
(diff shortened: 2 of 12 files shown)


Osmose:
@erikrose r?

erikrose:
I really like the idea of just a `module` filter that finds both modules and packages. That seems like the common use case to me: "Somebody imported snoo. Where are all the possible snoos?" I think we can dispense with the package filter, like you suggest.

dxr/plugins/python/tests/test_modules_packages/makefile
```diff
@@ -0,0 +1,13 @@
+all:
```
erikrose: You probably will actually want to rebase this PR, since all this boilerplate stuff can now go byebye.

erikrose: Sorry I've let it sit so long.

Osmose: Ain't no thang

Osmose:
Rebased to only include the module filter and to use the new test format.

dxr/indexers.py
```diff
@@ -12,7 +12,33 @@
 # An unanlyzed string property that points to a value that can be exact- or
 # prefix-matched against and carries start/end bounds for highlighting. Has
 # both a name and a qualname.
-QUALIFIED_NEEDLE = {
+QUALIFIED_FILE_NEEDLE = {
+    'type': 'object',
+    'properties': {
```
erikrose: It might be nice to factor the values of the 'name' and 'qualname' hashes out. Right now they're repeated 4 times.

dxr/plugins/python/indexers.py
```diff
@@ -197,6 +204,13 @@ def visitor(self):
             self._visitor.visit(syntax_tree)
         return self._visitor
 
+    def needles(self):
+        # Index module name. For practical purposes, this includes
+        # __init__.py files for packages even though that's not
+        # _technically_ a module.
+        yield file_needle('py_module', name=local_name(self.module_name),
```
erikrose: The style I've been trying to follow is "either all args on 1 line (if they fit and read easily), or 1 arg per line".

dxr/plugins/python/tests/test_modules_packages/code/common_name_module.py
```diff
@@ -0,0 +1 @@
+"""Module with a common filename for testing multiple results."""
```
erikrose: Nitty nit nit: this isn't a sentence, so the gods do not grant it a period. (And I just used a colon wrong. ;-))

dxr/plugins/python/tests/test_modules_packages/test_modules_packages.py
```diff
@@ -0,0 +1,49 @@
+from dxr.testing import DxrInstanceTestCase
+
```
erikrose: If I search for `module:snook.eee`, will it find both `some/snook/eee/__init__.py` and `other/snook/eee/__init__.py`? If so, could we add a test for that?

erikrose: Oh, and I suppose we could rename the test folder from test_modules_packages to test_modules.

erikrose:
Looks great to me; I had only those few minor comments and questions.

erikrose:
This review was really easy. :-)

dxr/plugins/python/tests/test_modules_packages/test_modules_packages.py
Osmose: It won't because right now we don't support partial matches on dotted names. If you searched for `module:eee` it'd work, but since `snook.eee` is only a partial match it doesn't work.

A future PR adding partial match support would be great and is on my todo list. :D

Osmose:
Updated with fixes, is this good to merge?

dxr/plugins/python/tests/test_modules_packages/test_modules_packages.py
erikrose: Cool.

erikrose:
r+!

PR #407, opened 2015-03-25
Add menus for class inheritance to Python plugin.

Osmose:


Osmose:
@erikrose r?

dxr/plugins/clang/menus.py
```diff
@@ -5,10 +5,10 @@
 from flask import url_for
 
 from dxr.app import DXR_BLUEPRINT
+from dxr.utils import search_url as search
```
erikrose: Good rename.

dxr/plugins/python/indexers.py
```diff
@@ -194,12 +187,23 @@ def visit_ClassDef(self, node):
                               name=name, qualname=qualname,
                               start=start, end=end)
 
+        # Show a menu when hovering over this class.
+        self.yield_ref(start, end,
+                       menus.class_menu(self.file_to_index.tree, class_name))
```
erikrose: It seems like either `class_menu` or `menus.class` (ignoring the keyword for now) would be sufficient to expose your intent. I'd recommend either importing deeper (my favorite) or shortening the routine names.

erikrose:
Aside from that one naming nitpick, r+. Looking good.

PR #409, opened 2015-04-01
Add "Jump to definition" menu for Python function calls.

Osmose:
In order to be able to find the location of a Python function based only on it's name, this adds a DefinitionTree to the analysis that keeps track of class and function definitions across the entire tree. It also stores info about the imports so that it can resolve any name to the definition it refers to.

Using this, we look up the function using the name from the call node and fetch the path/line data from the definition. As of now this doesn't account for things like proper variable scope. For example, a function call like this wouldn't be able to find the right definition for bar:

```python
def foo():
    def bar():
        return 8
    blah = bar()
    return blah
```

But in basic cases, it works. In the future we can make call nodes aware of their scope so that `bar()` would know that it's looking for `foo.bar` and not just `bar`.

In the future I'd like to add variable definitions to the tree, including class attributes. Eventually it might make sense to turn `Definition` and `Import` into proper classes and give them subclasses that can store extra data, such as a `ClassDefinition` subclass that stores class inheritance info. I'm pretty sure most of the info that `TreeAnalysis` currently stores as `defaultdict`s could actually be stored in the definition tree.


Osmose:
@erikrose r?

.gitignore
```diff
@@ -2,8 +2,8 @@
 *.o
 *.so
 /dxr/static/js/templates.js
-**/tests/*/dxr-logs-*
```
erikrose: I'm just curious: how did logs and temp files get elsewhere?

dxr/menus.py
```diff
@@ -0,0 +1,12 @@
+"""Common menu constructors for use in plugins."""
+from os.path import basename
+
+from dxr.utils import browse_url
```
erikrose: I'm afraid we're going to have to rebase this one (or at least merge in from es); browse_url doesn't exist anymore.

erikrose: nm; you re-added it

erikrose:
Soon I'd like to change Jump To Definition to do a search rather than precompute the line number: https://bugzilla.mozilla.org/show_bug.cgi?id=986522#c13. Does that affect the usefulness of the DefinitionTree?

.gitignore
Osmose: When running `dxr serve` inside the test cases in `dxr/plugins/python/tests`.

Osmose:
Not really, as I think it's a superior data structure than TreeAnalysis anyway. One's just some bits of data we're interested in, while the other is a full representation of the definitions in the program.

dxr/plugins/python/analysis.py
```diff
@@ -202,4 +203,70 @@ def analyze_import(self, node):
...
-            self.tree_analysis.names[absolute_local_name] = import_name
+            import_tuple = Import(absolute_name=absolute_local_name,
+                                  line=node.lineno, col=node.col_offset,
+                                  import_name=import_name)
+            self.tree_analysis.definition_tree.add(import_tuple)
+
+
+# Namedtuples that are stored in DefinitionTree.
+Definition = namedtuple('Definition', ('absolute_name', 'line', 'col', 'path'))
+Import = namedtuple('Import', ('absolute_name', 'line', 'col', 'import_name'))
```
erikrose: I assume you went with namedtuples for size? I believe attr-based lookup in them is a little slower than a class. Probably a `__slots__`-based class is the best of both worlds; I doubt you need index-based access.

dxr/plugins/python/analysis.py
```diff
@@ -202,4 +203,70 @@ def analyze_import(self, node):
...
+
+
+class DefinitionTree(object):
+    """Stores definitions across the entire project, such as classes and
```
erikrose: I could use a little more about how this data structure is shaped and what it lets us do.

dxr/plugins/python/analysis.py
```diff
@@ -202,4 +203,70 @@ def analyze_import(self, node):
...
+        """Fetch the definition for the given name, following imports if
+        necessary.
+
+        If the definition is not found, return None. Top-level modules
+        that have not been added to the tree are assumed external and
+        valid, and definitions will be returned for them with no path
+        or position data.
+        """
+        names = name.split('.')
+        module_name = names.pop(0)
```
erikrose: I'm still trying to work out whether this is recursive, but keep in mind that popping things off the left of a list is inefficient: it has to memcpy all the other items in the list down a notch.

dxr/plugins/python/analysis.py
```diff
@@ -202,4 +203,70 @@ def analyze_import(self, node):
...
+            while isinstance(value, Import):  # Follow imports.
+                value = self.get(value.import_name)
+
+        return value
+
+    def normalize_name(self, absolute_name):
+        """Given a name, figure out the actual module that the thing
+        that name points to lives and return that name.
+
+        For example, if you have `from os import path` in a module
```
erikrose: Ah, so this returns the canonical name. This example was super-spiffy helpful to me. More like this! :-)

.gitignore
erikrose: That's weird. **/tests should have applied to that.

erikrose:
> Not really, as I think it's a superior data structure than TreeAnalysis anyway.

Thanks. That makes sense now that I'm farther down this review. Loading everything back into my head post-PyCon. :-)

dxr/plugins/python/utils.py
```diff
@@ -40,13 +47,38 @@ def trim_end(string, end):
...
-    and handles them specifically.
+def all_subclasses(cls):
+    """Yield all subclasses for the given class."""
+    for subclass in cls.__subclasses__():
+        yield subclass
+        for descendant in all_subclasses(subclass):
+            yield descendant
+
+
+# ast.NodeVisitor doesn't implement any of the visit_* methods, which
```
erikrose: Ouch, I'm sorry.

dxr/plugins/python/tests/test_menus/test_menus.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+
+    def test_definition_menu_method(self):
+        """Make sure the definition menu works for calls within a
+        method.
+
+        """
+        menu_on(self.source_page('main.py'), 'foo_method',
+            {'html': 'Jump to definition',
+             'href': browse('functions.py', 1)},
```
erikrose: I  ♥ your use of browse() here. So readable.

dxr/plugins/python/tests/test_menus/test_menus.py
```diff
@@ -0,0 +1,70 @@
...
+from dxr.plugins.python.tests import PythonSingleFileTestCase
+from dxr.testing import DxrInstanceTestCase, menu_on
+
+
+def search(query):
+    return '/code/search?q=' + quote(query)
+
+
+def browse(path, lineno=None):
+    return '/code/source/' + path + (('#' + unicode(lineno)) if lineno else '')
```
erikrose: This will sometimes return unicode and other times str. It works out fine so far, but let's pick one.

erikrose:
That's all I've got—just those few minor things. I'll wait to hear back from you on those.

Osmose:
Ran into a ton of different issues while trying to merge the latest `es` in (handling the special case that @nickolay found with `DefinitionTree` is tricky). I think at this point I'd be better off starting fresh with a new branch and thinking harder about the data structure, so I'mma close this and get back to it later.

nickolay:
Yes, attempting to handle the special cases (which are not so special - I found them in the dxr code itself) is tricky. FWIW, after spending much more time than I expected on the import bug, I started wondering if re-using an existing static analysis engine for python (like http://jedi.jedidjah.ch/en/latest/ ?) would be easier.

erikrose:
It's something I've always been open to.

We tried it with ternjs for JavaScript, but tern was loath to give us a dump of all the info; being geared toward providing support for editor plugins, it needed us to iterate over each symbol (which we'd have to come with ourselves) and ask about it separately. And then it turned out we'd have to add a lot of functionality to it anyway. We decided it wasn't worth the bother.

But if Jedi is a good match for our needs, then I'd love to see a proof-of-concept worked up. The code looks decent. Its import-chasing stuff is at https://github.com/davidhalter/jedi/blob/master/jedi/evaluate/imports.py. Like tern, it's geared toward supporting interactive use, so I'd be curious if we can straightforwardly get a full dump out of it. I'm also curious to what degree it covers our existing filters.

nickolay:
Right, but it seemed that perhaps changing an existing engine to suit our needs would still be easier long-term. Too bad it feels like too large a project for me to take on...

PR #410, opened 2015-04-02
Fix infinite scroll

Osmose:
Fixes the infinite scroll to count lines properly (it was previously counting number of files returned) and to append new lines for a file that is already on the page to that file instead of creating a new file group in the search results.

This supersedes #384, which was very helpful in helping me understand what was going wrong. :D

### DIFF
--- a/dxr/static/js/dxr.js
+++ b/dxr/static/js/dxr.js
@@ -266,7 +266,7 @@ $(function() {
                         // Update result count
                         resultsLineCount = countLines(data.results);
                         // Use the results.html partial so we do not inject the entire container again.
-                        populateResults('partial/results.html', data, true);
+                        populateResults(data, true);
                         // update URL with new offset
                         setHistoryState(dataOffset);
                         // start the scrolling poller
@@ -320,11 +320,10 @@ $(function() {
 
     /**
      * Populates the results template.
-     * @param {object} tmpl - The template to use to render results.
      * @param {object} data - The data returned from the query
      * @param {bool} append - Should the content be appended or overwrite
      */
-    function populateResults(tmpl, data, append) {
+    function populateResults(data, append) {
         data.www_root = dxr.wwwRoot;
         data.tree = dxr.tree;
         data.top_of_tree = dxr.wwwRoot + '/' + data.tree + '/source/';
@@ -337,10 +336,10 @@ $(function() {
 
         // If no data is returned, inform the user.
         if (!data.results.length) {
-            data.user_message = contentContainer.data('no-results');
-            contentContainer.empty().append(nunjucks.render(tmpl, data));
+            contentContainer
+                .empty()
+                .append(nunjucks.render('partial/results_container.html', data));
         } else {
-
             var results = data.results;
             resultsLineCount = countLines(results);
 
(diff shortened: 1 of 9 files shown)


Osmose:
@erikrose r?

dxr/static/js/dxr.js
```diff
@@ -140,7 +140,7 @@ $(function() {
         requestsInFlight = 0,  // Number of search requests in flight, so we know whether to hide the activity indicator
         displayedRequestNumber = 0,
         didScroll = false,
-        resultCount = 0,
+        resultsLineCount = 0,
```
erikrose: Can you remind me what this is even for? :-)

Osmose: From what I can tell, `previousDataLimit` is the amount of lines returned by the last request to the API (which is determined by the amount of lines we can show on a single screen, I think). If `resultsLineCount` is equal to `previousDataLimit` then presumably we got as many lines as we requested from the API, and there may be more (or we might be "lucky" and asked for just the right amount and hit the end of the list.

So it's used to determine if we need to fetch any more results when an infinite scroll request is triggered. I don't think it's the best way to determine that but I also just want to make it work for now, and refactor it down later.

In a related note, I'm not entirely sure the math is working out exactly right anyway, as locally infinite scroll seems to trigger too early, but I couldn't see where the issue was and figured it wasn't a blocker as it doesn't affect functionality much.

dxr/static/js/dxr.js
```diff
@@ -339,8 +350,31 @@ $(function() {
...
+                    .append(nunjucks.render('partial/results_container.html', data));
+            } else {
+                var resultsList = contentContainer.find('.results');
+
+                // If the first result is already on the page (meaning we showed
+                // some of its lines but just fetched more), add new lines to
+                // it instead of adding a new result section.
+                var firstResult = data.results[0];
+                var domFirstResult = resultsList.find(
+                    '.result[data-path="' + firstResult.path + '"]');
```
erikrose: I'm going to make a few comments just for "here's something to think about" purposes, but they shouldn't block merging this.

Comment one: this is probably a linear search, depending on what kind of data structures the browser uses. We could speed it up, if we need to, by squirreling away the path of the last result in a known place somewhere (or even just looking it up `data-path` on the last result and comparing against that).

erikrose:
This isn't quite right. I can make it do all kinds of weird things, like cutting off (and printing the footer) before all the matches are found, or showing matches 1-73 and then skipping straight to 148-152. Repros to come, hopefully.

erikrose:
1. Put 152 lines of identical text containing "her" into an indexed file, and build the index.
2. Search for "her", pressing Return.
3. Scroll to the bottom.
4. Type in "her" again, not pressing Return.
5. Scroll to the bottom. Only matches 1-73 are found, and then the footer is shown. There should be 152 in all.

Osmose:
I believe I've found the issue; `previousDataLimit` and `dataOffset` were being updated within `doQuery` when a new query was being typed and the query wasn't long enough to trigger a search. Also `dataOffset` wasn't properly being reset to 0 when a new search was happening.

erikrose:
Goodie: I can't get the lines to skip anymore.

Baddie: Now the lines link to just `//source/#25`, leaving out the tree name and thus not working.

Osmose:
Ah, so nunjucks doesn't support `{% from 'foo.html' import blah with context %}`. Updated to explicitly pass the context variables we need (namely www_root and the tree name).

erikrose:
Yeah, it's safer that way anyway. Otherwise, it's effectively dynamic scoping, which is really hard to reason about in any program of size.

erikrose:
It's still not working for me, even after re-running `make`:
screen shot 2015-04-21 at 16 49 38

erikrose:
r+!

PR #411, opened 2015-04-03
Add `dxr shell` command.

Osmose:
`dxr shell` opens an interactive Python shell within a Flask app
context. It essentially does what `dxr serve` does, but instead of
running the app, gives you a shell to play around with.

I found it useful while trying to execute some code copied from a view, which required the app context to be set up or access to a config file.

### DIFF
--- a/dxr/cli/shell.py
+++ b/dxr/cli/shell.py
@@ -1,0 +1,49 @@
+from click import command, option
+
+from dxr.app import make_app
+from dxr.cli.utils import config_option
+
+
+def _plain_shell(namespace):
+    """Start a plain python shell."""
+    import code
+    import readline
+    import rlcompleter
+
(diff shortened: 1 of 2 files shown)


Osmose:
@erikrose r?

erikrose:
r+

dxr/cli/shell.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@command()
+@config_option
+@option('--plain', '-p',
+        is_flag=True,
+        default=False,
+        help='Use a plain shell instead of IPython if available.')
+def shell(config, plain):
+    """Run a Python interactive interpreter in the Flask app context."""
```
erikrose: Oh, you might want to make this a little shorter (and shunt the rest onto other lines) so `dxr --help` looks good. It truncates long (and even not-so-long) lines.

PR #414, opened 2015-05-03
Python plugin fixes

nickolay:
I tried running dxr on itself with the python plugin <https://github.com/mozilla/dxr/pull/362> enabled. Some of these changes were needed to get it to run, and others I made while figuring out how it was supposed to work.

The detailed comments are in the individual commits.

### DIFF
--- a/dxr/plugins/python/analysis.py
+++ b/dxr/plugins/python/analysis.py
@@ -227,5 +227,6 @@ class AnalyzingNodeVisitor(ast.NodeVisitor, ClassFunctionVisitorMixin):
                     package_path = package_for_module(self.abs_module_name)
                     if package_path:
                         abs_import_name = (package_path, alias.name)
-            logger.debug("import (%s,%s) -> %s", self.abs_module_name, local_name, abs_import_name)
-            self.tree_analysis.names[absolute_local_name] = abs_import_name
+            if absolute_local_name != abs_import_name:
+                logger.debug("import (%s,%s) -> %s", self.abs_module_name, local_name, abs_import_name)
+                self.tree_analysis.names[absolute_local_name] = abs_import_name
--- a/dxr/plugins/python/analysis.py
+++ b/dxr/plugins/python/analysis.py
@@ -59,8 +59,8 @@ class TreeAnalysis(object):
             self.ignore_paths.add(rel_path)
             return
 
-        module_path = path_to_module(self.python_path, path)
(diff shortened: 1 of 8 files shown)


erikrose:
These look like great fixes. Taking a look now.

erikrose:
Ha, know what it is? It's the info() ones. (Try `nosetests` in `python/tests/test_overrides`.) I bet if we lower them to debug()s, this whole issue goes away.

erikrose:
Hmm, nope.

erikrose:
(Btw, you're certainly correct that there's no "first" method in the FileToIndex contract. Thank you for reading the spec so closely; most people don't. :-))

I think I'm going to rebase out the logging commit for now, and if you want to figure out a way to keep debug()s quiet during tests, we can land them afterward. It would certainly be nice to be able to have them in there. Or, down the road, we could implement a more flexible status-indication system, one that plays more nicely with the multiprocess nature of plugins.

nickolay:
Thanks for finishing this up for me, you're fantastic!

Re logging -- you were right. Nose sets debug logging, and the reason I didn't see it in test_imports was the difference between how `DxrInstanceTestCase` and `SingleFileTestCase` work:

https://github.com/mozilla/dxr/blob/a994140ce557e1147ef7c75d5d297543ab5fe6a8/dxr/testing.py#L169
vs
https://github.com/mozilla/dxr/blob/a994140ce557e1147ef7c75d5d297543ab5fe6a8/dxr/testing.py#L208

The former runs `dxr index` in a separate process, the output of which is swallowed. The latter runs in-process and prints debugging noise. Thanks for catching that! I agree that logging should not be added until this is dealt with.

erikrose:
You're fantastic yourself! :-) It's always a treat to get such well-thought-through patches—and with tests to boot. And good sleuthing figuring out why test_imports is quiet.

Some background in case you're interested:

The behavior of DxrInstanceTestCase comes from a time when indexing was kicked off by a tree-specific makefile. Nowadays, we could easily change to an in-process model, but that actually makes things even worse for logging. Ideally, I suppose we'd have some kind of `multiprocessing` conduit back to the master process for logging—maybe put a logger-like object on TreeToIndex and/or FileToIndex that you could send messages to. At the moment, everything uses concurrent.futures, but that was really just for convenience and isn't the most efficient thing to do—we send large pickled objects repeatedly because of it—so I mean to eventually transition to lower-level constructs like Queues.

PR #415, opened 2015-05-07
WIP: Adding docker support for DXR dev, and customized elasticsearch

jamonation:
Here's a work in progress pull request that is much more inline with how I'd imagine folks will use docker for DXR dev. It does not yet include a standalone/fully baked image with Apache etc. The intent is to polish this up for dev work, and then create a separate image for a deployable version of DXR.

This PR includes makefile targets to generate images, and to provision a dxr virtual environment in an externally mounted volume using docker-compose.yml. Each image can also be run manually.
 
TODO: 
- [ ] update readme with instructions
- [ ] figure out what to do with peep.py and requirements.txt copies in dockerfiles/dxr directory
- [ ] add a deployable dxr image, e.g. publish to docker hub

### DIFF
--- a/dockerfiles/docker-compose.yml.template
+++ b/dockerfiles/docker-compose.yml.template
@@ -1,0 +1,10 @@
+elasticsearch:
+  image: mozilla/dxr_elasticsearch
+  net: host
+  
+dxr:
+  image: mozilla/dxr_application
+  net: host
+  links:
+    - elasticsearch
+  volumes:
--- a/dockerfiles/dxr/Dockerfile
+++ b/dockerfiles/dxr/Dockerfile
@@ -1,0 +1,35 @@
+FROM ubuntu
+MAINTAINER Jamon Camisso <jamon@jamon.ca>
+
+# Don't prompt for input
+ENV DEBIAN_FRONTEND noninteractive
+
+# Install packages all at once so as not to create a bunch of extra docker checkpoints
+RUN apt-get -q update && apt-get -q -y install python-pip npm sphinx-common libsqlite3-dev sqlite3 git mercurial llvm-3.5 libclang-3.5-dev clang-3.5 pkg-config && apt-get clean
+
+# Install old pip based packages using apt versions
+RUN apt-get -q -y install python-virtualenv python-hglib python-nose && apt-get clean
+
+RUN apt-get -q -y install python-dev && apt-get clean
+
+# Install llvm-config alternative
+RUN update-alternatives --force --install /usr/local/bin/llvm-config llvm-config /usr/bin/llvm-config-3.5 0
+# There is no clang++ until we do this:
+RUN update-alternatives --force --install /usr/local/bin/clang++ clang++ /usr/bin/clang++-3.5 0
+# And we might as well make a clang link so we can compile mozilla-central:
+RUN update-alternatives --force --install /usr/local/bin/clang clang /usr/bin/clang-3.5 0
+
+# Make 'node' symlink
+RUN ["/bin/ln", "-sf", "/usr/bin/nodejs", "/usr/local/bin/node"]
+
+COPY setup_dxr.sh /usr/local/bin/setup_dxr
+
+RUN useradd -u 1000 --create-home --home-dir /home/dxr dxr
+USER dxr
+WORKDIR /home/dxr
+
+ENV VIRTUAL_ENV /home/dxr/venv
+ENV PATH /home/dxr/venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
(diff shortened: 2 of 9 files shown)


dockerfiles/dxr/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM ubuntu
+MAINTAINER Jamon Camisso <jamon@jamon.ca>
+
+# Don't prompt for input
+ENV DEBIAN_FRONTEND noninteractive
+
+# Install packages all at once so as not to create a bunch of extra docker checkpoints
+RUN apt-get -q update && apt-get -q -y install python-pip npm sphinx-common libsqlite3-dev sqlite3 git mercurial llvm-3.5 libclang-3.5-dev clang-3.5 pkg-config && apt-get clean
```
erikrose: We can take out the sqlite stuff.

erikrose:
Thanks for redoing this, jamon! :-D

erikrose:
> The intent is to polish this up for dev work, and then create a separate image for a deployable version of DXR.

++ your intent. I think the Apache setup on our vagrant VM is broken right now, so it doesn't make it any worse to not have it for awhile.

makefile
```diff
@@ -4,6 +4,20 @@ BUILD_PLUGINS = $(PLUGINS:%=build-plugin-%)
 CHECK_PLUGINS = $(PLUGINS:%=check-plugin-%)
 CLEAN_PLUGINS = $(PLUGINS:%=clean-plugin-%)
 
+PWD = $(shell pwd)
+
+help:
```
erikrose: Nice idea!

erikrose:
I've been reading up on Docker a bit, and I wonder how we can set things up so that people submitting PRs can't do arbitrary things to the host. For example, https://docs.docker.com/articles/security/#docker-daemon-attack-surface lays out a way of doing arbitrary FS writes using the shared-directory mechanism. Any ideas?

erikrose:
https://wiki.jenkins-ci.org/display/JENKINS/Docker+Plugin exists. I'm not sure how many container security problems it solves, but it seems to plug the FS hole by putting a GUI around the FS Root option, moving the invocation of docker (which happens as root) under control of Jenkins itself rather than allowing it to exist in, say, a jenkins.sh script committed to the repo. Thus, random PR openers couldn't march through that hole, at least not as straightforwardly.

erikrose:
http://circleci.com/ is said to do this as a service. We could let them deal with the security implications. Other Mozilla webdev projects are using it.

jamonation:
I think https://circleci.com/docs/docker should be enough info to get going with this setup. It shows different ways of using Docker to build and test a project by overriding the default circleci commands.

erikrose:
drone.io is also a contender. It doesn't limit you to 1 concurrent build (per user?) on its free tier.

ehostunreach:
What is the status of this pull request? Does it still work with the master branch?

erikrose:
It's not likely that it will merge cleanly with master, as requirements.txt and the makefile have changed significantly in the meantime. However, the concepts are still sound, and it wouldn't be too hard to port.

erikrose:
@jamonation What's the magic here that allows localhost:9200 within the dxr container to actually be 9200 on the elasticsearch container?

dockerfiles/docker-compose.yml.template
```diff
@@ -0,0 +1,10 @@
+elasticsearch:
+  image: mozilla/dxr_elasticsearch
+  net: host
+  
+dxr:
+  image: mozilla/dxr_application
+  net: host
+  links:
```
erikrose: What was `links` for? Everything seems to work fine without for me (in my own version of this that draws liberally from yours).

jamonation: Trying to remember back to when I cooked this up, it's been so long!

IIRC, I was trying to not use `net: host` to keep containers talking among themselves over docker's bridge. But I think for doing dev work from outside the container and for ease of testing, the net:host is the better approach so I went with that. Long way of saying, links is redundant, since net: host will ignore them anyway.

jamonation:
The `net: host` is what binds the Elasticsearch in the container to the docker bridge IP. Otherwise docker assigns each container an IP on its own private 172.16.0.0/12 network and uses NAT to talk to the outside world.

erikrose:
Thanks! Here's what I ended up doing:

```
es:
  build: ./docker_es
  ports:
    - "8000:8000"

dev:
  build: ./
  net: container:es
  ports:
    - "8000:8000"
  volumes:
      - .:/home/dxr/dxr
```

I just had the dev container share es's networking stack and then forwarded port 8080 from dev to es and thence to my docker-machine VM, which I can reach from my host. At some point, docker got grumpy and started refusing to do net:host and links together.

PR #416, opened 2015-05-07
Set "bash strict mode" not only on vagrant_provision but on the subsh…

erikrose:
…ell it spawns via sudo.

https://ci.mozilla.org/job/dxr/551/ plowed ahead even after a peep error because we weren't doing this.

See http://redsymbol.net/articles/unofficial-bash-strict-mode/.

NOTE: This is going to fail on purpose to see if the peep failure promptly aborts the script.


erikrose:
Oh, never mind. We're switching to Docker pretty soon anyway.

PR #419, opened 2015-05-26
Refactored VCS, run file skimmers on requests, permalinks

pelmers:
This work is still a bit rough around the edges, but I wanted to put this up to make it easier to review.

Related bugs:
- https://bugzilla.mozilla.org/show_bug.cgi?id=962108
- https://bugzilla.mozilla.org/show_bug.cgi?id=1166397

Summary of changes:
- Store line endings in ES, and store regions and refs separately. This lets us FileToSkim when files are requested, and merge refs and regions at this point. Increment the format version.
- Add some tests for Git and Mercurial.
- Add permalink button to nav bar along with '/rev' route.
- Add dependence on [hglib](https://pypi.python.org/pypi/python-hglib/) to avoid [MercurialApi](http://mercurial.selenic.com/wiki/MercurialApi)
- Other refactoring and minor fixes.

Still needed:
- ~~Tests for Perforce.~~ Too expensive.
- ~~Put permalink generation in core.py.~~
- ~~Move link building into vcs.py~~
- ~~Construct some kind of VCS manager for each worker that caches path -> vcs mapping.~~
- ~~Update the docs. Which portions?~~
- ~~Better name than VCSTree ?~~
- Squash some of the back-and-forth commits
- Rename omniglot?

PR #421, opened 2015-05-28
Fix bug 1166549. Make binaries not clickable in search results.

pelmers:
Pass is_binary field of FILE to search_html, and gate the anchor tag
in the path_line template on whether the path is to a binary file.

### DIFF
--- a/dxr/static/templates/path_line.html
+++ b/dxr/static/templates/path_line.html
@@ -5,6 +5,6 @@
 <a href="{{ url }}" {%- if is_dir %} data-path="{{ data_path }}" {% endif %}>
 {%- endif -%}
     {{ display_path }}
-{%- if not is_binary -%}
+{%- if not is_binary and not is_dir -%}
 </a>
 {%- endif -%}
--- a/dxr/static/templates/path_line.html
+++ b/dxr/static/templates/path_line.html
@@ -1,4 +1,10 @@
 {%- if not is_first_or_only -%}
 <span class="path-separator">/</span>
 {%- endif -%}
-<a href="{{ url }}" {%- if is_dir %} data-path="{{ data_path }}" {% endif %}>{{ display_path }}</a>
+{%- if not is_binary and not is_dir -%}
+<a href="{{ url }}" {%- if is_dir %} data-path="{{ data_path }}" {% endif %}>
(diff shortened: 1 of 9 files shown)


dxr/static/templates/path_line.html
```diff
@@ -1,4 +1,10 @@
 {%- if not is_first_or_only -%}
 <span class="path-separator">/</span>
 {%- endif -%}
-<a href="{{ url }}" {%- if is_dir %} data-path="{{ data_path }}" {% endif %}>{{ display_path }}</a>
\ No newline at end of file
+{%- if not is_binary and not is_dir -%}
+<a href="{{ url }}" {%- if is_dir %} data-path="{{ data_path }}" {% endif %}>
+{%- endif -%}
+    {{ display_path }}
+{%- if not is_binary -%}
```
erikrose: Since this condition isn't the same as the one enclosing the `<a>`, are there any situations in which we might end up printing the close tag without the opener?

tests/test_binary_files/test_binary_files.py
```diff
@@ -43,3 +43,10 @@ def test_some_bytes(self):
         ok_('some_bytes' in response.data)
         response = self.client().get('/code/source/some_bytes')
         ok_(response.status_code, 404)
+
+    def test_some_bytes(self):
+        """We want some_bytes to show on search page, but not be clickable."""
+        response = self.client().get('/code/search?q=path%3Asome_bytes&redirect=true')
+        ok_('some_bytes' in response.data)
```
erikrose: I love how far we've gotten without actually parsing anything. ;-)

dxr/static/templates/path_line.html
pelmers: My mistake! We'd have an extra closing tag for all of the path crumbs leading up to the file.

erikrose:
It looks great when you search by pressing Return, but it doesn't take effect when you just wait 300ms after typing the query. You may have to edit search.html as well. And if the repetition gets to be too much (and it's probably getting there), there's always https://bugzilla.mozilla.org/show_bug.cgi?id=970405 to consider.

erikrose:
r+! We should probably merge #419 in first, though.

PR #422, opened 2015-06-01
Empty out search.html; perform AJAX request instead.

pelmers:
Addresses bug [970405](https://bugzilla.mozilla.org/show_bug.cgi?id=970405).
~~Still a bit rough; sometimes when loading a page the location will change after about half a second.~~ Fixed by #423.

Here we remove the content block of search.html and instead perform an AJAX request on load of the search page based on the search portion of the location.
Remark: I do not completely remove search.html because I want to preserve the back button and the ability to link directly to some search results.

### DIFF
--- a/dxr/static/js/dxr.js
+++ b/dxr/static/js/dxr.js
@@ -7,7 +7,6 @@ $(function() {
     'use strict';
 
     var constants = $('#data');
-    var stateConstants = $('#state');
     var dxr = {},
         docElem = document.documentElement;
 
@@ -175,7 +174,7 @@ $(function() {
      * @param {string} query - The query string
      * @param {bool} isCaseSensitive - Whether the query should be case-sensitive
      * @param {int} limit - The number of results to return.
-     * @param [int] offset - The cursor position
+     * @param {int} offset - The cursor position
      */
     function buildAjaxURL(query, isCaseSensitive, limit, offset) {
         var search = dxr.searchUrl;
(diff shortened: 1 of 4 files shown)


pelmers:
Strange behavior: if you close a tab, and then re-open it, then the page shows a big JSON dump. Hitting refresh solves the issue. This occurs in Chrome and Firefox.

erikrose:
That is weird. IHNI. It happens both for Return-key-pressed results and non.

erikrose:
Unsurprisingly, when you hit Return, there's a little flash of blank-search-results before the AJAX pulls them in. It's trivial locally, but I worry it'll be more noticeable on real servers. Why don't we intercept the Return-key press rather than actually submitting the form and just substitute the results onto the existing page?

pelmers:
If we intercept the form we will still have a delay before the new results are shown. In fact I think it would be pretty much the same delay, since it would be sending the same request. Do you think it would just be less noticeable?

erikrose:
The difference would be that the search bar itself wouldn't flash: there wouldn't be that appearance of a reload without actually being complete. And we can and should activate the spinner while the AJAX request is in flight.

erikrose:
Obviously, we'll need some special code to handle redirects, but it should be trivial.

pelmers:
Updated, I made the change by responding with 'redirect: {location}' when a direct hit is found.

erikrose:
The docstrings for _search_json() and _search_html() need to be updated.

dxr/static/js/dxr.js
```diff
@@ -423,19 +408,32 @@ $(function() {
...
+                if (data.redirect) {
+                    window.location.href = data.redirect;
+                    return;
+                }
+                data.query = query;
+                // New results, overwrite
+                if (myRequestNumber > displayedRequestNumber) {
+                    displayedRequestNumber = myRequestNumber;
+                    populateResults(data, false);
+                    historyWaiter = setTimeout(history.pushState.bind(history, {}, '', queryString),
```
erikrose: If the user hit Return (i.e. redirect is true), let's update the URL immediately instead of waiting.

erikrose:
Otherwise, looks really good. This'll make our code simpler with a break-even effect on UX (piecewise load for hits to search.html from scratch but no more page flash when hitting Return in the search field).

erikrose:
r+ after you make that last tweak. No need to go through another round.

PR #423, opened 2015-06-08
Fixes bug 1171229. Correctly update title on dynamic searches.

pelmers:
Attach the contents of the query field to the data parameter to populateResults, so the document title updates correctly.


erikrose:
r+! And let's add a space while we're at it: "- DXR Search" → " - DXR Search". (Also, that should totally be an em dash—but that's extra credit.)

PR #427, opened 2015-06-09
Get elasticsearch.org's GPG key over HTTPS.

erikrose:
It's not perfect, but it's better than cleartext.


erikrose:
Merged in 3413ad4637149559584c34f21662364495dd4792.

PR #428, opened 2015-06-09
Support id and ref filters.

pelmers:
These filters `or` together the filters of enabled plugins that either
find identifiers (i.e. definitions and declarations) or references to
them.
This work is a part of the query language refresh, and a lays groundwork
for result mixing.

To implement this I change the Filter constructor to take a list of enabled plugins, and I add properties is_reference and is_identifier which filters set to True when applicable.

### DIFF
--- a/dxr/filters.py
+++ b/dxr/filters.py
@@ -43,16 +43,23 @@ class Filter(object):
     """
     domain = LINE
     description = u''
+    is_reference = False
+    is_identifier = False
 
-    def __init__(self, term):
+    def __init__(self, term, enabled_plugins):
         """This is a good place to parse the term's arg (if it requires further
         parsing) and stash it away on the instance.
+
+        :arg term: a query term as constructed by a :class:`~dxr.query.QueryVisitor`
+        :arg enabled_plugins: an iterable of the enabled :class:`~dxr.plugins.Plugin` instances,
+            for use by filters that build upon the filters provided by plugins
 
         Raise :class:`~dxr.exceptions.BadTerm` to complain to the user: for
         instance, about an unparseable term.
 
         """
         self._term = term
+        self._enabled_plugins = enabled_plugins
 
     def filter(self):
         """Return the ES filter clause that applies my restrictions to the
@@ -129,8 +136,8 @@ class NameFilterBase(Filter):
         familiar.
 
     """
-    def __init__(self, term):
-        super(NameFilterBase, self).__init__(term)
+    def __init__(self, term, enabled_plugins):
+        super(NameFilterBase, self).__init__(term, enabled_plugins)
         self._needle = '{0}_{1}'.format(self.lang, self.name.replace('-', '_'))
 
     def _term_filter(self, field):
@@ -182,12 +189,7 @@ class NameFilterBase(Filter):
         return maybe_lower(entity['name']) == maybe_lower(self._term['arg'])
 
     def highlight_content(self, result):
-        """Highlight any structural entity whose name matches the term.
-
-        Compare short names and qualnames if this isn't a fully-qualified
-        search. Otherwise, compare just qualnames.
-
-        """
+        """Highlight any structural entity whose name matches the term."""
         if self._term['not']:
             return []
         return ((entity['start'], entity['end'])
--- a/dxr/plugins/clang/filters.py
+++ b/dxr/plugins/clang/filters.py
@@ -13,11 +13,13 @@ class _CNameFilter(NameFilterBase):
 
 class FunctionFilter(_CQualifiedNameFilter):
     name = 'function'
+    is_identifier = True
     description = Markup('Function or method definition: <code>function:foo</code>')
 
 
 class FunctionRefFilter(_CQualifiedNameFilter):
     name = 'function-ref'
+    is_reference = True
     description = 'Function or method references'
 
 
@@ -28,6 +30,7 @@ class FunctionDeclFilter(_CQualifiedNameFilter):
 
 class TypeRefFilter(_CQualifiedNameFilter):
     name = 'type-ref'
+    is_reference = True
     description = 'Type or class references, uses, or instantiations'
 
 
@@ -38,16 +41,19 @@ class TypeDeclFilter(_CQualifiedNameFilter):
 
 class TypeFilter(_CQualifiedNameFilter):
     name = 'type'
+    is_identifier = True
     description = Markup('Type or class definition: <code>type:Stack</code>')
 
 
 class VariableFilter(_CQualifiedNameFilter):
     name = 'var'
+    is_identifier = True
     description = 'Variable definition'
 
 
 class VariableRefFilter(_CQualifiedNameFilter):
     name = 'var-ref'
+    is_reference = True
     description = 'Variable uses (lvalue, rvalue, dereference, etc.)'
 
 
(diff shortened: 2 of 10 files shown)


dxr/filters.py
```diff
@@ -43,8 +43,10 @@ class Filter(object):
     """
     domain = LINE
     description = u''
+    is_reference = False
+    is_identifier = False
 
-    def __init__(self, term):
+    def __init__(self, term, enabled_plugins):
```
erikrose: We should add to the docstring an `:arg enabled_plugins` line that outlines the intended use of the plugin list by filters. (And we might as well document `term` as well, so it doesn't feel left out.)

dxr/filters.py
```diff
@@ -182,12 +185,7 @@ def _should_be_highlit(self, entity):
         return maybe_lower(entity['name']) == maybe_lower(self._term['arg'])
 
     def highlight_content(self, result):
-        """Highlight any structural entity whose name matches the term.
-
-        Compare short names and qualnames if this isn't a fully-qualified
-        search. Otherwise, compare just qualnames.
-
-        """
+        """Highlight any structural entity whose name matches the term."""
```
erikrose: Good catch!

dxr/query.py
```diff
@@ -326,6 +327,62 @@ def generic_visit(self, node, visited_children):
         return visited_children or node
 
 
+# TODO next next: write some tests
+class FilterAggregator(Filter):
+    """Filter class that acts by constructing the union of some subset of the
+    filters of all currently enabled plugins.
+
+    """
+
```
erikrose: Skipping a line within the docstring means you don't have to afterward.

dxr/plugins/clang/filters.py
```diff
@@ -100,9 +115,9 @@ class CallerFilter(_CQualifiedNameFilter):
     name = 'callers'
     description = Markup('Functions which call the given function or method: <code>callers:GetStringFromName</code>')
 
-    def __init__(self, term):
+    def __init__(self, term, enabled_plugins):
```
erikrose: We call this just `plugins` down in FilterAggregator. Let's choose one or the other.

dxr/query.py
```diff
@@ -326,6 +327,62 @@ def generic_visit(self, node, visited_children):
...
+    def __init__(self, term, plugins, condition=None):
```
erikrose: I like that you made `condition` part of the constructor. There's no reasonable default, and this makes sure you can't leave it out.

dxr/query.py
```diff
@@ -326,6 +327,62 @@ def generic_visit(self, node, visited_children):
...
+        return chain.from_iterable(
+            f.highlight_content(result) for f in self.filters)
+
+
+class IdFilter(FilterAggregator):
+    """Case-sensitive filter for exact matching on declarations."""
+
+    name = 'id'
+    domain = LINE
+    description = Markup('Definition: <code>id:var</code>.')
```
erikrose: I think you forgot to describe the meaning of the filter (here and in RefFilter).

dxr/query.py
```diff
@@ -326,6 +327,62 @@ def generic_visit(self, node, visited_children):
...
+        return {'or': filter(None, (f.filter() for f in self.filters))}
+
+    def highlight_content(self, result):
+        # Union all of our underlying filters.
+        return chain.from_iterable(
+            f.highlight_content(result) for f in self.filters)
+
+
+class IdFilter(FilterAggregator):
+    """Case-sensitive filter for exact matching on declarations."""
```
erikrose: This should (and probably does) obey the case-sensitive checkbox, rather than being always one way or the other.

dxr/query.py
```diff
@@ -326,6 +327,62 @@ def generic_visit(self, node, visited_children):
...
+    name = 'ref'
+    domain = LINE
+    description = Markup('Definition: <code>ref:var</code>.')
+
+    def __init__(self, term, plugins):
+        super(RefFilter, self).__init__(term, plugins, lambda f: f.is_reference)
+
+
+# TODO: cache this
+def some_filters_by_name(plugins, condition):
```
erikrose: I guess this really isn't by name [anymore].

dxr/query.py
```diff
@@ -337,7 +394,8 @@ def filters_by_name(plugins):
     return append_update(
         {},
         ((f.name, f) for f in
-         chain.from_iterable(p.filters for p in plugins)))
+         chain(chain.from_iterable(p.filters for p in plugins),
+               (IdFilter, RefFilter))))
```
erikrose: Why can't we register these as normal filters in dxr.plugins.core?

tests/test_filter_aggregates/code/main.c
```diff
@@ -0,0 +1,13 @@
+#include <stdio.h>
```
erikrose: This would be a good candidate for a SingleFileTestCase.

dxr/query.py
```diff
@@ -326,6 +327,62 @@ def generic_visit(self, node, visited_children):
         return visited_children or node
 
 
+# TODO next next: write some tests
```
erikrose: It looks like you did this!

dxr/query.py
```diff
@@ -326,6 +327,62 @@ def generic_visit(self, node, visited_children):
...
+
+    name = 'ref'
+    domain = LINE
+    description = Markup('Definition: <code>ref:var</code>.')
+
+    def __init__(self, term, plugins):
+        super(RefFilter, self).__init__(term, plugins, lambda f: f.is_reference)
+
+
+# TODO: cache this
```
erikrose: I'm not even sure how you're going to cache this: it'll be passed a new lambda each time.

erikrose: Maybe YAGNI and we should wait to see benchmarks show this is a slow thing.

dxr/plugins/clang/filters.py
pelmers: I'll go with enabled_plugins, since that's what they are. We can rename it if we later decide to do things differently.

erikrose:
I'll give this a try tomorrow. I want to see how it performs on a full moz-central index before merging it in. (If you have one, you can give it a swing yourself.)

pelmers:
I'm not sure how to build the full documentation. When I run `make html` from the docs directory, I can browse through all the .rst files in source, but I can't find the Python code documentation.

erikrose:
After running `make html`, look in docs/build/html.

pelmers:
My section 6.6.5 is empty, unlike https://dxr.readthedocs.org/en/es/development.html?highlight=filter#dxr.filters.Filter. I only have the Filter section header, but there are no documentation contents under it (same story for 6.6.2 TreeIndexer).

erikrose:
Do you see *any* of the documentation that's generated from code, or is it all missing? If the latter, it could be something to do with the PYTHONPATH. FWIW, it works for me in my vagrant box.

pelmers:
Ah, I wasn't building it on the vagrant box; I switched to that and it worked. And you're right about the `s thing, it didn't link properly on the doc page.

erikrose:
r+

Benchmarks from pelmers:
* Search id:kungFuDeathGrip took 0.07069 seconds
* Search id:webSocket took 0.02859 seconds
* Search id:dom took 0.10922 seconds (6000 + results)
* Search function:SetAlarm took 0.03684 seconds

We can easily get away with the OR queries.

erikrose:
As requested, I've audited the choices of identifiers and references. I think we should…
* Make decls (in clang) no longer identifiers. I'd want to find the function source code, not just its prototype.
* Make callers (in clang, python, and rust) references. (They're references that also happen to call.)
* I strongly suspect ModuleAliasRefFilter should be a reference.
* I think ImplFilter should be a reference, too.

pelmers:
It looks like we'll have to finish up #434 first so that some of the tests added in this commit will pass, now that callers are combined into the ref filter.

erikrose:
lgtm. r+ again!

PR #429, opened 2015-06-10
Use pkg_resources.resource_filename to find hgext.

pelmers:
When using setup.py install, the hg extension was not installed because it was not a Python package and it was not specified in the Manifest.
This patch adds the folder to the Manifest, and we refer to the filename in code using pkg_resources, so it will be extracted to a temporary cache directory if it's stuck in an archive.

### DIFF
--- a/dxr/vcs.py
+++ b/dxr/vcs.py
@@ -18,15 +18,13 @@ Todos:
 import marshal
 import os
 from os.path import relpath, join, dirname
+from pkg_resources import resource_filename
 import subprocess
 import urlparse
 from warnings import warn
 
-
 import hglib
 from ordereddict import OrderedDict
-
-import dxr
 
 class Vcs(object):
     """A class representing an abstract notion of a version-control system.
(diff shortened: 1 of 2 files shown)


dxr/vcs.py
```diff
@@ -84,9 +84,9 @@ def display_rev(self, path):
 class Mercurial(Vcs):
     def __init__(self, root):
         super(Mercurial, self).__init__(root, 'hg')
-        hgext = join(dirname(dxr.__file__), 'hgext', 'previous_revisions.py')
+        hgext = resource_filename(__name__, 'hgext/previous_revisions.py')
```
erikrose: We probably shouldn't use `__name__` here. Take a look at https://pythonhosted.org/setuptools/pkg_resources.html#basic-resource-access. This becomes "dxr.vcs", which fortunately works because `vcs` is a module. But someday, when `vcs` grows and becomes a package, this will quit working. The most durable thing would probably be to import dxr and then use `dxr.__name__` here. Or we could just hard-code 'dxr', which would work unless we renamed the whole project.

erikrose:
r+

PR #430, opened 2015-06-11
Fix condition for whether to add path anchor.

pelmers:
If it’s not binary, definitely add path anchor.
Otherwise, add the anchor if it is a directory.
i.e. not is_binary OR is_dir.


erikrose:
r+

PR #432, opened 2015-06-13
Bug fixes and PEP-8 conformance in vcs.py

pelmers:
The main issue is that find_previous_revisions was rather delicate and would fail the entire deploy if hg hiccups at all. Now the function should continue regardless of what `hg previous-revisions` prints. 

I changed a few style nits while I was in there, and fixed an extra `=` in the hg upstream url handling code.


erikrose:
What kind of hiccups was hg having?

pelmers:
Looking at http://git.io/vIpWn, the only ways I can see the code failing are if a filename contains a colon, or if a filename contains a line break, so "hg hiccup" should instead be "unexpected filenames."

dxr/vcs.py
```diff
@@ -113,8 +114,9 @@ def find_previous_revisions(self, client):
         """
         last_change = {}
         for line in client.rawcommand(['previous-revisions']).splitlines():
-            path, node = line.split(':')
-            last_change[path] = node
+            if ':' in line:
+                path, node = line.split(':', 1)
+                last_change[path] = node
```
erikrose: If there is a newline in a filename, this is still going to make a hash entry. I think this fix would benefit from a test or 3.

erikrose: Actually, we could probably avoid this whole thing by having previous_revisions print repr(filename) and then using https://docs.python.org/2/library/ast.html#ast.literal_eval to read it back in.

pelmers: Good point on adding tests! I've learned that hg does not allow tracking files containing line breaks, so that first check isn't necessary. However I'm having trouble with paths containing a colon since currently I use that as the marker between the two parts.

erikrose:
r+

PR #434, opened 2015-06-15
Fix closing extent on the call filter.

pelmers:
For call expressions, Clang gives us the span of the entire foo(x,y,z)
expression, except the closing parenthesis. Subtracting one from the
span ends up cutting off the last two character of the argument list and
the closing parenthesis. We manually add one back to get the right span.

### DIFF
--- a/dxr/plugins/clang/tests/test_callers.py
+++ b/dxr/plugins/clang/tests/test_callers.py
@@ -11,7 +11,7 @@ class DirectCallTests(SingleFileTestCase):
         {
         }
 
-        void called_once()
+        void called_once(int a)
         {
         }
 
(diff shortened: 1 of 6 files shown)


dxr/plugins/clang/tests/test_callers.py
```diff
@@ -36,15 +36,15 @@ def test_no_caller(self):
         self.found_nothing('callers:orphan')
 
     def test_one_caller(self):
-        self.found_line_eq('callers:called_once', '<b>called_once</b>();')
+        self.found_line_eq('callers:called_once', '<b>called_once(5)</b>;')
```
erikrose: Let's add to the docstring here that we're also testing the highlighting of arguments. Do that, and r+!

PR #435, opened 2015-06-15
Bring back flag to allow post-test interaction.

pelmers:
Revive the flag should_delete_instance as stop_for_interaction, which
allows hosting a single-file test case after running tests for
interaction via web browser. Update the docs appropriately.

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -148,9 +148,9 @@ DXR supports two kinds of integration tests:
    stores the code as a Python string within a subclass of
    ``SingleFileTestCase``. At test time, it instantiates the file on
    disk in a temp folder, builds it, and makes assertions about it. If
-   the ``should_delete_instance`` class variable is truthy (the default), it
-   then deletes the instance. If you want to examine the instance manually for
-   troubleshooting, set this to ``False``.
+   the ``stop_for_interaction`` class variable is falsy (the default), it
+   then deletes the index. If you want to browse the instance manually for
+   troubleshooting, set this to ``True``.
 
 2. A heavier sort of test: a folder containing one or more source trees and a
    DXR config file. These are useful for tests that require a multi-file tree
--- a/dxr/testing.py
+++ b/dxr/testing.py
@@ -222,11 +222,9 @@ class SingleFileTestCase(TestCase):
     then kick off the usual build process, deleting the instance afterward.
 
     """
-    # Set this to False in a subclass to keep the generated instance around and
-    # print its path so you can examine it:
-    # Note: this is currently broken because no on-disk artifiact is actually
-    # created; everything goes into Elasticsearch directly.
-    should_delete_instance = True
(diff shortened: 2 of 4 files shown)


dxr/testing.py
```diff
@@ -267,11 +265,12 @@ def config_input(cls, config_dir_path):
 
     @classmethod
     def teardown_class(cls):
-        if cls.should_delete_instance:
+        if cls.stop_for_interaction:
+            print 'Not deleting instance in %s.' % cls._config_dir_path
+            make_app(cls.config()).run(host='0.0.0.0', port=8000)
```
erikrose: Can we import and call dxr.cli.serve.serve() instead, so we don't have to repeat the defaults (like 8000)? Also, we should remove the language here and in the .rst about deleting instances. There's no such thing as an instance anymore. (It used to be a folder full of HTML documents and a SQLite DB.)

docs/source/development.rst
```diff
@@ -148,9 +148,9 @@ DXR supports two kinds of integration tests:
    stores the code as a Python string within a subclass of
    ``SingleFileTestCase``. At test time, it instantiates the file on
    disk in a temp folder, builds it, and makes assertions about it. If
-   the ``should_delete_instance`` class variable is truthy (the default), it
+   the ``stop_for_interaction`` class variable is falsy (the default), it
    then deletes the instance. If you want to examine the instance manually for
```
erikrose: For example, we could say "deletes the index" and "If you want to browse the index manually", and it'd be accurate.

dxr/testing.py
pelmers: I'm getting an error "Command has no attribute serve".

pelmers: Silly me, I wrote an extra serve there.
```
from dxr.cli import serve
serve()
```

dxr/testing.py
```diff
@@ -267,11 +265,12 @@ def config_input(cls, config_dir_path):
 
     @classmethod
     def teardown_class(cls):
-        if cls.should_delete_instance:
+        if cls.stop_for_interaction:
+            print "Pausing for interaction at 0.0.0.0:8000..."
```
pelmers: Funnily, I thought that Flask would tell us when it launches the server, but it didn't in this case.

erikrose: We should wrap this whole thing in an `except KeyboardInterrupt` that catches ctrl-C and then cleans up the indices before exiting. I just noticed that, if an index is referenced in the catalog index but it isn't in the config file, then we throw 500s when trying to view files from it, since config.trees[the tree] in _browse_file fails. (We should also harden that, but that's another bug.)

pelmers: Flask already catches ctrl-c and returns cleanly from make_app.run(), so I removed the else clause to allow fall-through to the cleanup code.

erikrose:
Sweet. r+

PR #436, opened 2015-06-16
Do not build any vcs_cache in make_app.

pelmers:
Make get_contents of Vcs a classmethod, and when we request a specific
revision just try to invoke each vcs to get that contents.
This address bug [1175137](https://bugzilla.mozilla.org/show_bug.cgi?id=1175137).

### DIFF
--- a/dxr/vcs.py
+++ b/dxr/vcs.py
@@ -17,7 +17,7 @@ Todos:
 """
 import marshal
 import os
-from os.path import relpath, join, dirname
+from os.path import relpath, join, split
 from pkg_resources import resource_filename
 import subprocess
 import urlparse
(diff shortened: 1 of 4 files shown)


dxr/vcs.py
```diff
@@ -303,6 +315,17 @@ def tree_to_repos(tree):
...
 
+def file_contents_at_rev(abspath, revision):
+    """Attempt to return the contents of a file at a specific revision."""
+
+    for cls in [Mercurial, Git]:
+        try:
+            return cls.get_contents(abspath, revision)
+        except subprocess.CalledProcessError:
+            continue
+    return None
```
erikrose: Just FYI, this isn't necessary.

erikrose:
r+. Let's give it a try.

PR #437, opened 2015-06-18
Fix Rust indexing

nrc:


dxr/config.py
```diff
@@ -105,7 +105,7 @@ def __init__(self, input, relative_to=None):
                     AbsPath,
                 Optional('workers', default=if_raises(NotImplementedError,
                                                       cpu_count,
-                                                      1)):
+                                                      1) / 2):
```
erikrose: That's going to cause interesting behavior when cpu_count() is 1 (like on the default vagrant VM), since 1 / 2 == 0. Also, VMs assigned <= the physical number of CPUs on the host will be unjustly penalized by this. We could start using https://pypi.python.org/pypi/psutil/, which can report the number of physical cores, but that's a big dependency with a platform-specific component and a compilation step for a fairly small convenience. I'm also not sure that setting `workers` to include the logical cores isn't a help during clang indexing, when some processes sleep while sending data to ES. (I notice my VM is currently set to physical+logical cores, presumably for a reason.) Do you think a note in the documentation would suffice? Do you think depending on a 200K library is worth it? (This may be clearer to me after lunch.)

dxr/plugins/rust/__init__.py
```diff
@@ -436,8 +436,13 @@ def process_csv(self, file_name, header_only):
                         print " - 'process_" + line[0] + "' not implemented!"
                         continue
 
-                    if 'file_name' in args and args['file_name'].startswith(self.tree.source_folder):
-                        args['file_name'] = args['file_name'][len(self.tree.source_folder)+1:]
+                    if 'file_name' in args:
+                        file_name_value = os.path.normpath(args['file_name'])
```
erikrose: Did you consider https://docs.python.org/2/library/os.path.html#os.path.relpath? I think it does all this for you.

nrc: I'm not sure if it does - the docs don't say anything about normalisation, so I'm not sure if it does that and if it does, if it can be relied on. It also doesn't say what would happen if start is not part of path - will it error out or return the absolute path? I specifically want the latter behaviour.

dxr/config.py
nrc: I'm going to move this to a separate PR in the interests of landing the bug fix quickly.

I don't think it is worth making another dependency for. It's only a default and I think we can come up with a good heuristic. My VM is also set to have cores = physical + logical, but if the OS then tries to really use that many threads, it really seizes up the machine. It might be worth experimenting with the Clang plugin to find out what the best number of cores is, rather than trying to guess. For Rust going from 8 to 4 workers meant 90-ish to 50-ish minutes, which is a massive saving. Lets chat about this later and I'll send in a PR when we figure out what to do.

nrc:
Pushed just the bug fix.

dxr/plugins/rust/__init__.py
erikrose: It does normalization. Here's the source, which probably hasn't changed for 10 years:

```python
def relpath(path, start=curdir):
    """Return a relative version of a path"""

    if not path:
        raise ValueError("no path specified")

    start_list = [x for x in abspath(start).split(sep) if x]
    path_list = [x for x in abspath(path).split(sep) if x]

    # Work out how much of the filepath is shared by start and path.
    i = len(commonprefix([start_list, path_list]))

    rel_list = [pardir] * (len(start_list)-i) + path_list[i:]
    if not rel_list:
        return curdir
    return join(*rel_list)
```

If `start` is not part of `path`, it doesn't error out, but it doesn't return an abspath either: it returns a relpath, like `../../../a/b` for `relpath('/a/b', start='/home/vagrant/src')`. Will that do?

dxr/config.py
erikrose: Sounds good. Looking forward to seeing you in Whistler! :-D

dxr/plugins/rust/__init__.py
nrc: I don't think that does what I want because the paths might be relative paths, but not relative to the current working directory, then I don't think the paths will match up in the relative case.

PR #438, opened 2015-06-20
Updated Getting Started docs about trilite

coldmind:
Someone will need to set `LD_LIBRARY_PATH` env variable to make commands work.


erikrose:
Thanks for the patch! However, we're about to merge a big branch (es, if you want to try it out) into master which will obviate the need for trilite altogether, so I think we'll just wait for that.

PR #439, opened 2015-06-24
Fix bug 1176098, infinite scroll with file results.

pelmers:
Paths that contain no lines are results too, so make sure to add 1 to the result count.

https://bugzilla.mozilla.org/show_bug.cgi?id=1176098

### DIFF
--- a/dxr/static/js/dxr.js
+++ b/dxr/static/js/dxr.js
@@ -232,6 +232,8 @@ $(function() {
                 threshold = window.innerHeight + 500;
 
             // Has the user reached the scrolling threshold and are there more results?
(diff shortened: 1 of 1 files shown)


dxr/static/js/dxr.js
```diff
@@ -270,7 +270,7 @@ $(function() {
     function countLines(results) {
         var total = 0;
         for (var k = 0; k < results.length; k++) {
-            total += results[k].lines.length;
+            total += Math.max(results[k].lines.length, 1);
```
erikrose: At the least, we should rename countLines, because that's not what it does anymore. But this certainly does fix a bug; good job!

erikrose:
r+

PR #440, opened 2015-06-25
Add direct search for exact matches of filenames or trailing path seg…

erikrose:
…ments. Fixes bug 1171313.

* Introduce a new type of direct searcher than can return LINE-domain results.
* Add a new filename direct searcher that uses it.
* Switch test harness to use url_for() rather than hard-coding URLs.
* Change direct_result_eq(..., None) to mean "returns a FILE-domain direct result" rather than "doesn't return a direct result".

### DIFF
--- a/dxr/plugins/clang/tests/test_direct.py
+++ b/dxr/plugins/clang/tests/test_direct.py
@@ -45,8 +45,8 @@ class TypeAndMethodTests(SingleFileTestCase):
         self.direct_result_eq('MemberFunction::member_function(int)', 16)
 
     def test_qualified_function_name_multiple_matches(self):
-        """Multiple matches on fully qualified function name should return
-        None."""
+        """Multiple matches on fully qualified function name should not yield
+        a direct result."""
         self.is_not_direct_result('MemberFunction::member_FUNCTION(int)')
 
     def test_qualified_type_name_insensitive(self):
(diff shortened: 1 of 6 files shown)


dxr/testing.py
```diff
@@ -30,9 +30,14 @@ class TestCase(unittest.TestCase):
...
 
     def client(self):
-        app = make_app(self.config())
-        app.config['TESTING'] = True  # Disable error trapping during requests.
-        return app.test_client()
+        self._app = make_app(self.config())
+        self._app.config['TESTING'] = True  # Disable error trapping during requests.
+        return self._app.test_client()
+
+    def url_for(self, *args, **kwargs):
```
pelmers: I like the refactoring here.

dxr/plugins/clang/tests/test_direct.py
```diff
@@ -47,7 +47,7 @@ def test_qualified_function_name_sensitive(self):
     def test_qualified_function_name_multiple_matches(self):
         """Multiple matches on fully qualified function name should return
```
pelmers: Does it still make sense to say "should return None" in these docstrings?

erikrose: Nope! Good catch

pelmers:
r+

PR #441, opened 2015-06-26
Fix bug 1177191, pass devnull as the stderr stream for get_contents.

pelmers:
I am not sure whether the [404's](https://dxr.allizom.org/rust/rev/6065bed37be55427ed0ac888b863a85696b28f9b/mk/llvm.mk) on the Rust tree are related to this.
ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1177191

### DIFF
--- a/dxr/vcs.py
+++ b/dxr/vcs.py
@@ -45,11 +45,11 @@ class Vcs(object):
         return type(self).__name__
 
     @classmethod
-    def invoke_vcs(cls, args, cwd):
+    def invoke_vcs(cls, args, **kwargs):
         """Return the result of invoking the VCS command on the repository,
-        with the current working directory set to the root directory.
+        with extra kwargs passed along to the Popen constructor.
         """
-        return subprocess.check_output([cls.command] + args, cwd=cwd)
+        return subprocess.check_output([cls.command] + args, **kwargs)
 
     def is_tracked(self, path):
         """Does the repository track this file?"""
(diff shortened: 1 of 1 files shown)


pelmers:
The Rust 404's aren't directly related, but I've a fix for that ready at https://github.com/pelmers/dxr/commit/77009a2455d72a7da83189637843e596fa2461ae once this merges (since it touches a couple of the same lines).

dxr/vcs.py
```diff
@@ -45,11 +45,11 @@ def get_vcs_name(self):
...
 
     @classmethod
-    def invoke_vcs(cls, args, cwd):
+    def invoke_vcs(cls, args, **kwargs):
         """Return the result of invoking the VCS command on the repository,
-        with the current working directory set to the root directory.
+        with extra kwargs passed along to the Popen constructor.
         """
-        return subprocess.check_output([cls.command] + args, cwd=cwd)
+        return subprocess.check_output([cls.command] + args, **kwargs)
```
erikrose: I'd prefer we explicitly have a `cwd` arg rather than passing all kwargs through. This makes it easier to learn, since you don't have to traverse multiple levels of documentation. Also, it lets us change this method later (to, say, use something other than check_output()) without first checking all call sites to make sure they're not passing some unanticipated kwargs in.

This is all assuming we want to pass relatively few kwargs through to check_output() for the forseeable future.

erikrose: Also, is there a reason to not make cwd required? It seems like asking for trouble to not specify it. We've already had that amply demonstrated, right?

pelmers: Good point, cwd shouldn't be optional.

erikrose:
r+

PR #442, opened 2015-06-29
Fix bug 1178103, 404 on dir/ listings.

pelmers:
Quick fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1178103

### DIFF
--- a/dxr/app.py
+++ b/dxr/app.py
@@ -205,7 +205,8 @@ def browse(tree, path=''):
     """
     config = current_app.dxr_config
     try:
-        return _browse_folder(tree, path, config)
+        # Strip any trailing slash because we do not store it in ES.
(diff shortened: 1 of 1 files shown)


erikrose:
r+

erikrose:
Actually, an even better solution would probably be to redirect to the slashless URL. That way, we'd be nicer to any caching we someday put in front.

erikrose:
GitHub doesn't do the redirect. I'm surprised.

erikrose:
What do you think?

pelmers:
I think either way would work; the only things affected would be relative URL's, but we do not use them.

erikrose:
At any rate, r+. It's better than it was, and we can always go the other way later.

pelmers:
merged, not sure why github didn't pick it up automatically.

PR #443, opened 2015-06-30
Add path promotion and path highlighting.

pelmers:
Inspired by https://wiki.mozilla.org/DXR_Result_Mixing and refined by kitchen polls, we promote probable paths (e.g. exact path fragment matches) at the top of ‘single-term’ queries (i.e. queries without a filter) and offer a link to see all path results.
Also include a path highlighter in `PathFilter` and some tests.
screen shot
^note the `sub` exact filename match listed first


Gruntfile.js
```diff
@@ -1,3 +1,15 @@
+var templateFiles = [
```
erikrose: +1. Now if we could just get rid of grunt altogether....

dxr/query.py
```diff
@@ -108,50 +136,30 @@ def results(self, offset=0, limit=100):
...
-                'results': self._line_query_results(filters, results, path_highlighters)
-                           if is_line_query
-                           else self._file_query_results(results, path_highlighters)}
+    def promoted_paths(self, term, promote_limit=5):
+        """Return a tuple (total number of path results, promoted paths),
+        where the number of promoted paths is no more than promote_limit."""
 
-        # Test: If var-ref (or any structural query) returns 2 refs on one line, they should both get highlit.
+        # We cannot import PathFilter at top of file because of circular dependency.
+        from dxr.plugins.core import PathFilter
```
erikrose: I think we can untangle this circular dependency by moving some_filters() and perhaps filters_by_name() to dxr.filters.

dxr/query.py
```diff
@@ -29,6 +30,34 @@ def direct_searchers(plugins):
     return [searcher for searcher, _ in sortables]
 
 
+def _construct_query(filters):
```
erikrose: Can we name this more informatively? Right now, "query" could be an ES query or a Query object (and Query objects are, after all, constructed). `_make_es_query`?

dxr/query.py
```diff
@@ -108,50 +136,30 @@ def results(self, offset=0, limit=100):
...
-        # Test: If var-ref (or any structural query) returns 2 refs on one line, they should both get highlit.
+        # We cannot import PathFilter at top of file because of circular dependency.
+        from dxr.plugins.core import PathFilter
+
+        filters = [[PathFilter(term, self.enabled_plugins)]]
+        path_query = {
+            'constant_score': {'query': _construct_query(filters), 'boost': 0.5}}
+        # We add a second query that boosts exact path part matches.
+        exact_query = {
+            'constant_score': {'query': {'match': {'path.parts': term['arg']}}, 'boost': 2.0}}
```
erikrose: Since you're using a constant_score query here and thus not taking advantage of scoring, is it possible to use an ES filter (rather than an ES query)? Filters are an order of magnitude faster and cacheable.

dxr/query.py
```diff
@@ -189,6 +197,39 @@ def direct_result(self):
...
+    def _perform_search(self, filters, is_line_query, offset, limit, query=None, sort=True):
+        """Perform a query using self.es_search, returning a tuple (total count of results,
+        list of results)
+
+        :arg filters: nested list of filters
+        :arg is_line_query: whether to query lines (otherwise query files)
+        :arg offset: return results starting from this number
+        :arg limit: maximum number of results to return
+        :arg query: valid ES query; if omitted then construct one from filters
+        :arg sort: whether to sort the results alphabetically by path and numerically by line
```
erikrose: It always sorts by something, so maybe something like `sort_by_path` or `score` or `alphabetize` would be better?

dxr/query.py
```diff
@@ -189,6 +197,39 @@ def direct_result(self):
...
+
+        """
+        if not query:
```
erikrose: I think a default behavior that is used once might as well be moved up to the callsite that uses it, uncomplicating this function.

erikrose:
I'm picking a lot of nits, but I want to say that I think your scoring strategy that forms the meat of this PR is really good: go for the complete path segments first and then fall back to containment.

dxr/query.py
```diff
@@ -189,6 +197,39 @@ def direct_result(self):
                 elif len(results) > 1:
                     return None
 
+    def _perform_search(self, filters, is_line_query, offset, limit, query=None, sort=True):
```
erikrose: Is there a more descriptive name we can give this? To me, it sounds like it has the same semantics as another attr, `es_search`. Maybe mention the highlighting or the things it returns?

dxr/query.py
```diff
@@ -372,6 +413,25 @@ def filter_menu_items(plugins):
...
+    extents.
+
+    :arg path: the path to highlight as a single string
+    :arg extents: iterable of unsorted, possibly overlapping (start, end) offset tuples
+
+    """
+    highlighted_content = highlight(path, extents)
+    # Split on slash, except the slash in </b>.
+    highlighted_fragments = []
+    for possibly_fragment in highlighted_content.split(sep):
```
erikrose: We actually shouldn't use os.path.sep here; it'll always be "/", no matter what. (In practice, it doesn't matter, since we support only UNIX, but stil…)

erikrose: I see I've perpetrated a similar wrong in build.py, whence the path arg of FileToIndex comes: using relpath() to just split up an OS-native path. So you get to do it better here. :-)

dxr/app.py
```diff
@@ -114,27 +122,34 @@ def _search_json(query, tree, query_text, is_case_sensitive, offset, limit, conf
...
-        results = [{'icon': icon,
-                    'path': path,
-                    'lines': [{'line_number': nb, 'line': l} for nb, l in lines],
-                    'is_binary': is_binary}
-                   for icon, path, lines, is_binary in count_and_results['results']]
+        count, results = query.results(offset, limit)
+        term = query.single_term()
+        promoted_count, promoted = 0, []
+        # Only offer promoted results when at the top of the page and we have only a single term.
+        if offset == 0 and term:
```
erikrose: Actually playing with the UI, I was surprised to find that, when I brought up a context menu on a path result (cool—that works!) and chose "Limit search to folder", I stopped getting path results. I think the condition we really want here isn't "there's a single term" but rather "there's a single text term, and all the rest of the filters are in the FILE domain". That way, the context menu would work, and we could get path results when searching for things like `ext:c main`.

dxr/static/templates/path_line.html
```diff
@@ -4,7 +4,7 @@
 {%- if not is_binary or is_dir -%}
 <a href="{{ url }}" {%- if is_dir %} data-path="{{ data_path }}" {% endif %}>
 {%- endif -%}
-    {{ display_path }}
+    {{ display_path|safe }}
```
erikrose: This is called from JS, right? Too bad. (Rather than putting `|safe` in the template, it's safer to return a Markup("some string") (rather than "some string") from Python.)

dxr/testing.py
```diff
@@ -50,7 +50,7 @@ def source_page(self, path):
 
     def found_files(self, query, is_case_sensitive=True):
         """Return the set of paths of files found by a search query."""
-        return set(result['path'] for result in
+        return set(sep.join(result['path']).replace('<b>', '').replace('</b>', '') for result in
```
erikrose: sep

dxr/query.py
erikrose: Never mind. I'm backpedaling on this after searching through the codebase and finding a bunch of other `sep` uses. Apparently, I decided to lean harder on the "runs only on UNIX" thing than I remembered.

dxr/query.py
```diff
@@ -372,6 +413,25 @@ def filter_menu_items(plugins):
             if filter.description)
 
 
+def highlight_path(path, extents):
```
erikrose: I'm not a fan of the backtracking this does, but I can't think of an alternative that doesn't greatly increase the complexity of highlight().

So I'll just complain about the docstring, which confused me until I really studied the procedure. Something like "Return a list of path segments, with <b> tags around highlit extents." (I at first thought you mean "Return a list of path fragments that contain <b> tags"—and do something else to those that don't.)

I think this will be fine if a <b> tag opens in one segment and closes in another, but that's only because the breadcrumbs happen to be made out of all inline elements atm. It's worth a comment in the template to make sure we don't change that at some point.

dxr/static/css/dxr.css
```diff
@@ -198,6 +195,31 @@ mark {
     background-color: transparent;
 }
 
+/* promoted stuff */
```
erikrose: Have you played with moving the path results into a right-hand sidebar? There was that violent reaction against using vertical real estate for them at the Whistler session.

dxr/static/css/icons.css
```diff
@@ -85,6 +85,9 @@ a.type {
 }
 /* End of Panel icons that are always shown */
 
+.icon-container.folder, .promoted-icon.folder {
```
erikrose: I can't get icons to show up for promoted results, no matter how much I recompile the templates.

dxr/query.py
pelmers: good call!

dxr/query.py
pelmers: Looking at the ES docs on querying and filtering, I'm not sure I can find a filter equivalent to the [match query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html). Do filters have a restriction against analyzing the search term? Maybe I wasn't looking in the right place.

dxr/app.py
pelmers: We can find out by doing the first few lines of results(), which would have to happen anyway. http://git.io/vm3Bp
Then we can iterate over and count the number of FILE and LINE filter.domain's.
We could move that outside of the call to results and pass it in instead.

pelmers: (so no increase in inefficiency, we'd just be moving things around)

dxr/static/templates/path_line.html
pelmers: Yeah, this is added in [JS](https://github.com/mozilla/dxr/blob/1666883ed0d28a7d34280c00dd477beac365f6b0/dxr/static/js/dxr.js#L119-L126). Someone could conceivably put html tags in their folder names and we would render them as markup.

dxr/query.py
erikrose: Filters don't in general have anything against query analysis, but the `term` filter does. Looking at NameFilterBase, it appears that the best I was able to come up with was to do a `match` query when case insensitive and a `term` filter when case sensitive. How does the case sensitivity switch work here? I notice that it does the right thing. The query emitted by _construct_query() would do it in the usual way, but I don't see how exact_query works.

erikrose: Ah, I think I see. path_query does the right thing, but exact_query will match case-sensitive all the time. Indeed, we don't get the whole-segment boost when searching case-insensitive:

1. Add 2 files: `ac.c` and `c.c`.
2. Uncheck Case-sensitive.
3. Search for c.C.

ac.c comes up first, but c.c should.

dxr/app.py
erikrose: Sounds good.

dxr/static/templates/path_line.html
erikrose: Hmm, we can't have that. We shouldn't have to trust committers to every tree to prevent XSS attacks. So I guess we need to either find a way to get rid of |safe here or explicitly escape path segments.

dxr/plugins/core.py
```diff
@@ -36,6 +36,10 @@
         'trigrams': {
             'type': 'string',
             'analyzer': 'trigramalyzer'
+        },
+        'parts': {
```
erikrose: Let's call this "segments" to jibe with what we call them everywhere else.

dxr/query.py
pelmers: Yes, I kind of assumed that the exact part of the query should always be case sensitive, since it's exact. Would this change entail adding a segments.lower field and matching on that when case is insensitive?

erikrose: That's the obvious way to me.

pelmers:
Updated, the [commit message](https://github.com/pelmers/dxr/commit/45821b4009f71145e1f38cb224b56fc7c027e7bf) summarizes changes.
todo next:
 - Figure out how to deal with the path line escaping issue.
 - Add a segments.lower field to allow case-insensitive exact matching.

dxr/static/templates/path_line.html
```diff
@@ -1,10 +1,13 @@
...
 {%- endif -%}
 {%- if not is_binary or is_dir -%}
-<a href="{{ url }}" {%- if is_dir %} data-path="{{ data_path }}" {% endif %}>
+    <a href="{{ url }}" {%- if is_dir %} data-path="{{ data_path }}" {% endif %}>
 {%- endif -%}
-    {{ display_path }}
+{# Remark: it is possible for <b> tags to be unbalanced within fragments if it opens in one
+   fragment and closes in another, but we avoid any issue by showing the entire path as a
+   single element. #}
+{{ display_path|safe }}
```
pelmers: Since we need to be sure we trust the input when we use `|safe`, I have traced execution path from request to template here: https://gist.github.com/pelmers/7d53a0fa02364f84542c.
In particular I think the portion to focus on is lines [15-34](https://gist.github.com/pelmers/7d53a0fa02364f84542c#file-xsstrace-py-L15-L34), where at the bottom I explain my reasoning for why I believe that the approach is safe.

pelmers: Also we trust our escaping in [result_lines.html](http://git.io/vmZnB), which follows almost the exact same execution flow as paths (only without the splitting on slash) since they share the query.highlight method.

dxr/plugins/core.py
```diff
@@ -36,6 +35,14 @@
         'trigrams': {
             'type': 'string',
             'analyzer': 'trigramalyzer'
+        },
+        'segments': {
```
erikrose: I don't think we ever use segments or segments_lower in a LINE query (confirm?), so we can save some index space (and maybe indexing CPU) by not including them on LINE docs.

pelmers: Confirmed, good point

pelmers:
Bumped with a switch to a flexbox-based layout for the search results page. The breadcrumbs and tree-selector now share a flexbox row. The regular and promoted results also share a flexbox row. A magic is that you can add flex-direction:column-reverse to the result container, and you'll get paths on the top rather than on the right.

dxr/static/templates/partial/promoted.html
```diff
@@ -0,0 +1,37 @@
...
+        {% endfor %}
+      </div>
+    </div>
+  {% endfor %}
+{%- endmacro %}
+
+{%- macro promoted_header(formatted_query, promoted_count, promoted_count_formatted, top_of_tree, tree) -%}
+    <p class="promoted-see-more">
+        <a href="{{ www_root }}/{{ tree }}/search?q=path:{{ formatted_query }}">{{ promoted_count_formatted }}
+            path {{ 'result' if promoted_count == 1 else 'results' }}</a> from the <a
```
erikrose: I don't think we need to include "from the foo tree" here after moving promoted results to the right. The main results that need nicely and are the first thing the eye sees.

dxr/static/css/icons.css
erikrose: I can now. *boggle*

dxr/static/css/dxr.css
```diff
@@ -198,7 +195,39 @@ mark {
     background-color: transparent;
 }
 
+/* promoted stuff, on the right */
+.promoted-container {
+    /* cannot grow, can shrink, prefers 30% size */
+    flex: 0 1 30%;
+}
```
erikrose: I did a little basic painting so we can get this landed. This doesn't look too bad:
```
border: 1px solic #a0a0a0;
border-radius: 4px;
padding: 1.5em 2em;
```
To make the perceived padding inside the box look symmetrical, we also need to apply this to .promoted-see-more:
```
margin: 0 0 1rem
```
(That also makes it consistent with .top-of-tree, the header for the ordinary results.)

And, finally, we need to change .promoted-row from `margin-bottom :5px` to…
```
margin-top: .4rem
```

dxr/static/css/dxr.css
```diff
@@ -228,7 +262,7 @@ mark {
     padding: 0 .5rem;
 }
 .results code {
-    white-space: pre;
+    white-space: pre-wrap;
```
erikrose: Really, really good call.

dxr/query.py
```diff
@@ -78,23 +108,31 @@ def _line_query_results(self, filters, results, path_highlighters):
...
                              chain.from_iterable(
                                  h(file) for h in path_highlighters)),
                    [],
                    file.get('is_binary', False))
 
-    def results(self, offset=0, limit=100):
-        """Return a count of search results and, as an iterable, the results
-        themselves::
+    def instantiate_filters(self):
+        """Instantiate applicable filters, yielding a list of lists, each inner list
```
erikrose: Here and in several other places, you wrap a docstring to a funny width. If we wrap them, it should be to 79 columns.

dxr/app.py
```diff
@@ -114,27 +122,36 @@ def _search_json(query, tree, query_text, is_case_sensitive, offset, limit, conf
...
-        count_and_results = query.results(offset, limit)
-        # Convert to dicts for ease of manipulation in JS:
-        results = [{'icon': icon,
-                    'path': path,
-                    'lines': [{'line_number': nb, 'line': l} for nb, l in lines],
-                    'is_binary': is_binary}
-                   for icon, path, lines, is_binary in count_and_results['results']]
+        filters = query.instantiate_filters()
+        # Pull up all the non-negated text terms.
+        text_terms = [term for term in query.terms if term['name'] == 'text' and not term['not']]
```
erikrose: I should really have named it `_terms`, as I intended it to be private to `Query`. Can we pull the manipulation of `terms` into `Query`? In fact, isn't this just a duplication of the `Query.single_term()` logic `direct_results()` uses? Can `promoted_paths()` just call `single_term()` and, like `direct_result()`, return None if there isn't one?

Also, instead of requiring the caller to call instantiate_filters(), hang onto the result, and pass it into results() and promoted_paths()—thus making the filters a public data structure—why don't we just move filter instantiation to the Query constructor? We can keep the filters around in self.filters or wherever and reference them from the methods that need them.

dxr/static/js/dxr.js
```diff
@@ -322,22 +323,39 @@ $(function() {
...
+                    var resultHead = buildResultHead(results[i].path, data.tree, icon, results[i].is_binary);
+                    // After we build the result head, join up the path so we can set it as an href.
+                    results[i].path = results[i].path.join('/');
+                    results[i].iconClass = resultHead[0];
+                    results[i].pathLine = resultHead[1];
+                }
+            });
 
             if (!append) {
+                // If we have promoted results then pass in a formatted query which bubbles the
```
erikrose: I would far rather do this server-side than try to reimplement query parsing in a half-way fashion here. For example, this will do the wrong thing if a query term has a colon or a space in a quoted string. Let's just construct the "more promoted results" query on the server and pass it down as part of the promoted results data structure.

dxr/app.py
pelmers: We don't use single_term for promoted paths because the condition is not quite the same: we want promoted paths even when the query is `text path:folder1 ext:abc`, but single_term is for when the query is just `text`.

dxr/query.py
```diff
@@ -48,8 +79,9 @@ def single_term(self):
         non-textual one, return None.
```
erikrose: You've changed the behavior of the funciton without updating its docstring.

dxr/query.py
```diff
@@ -48,8 +79,9 @@ def single_term(self):
...
 
         """
-        if len(self.terms) == 1:
-            term = self.terms[0]
+        text_terms = [term for term in self._terms if term['name'] == 'text' and not term['not']]
+        if len(text_terms) == 1:
+            term = text_terms[0]
             if term['name'] == 'text' and not term['not']:
```
erikrose: Now we're checking this condition redundantly.

dxr/app.py
```diff
@@ -114,27 +122,38 @@ def _search_json(query, tree, query_text, is_case_sensitive, offset, limit, conf
...
-                    'is_binary': is_binary}
-                   for icon, path, lines, is_binary in count_and_results['results']]
+        # Pull up all the non-negated text terms.
+        single_term = query.single_term()
+        count, results = query.results(offset, limit)
+        promoted_count, promoted, promoted_query = 0, [], ""
+        if offset == 0 and single_term:
+            # Now we pull out the single line term and pass into promoted paths.
+            try:
+                promoted_count, promoted, promoted_querystring = query.promoted_paths(single_term)
```
erikrose: This reads much better now. Maybe we should rename single_term() to single_text_term() to reflect its behavior change.

dxr/filters.py
```diff
@@ -66,6 +67,13 @@ def __init__(self, term, enabled_plugins):
         self._term = term
         self._enabled_plugins = enabled_plugins
 
+    def __str__(self):
+        """Return a string representation of this Filter as name:term, as it
+        appeared in the query string.
+        """
+
+        return "%s:%s" % (self.name, self._term['arg'])
```
erikrose: Like the JS, this is still naive. What if the term has a space in it? You basically have to do the reverse of the query parser—at least as far as reconstructing a single term.

dxr/query.py
```diff
@@ -78,23 +110,32 @@ def _line_query_results(self, filters, results, path_highlighters):
...
-            yield (icon(file['path'][0]),
-                   highlight(file['path'][0],
+            yield ('folder' if file['is_folder'] else icon(file['path'][0]),
+                   highlight_path(file['path'][0],
                              chain.from_iterable(
                                  h(file) for h in path_highlighters)),
                    [],
                    file.get('is_binary', False))
 
+    def _instantiate_filters(self, terms):
```
erikrose: The methods on Query (and most of the rest of the project) are tending toward nouns (single_term(), direct_result(), results()). Let's either stick with that for consistency (and inline this into `__init__` to avoid a collision on "_filters") or name it _make_filters() (consistent with _make_es_query()).

dxr/static/js/dxr.js
```diff
@@ -97,24 +97,25 @@ $(function() {
...
+    function buildResultHead(paths, tree, icon, isBinary) {
         var pathLines = '',
             pathRoot = '/' + tree + '/source/',
-            paths = fullPath.split('/'),
             splitPathLength = paths.length,
             dataPath = [],
             iconClass = icon.substring(icon.indexOf('/') + 1);
 
-        for (var pathIndex in paths) {
+        for (var pathIndex = 0; pathIndex < paths.length; pathIndex++) {
```
erikrose: Good fix, since for…in isn't ordered over Arrays.

dxr/static/js/dxr.js
```diff
@@ -97,24 +97,25 @@ $(function() {
...
             var index = parseInt(pathIndex),
```
erikrose: This has no point [anymore], right?

dxr/filters.py
```diff
@@ -66,6 +67,18 @@ def __init__(self, term, enabled_plugins):
         self._term = term
         self._enabled_plugins = enabled_plugins
 
+    def __str__(self):
+        """Return a string representation of this Filter as it appeared in the
+        query string, defaulting to name:term. Override this method for Filters
+        whose string representation does not match this format.
+        """
+
+        # If the term has spaces, then surround with double quotes and escape internal quotes.
```
pelmers: Here I reverse part of the logic in QueryVisitor.

dxr/static/templates/partial/promoted.html
```diff
@@ -0,0 +1,36 @@
...
+        {% endfor %}
+      </div>
+    </div>
+  {% endfor %}
+{%- endmacro %}
+
+{%- macro promoted_header(promoted_query, promoted_count, promoted_count_formatted, top_of_tree, tree) -%}
+    <p class="promoted-see-more">
+        <a href="{{ promoted_query }}">{{ promoted_count_formatted }}
+            path {{ 'result' if promoted_count == 1 else 'results' }}</a>
```
erikrose: We lost the colon.

erikrose:
This looks pretty good now. The only remaining things are UI. For example, I wish the promotion box were only as large as its contents. I believe that's what I had going on last night with my recommended CSS changes, so I'm not sure what's different.

screen shot 2015-07-28 at 17 30 44

And it would be nice (though it isn't blocking) for binary files to show up greyed out in the promoted results, just as they do in directory listings.

erikrose:
Great, now all that remains is to fix the text wrap somehow and make the promotion box less tall.

erikrose:
Eventually I'd like to revise the markup so we wrap on a per-line basis, not a per-file one. (And we can shrink the markup by emitting fewer invisible accessibility table headers, since we'll be using proper LIs.) But that doesn't need to block this. Merge it in! (And once Jenkins passes, ask fubar to kick off builds of everything to support the new format version.)

PR #444, opened 2015-06-30
Fix bug 1178882, extra line breaks on copy.

pelmers:
Avoid inserting extra whitespace in the page by asking Jinja to strip
the template block.

Thanks to @potch for help in fixing this!

Funnily enough, Chrome was working correctly before this fix, and now copying there is broken (no line breaks are kept on copy).

### DIFF
--- a/dxr/static/templates/text_file.html
+++ b/dxr/static/templates/text_file.html
@@ -57,7 +57,7 @@
 <pre>
 {% for line, annotations in lines -%}
 <code id="line-{{ loop.index }}" aria-labelledby="{{ loop.index }}">{{ line }}</code>
-{% endfor -%}
+{%- endfor -%}
 </pre>
         </td>
       </tr>


erikrose:
Can we figure out how to break neither of them?

pelmers:
possibly helpful details:
Looks like there's an issue filed against Chrome for this: https://code.google.com/p/chromium/issues/detail?id=317365
and up until Firefox 38 we had the same issue there: https://bugzilla.mozilla.org/show_bug.cgi?id=116083

erikrose:
Ah, this probably happened when we started storing linebreaks in ES (and then outputting them to the page, inside each `<code>` tag.

erikrose:
In my testing, Chrome (45), FF (38), and Safari all work fine with this patch. r+. What Chrome are you using?

pelmers:
I tried with Chrome 43 and it was broken; then I tested 45 and it worked fine.

I don't think the cause was storing linebreaks -- the behavior is broken currently on the production version too. [example](https://dxr.mozilla.org/mozilla-central/source/ipc/glue/BackgroundParent.h)

erikrose:
The reason I suspect the linebreak storage is that master currently doesn't put the linebreak inside the `<code>` tag; es did.

PR #445, opened 2015-06-30
Let searches for qualified but argless functions work

erikrose:
Now we can search for Scoped::function rather than needing to say Scoped::function(int, char). Fix https://bugzilla.mozilla.org/show_bug.cgi?id=1171242

### DIFF
--- a/dxr/filters.py
+++ b/dxr/filters.py
@@ -3,6 +3,8 @@
 from functools import wraps
 from funcy import identity
 
+from dxr.utils import is_in
+
 
 # Domain constants:
 FILE = 'file'
@@ -233,4 +235,4 @@ class QualifiedNameFilterBase(NameFilterBase):
         """
         return ((not self._term['qualified'] and
                  super(QualifiedNameFilterBase, self)._should_be_highlit(entity))
-                or entity['qualname'] == self._term['arg'])
+                or is_in(self._term['arg'], entity['qualname']))
--- a/dxr/plugins/clang/needles.py
+++ b/dxr/plugins/clang/needles.py
@@ -74,7 +74,8 @@ def ref_needles(condensed,
                 subkind=None,
                 include_qualname=True,
                 include_typeless_qualname=False):
-    """Return needles for references to a certain kind of thing.
+    """Return needles for references to a certain kind of language construct,
+    e.g. functions, variables, or namespaces.
 
     References are assumed to have names and qualnames.
 
@@ -84,7 +85,8 @@ def ref_needles(condensed,
     """
     subkind = subkind or name
     return needles(condensed,
-                   name, suffix='_ref',
+                   name,
+                   suffix='_ref',
                    kind='ref',
                    subkind=subkind,
                    include_qualname=include_qualname,
--- a/dxr/plugins/clang/needles.py
+++ b/dxr/plugins/clang/needles.py
@@ -8,18 +8,13 @@ from dxr.indexers import (iterable_per_line, with_start_and_end,
                           split_into_lines, Extent, Position)
 
 
+# TODO: Use.
 def sig_needles(condensed):
     """Return needles ((c-sig, type), span)."""
     return ((('c-sig', str(o['type'])), o['span']) for o in
             condensed['function'])
 
 
-# def needles(condensed, inherit, graph):
-#     """Return all C plugin needles."""
-#
-#     return chain(
-#         sig_needles(condensed),
-#     )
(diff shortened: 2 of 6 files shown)


dxr/plugins/clang/needles.py
```diff
@@ -1,28 +1,23 @@
...
-
-from funcy import imap, is_mapping, repeat, icat
+from itertools import chain
 
 from dxr.indexers import (iterable_per_line, with_start_and_end,
-                          split_into_lines, Extent, Position)
+                          split_into_lines)
 
 
+# TODO: Use.
```
pelmers: It looks like this is only used from the test cases. Should we move the function there?

dxr/plugins/clang/needles.py
```diff
@@ -35,24 +30,50 @@ def needles(condensed, name, suffix='', kind=None, subkind=None, keys=('name', '
...
+        if include_qualname:
+            qualname = value['qualname'] = entity['qualname']
+            if include_typeless_qualname:
+                try:
+                    # This is good enough for functions but probably not for
+                    # templates yet:
+                    offset = qualname.index('(')
+                except ValueError:
+                    pass
+                else:
```
pelmers: Is there a benefit to using the else clause here rather than putting its line into the try section?
At first I thought this was an indentation bug since I was looking at the two if's the else is nested within.

dxr/plugins/clang/needles.py
```diff
@@ -35,24 +30,50 @@ def needles(condensed, name, suffix='', kind=None, subkind=None, keys=('name', '
...
-    return needles(condensed, name, kind=kind)
-
-
-def ref_needles(condensed, name, subkind=None, keys=('name', 'qualname')):
+def ref_needles(condensed,
+                name,
+                subkind=None,
+                include_qualname=True,
+                include_typeless_qualname=False):
     """Return needles for references to a certain kind of thing.
```
pelmers: I realize you didn't change the line in this request, but I'm reading "certain kind of thing" and finding myself asking "what kind of what thing?" Can we clarify this?

dxr/plugins/clang/needles.py
```diff
@@ -1,28 +1,23 @@
...
-#     return chain(
-#         sig_needles(condensed),
-#     )
-
-
-def needles(condensed, name, suffix='', kind=None, subkind=None, keys=('name', 'qualname')):
+def needles(condensed,
+            name,
+            suffix='',
+            kind=None,
```
pelmers: Looks like the function doesn't use the `kind` argument.

pelmers:
I expect we'll need to bump format version here.

dxr/plugins/clang/needles.py
erikrose: It's not a test helper; it's a needle generator we never got around to building a filter for. We should do that at some point.

dxr/plugins/clang/needles.py
erikrose: In general, it's good to keep `try` sections as short as possible, so you don't catch (in this case) ValueErrors you didn't mean to. It also makes clear which lines you expect the ValueError to come from.

erikrose:
We actually can get away without a format bump because the new code still tolerates the old format. The new qualnames will be generated because the builds are currently on a timer. If the build triggers were smarter, we might have to bump the format, but, in this case, we don't.

dxr/plugins/clang/needles.py
erikrose: Look toward the bottom: `for entity in condensed[kind]`.

pelmers: Ah, oops!

dxr/filters.py
```diff
@@ -233,4 +235,4 @@ def _should_be_highlit(self, entity):
         """
         return ((not self._term['qualified'] and
                  super(QualifiedNameFilterBase, self)._should_be_highlit(entity))
-                or entity['qualname'] == self._term['arg'])
+                or is_in(self._term['arg'], entity['qualname']))
```
pelmers: I like that we're able to implement this feature without adding another field to the doc.

pelmers:
lgtm, r+.

dxr/filters.py
erikrose: I thought about doing a prefix match on the existing data as well, but converting direct searches to "any old prefix" logic starts a whole rabbit trail of design that I'd rather do another time. :-)

PR #446, opened 2015-07-01
Fix bug 770998, ability to search all javascript files.

pelmers:
Implement a field `union_only` on the Filter class, which tells
query construction that the filter should always be `or` with
other filters of the same name.

We can consider adding the union_only field to the NameFilter (and IdFilter/RefFilter) as well, which would let us search for functions of name `append` or `extend`, for example, where currently the intersection of these filters would always be empty since they go by exact matches (edit: unless they occur on the same line).

remark: This work would be obsoleted if we in the future implement a general boolean query language, but I feel that the functionality here is not worth delaying.


dxr/filters.py
```diff
@@ -41,12 +41,15 @@ class Filter(object):
         one encountered will be used. An empty description will hide a filter
         from the menu. This should probably be used only internally, by the
         TextFilter.
+    :ivar union_only: Whether this filter will always be ORed with others of the same name,
+        useful for filters where the intersection would always be empty, such as extensions.
```
erikrose: No period, since this is not a sentence but merely a noun phrase :-)

dxr/query.py
```diff
@@ -102,8 +102,27 @@ def results(self, offset=0, limit=100):
...
+            for term in self.terms:
+                for f in enabled_filters_by_name[term['name']]:
+                    if predicate(f):
+                        d.setdefault(term['name'], []).append(f(term, self.enabled_plugins))
+            return d.values()
+
+        # Some filters, such as ExtFilter, do not make sense to be AND'ed together, so we move
+        # them all to their own lists at the end of the regular filters list, such that they
+        # will be joined by OR instead.
+        filters = (group_filters_by_term(lambda f: not f.union_only) +
```
erikrose: Probably not an issue in practical terms, but we'd likely see a little efficiency boost if we could avoid +ing lists together. We could make the two group_filters_by_* functions into generators and list(the first) and the_list.extend(the second).

dxr/plugins/core.py
```diff
@@ -311,6 +311,8 @@ class ExtFilter(Filter):
     domain = FILE
     description = Markup('Filename extension: <code>ext:cpp</code>. Always '
                          'case-sensitive.')
+    # The intersection of two different Ext filters would always be nothing.
+    union_only = True
```
erikrose: I like this. It's really easy to understand and apply to new filters.

dxr/query.py
```diff
@@ -102,8 +102,27 @@ def results(self, offset=0, limit=100):
         # will OR the elements of the inner lists and then AND those OR balls
         # together.
         enabled_filters_by_name = filters_by_name(self.enabled_plugins)
-        filters = [[f(term, self.enabled_plugins) for f in enabled_filters_by_name[term['name']]]
-                   for term in self.terms]
+
+        def group_filters_by_term(predicate):
+            # Return a nested list of ES filters for each term, filtered on predicate(Filter).
```
erikrose: Any reason not to use a docstring?

dxr/query.py
```diff
@@ -102,8 +102,27 @@ def results(self, offset=0, limit=100):
         # will OR the elements of the inner lists and then AND those OR balls
         # together.
```
erikrose: We should move this comment down to merge with your "Some filters" comment; it hasn't been in the right place for quite some time.

tests/test_extensions/test_extensions.py
```diff
@@ -9,6 +9,8 @@ def test_extensions(self):
         self.found_files_eq('ext:c', ['main.c', 'dot_c.c'])
         self.found_files_eq('ext:cpp', ['hello-world.cpp'])
         self.found_files_eq('ext:inc', ['hello-world.inc'])
+        # Make sure that multiple ext filters are joined with 'or' instead of 'and.'
+        self.found_files_eq('ext:c ext:cpp', ['main.c', 'dot_c.c', 'hello-world.cpp'])
```
erikrose: I'd like to see this broken off into a separate test, just to accentuate that it's testing the underlying union_only machinery, not (so much) the ExtFilter itself.

erikrose:
Otherwise, it looks good to me. The lambda predicates I wasn't so sure about since they're always the same, but I came to believe they are a clear factoring, since they move the "hey, just the unions" bit to the caller, where it's relevant. (If it causes a perf problem, we can hard-code them and move the "hey, just the unions" bit into the function names.)

I'd like to have one more quick look once you address those few little things, and then we can push this up. This'll be a definite improvement in Just Doing The Right Thing.

dxr/query.py
pelmers: The Github note isn't marked as outdated, but I've moved this comment.

erikrose: I didn't mean to pin it on you; I just noticed that it's in the wrong place while reading the code.

dxr/query.py
```diff
@@ -97,13 +97,35 @@ def results(self, offset=0, limit=100):
                          ...]}
 
         """
+        enabled_filters_by_name = filters_by_name(self.enabled_plugins)
+
+        def group_filters_by_term(predicate):
+            """Return an iterable of a list of ES filters for each term, filtered on
```
erikrose: "an iterable of lists of" makes more sense. Or sometimes I say "Yield lists of…" for conciseness.

tests/test_extensions/test_extensions.py
erikrose: Whoa, I didn't mean you had to make a whole new instance for it; I just meant a separate test, as in a separate method on the same class. Sorry for the misunderstanding.

pelmers:
I probably should have asked to clarify on where to put the test! It felt a little funny making an entirely new instance for it.

erikrose:
Yep, if you feel funny, you should definitely say something. :-) r+

PR #447, opened 2015-07-01
Fix bug 1177761, non-root directory git permalinks.

pelmers:
git’s show command requires a "./" prepended to the file name to make
it relative to cwd, rather than the git root.
Add a test to make sure this does not regress.


tests/test_vcs_git/test_vcs_git.py
```diff
@@ -13,27 +9,40 @@ class GitTests(DxrInstanceTestCaseMakeFirst):
     def test_diff(self):
         """Make sure the diff link exists and goes to the right place."""
         response = self.client().get('/code/source/main.c')
-        ok_('/commit/cb339834998124cb8165aa35ed4635c51b6ac5c2" title="Diff" class="diff icon">Diff</a>' in response.data)
+        ok_('/commit/5e2b2b554eb86f90e189217fa9dc2eba66259910" title="Diff" class="diff icon">Diff</a>' in response.data)
```
erikrose: It might be nice to factor this out as a name: `LATEST_REV` or something. Then we wouldn't have to touch every test (like you had to) whenever we change anything. Plus, the semantics would be clear.

tests/test_vcs_git/test_vcs_git.py
```diff
@@ -13,27 +9,40 @@ class GitTests(DxrInstanceTestCaseMakeFirst):
...
+        response = self.client().get('/code/rev/cb339834998124cb8165aa35ed4635c51b6ac5c2/main.c')
+        eq_(response.status_code, 200)
+
+    def test_deep_permalink(self):
+        """Make sure the permalink link exists and goes to the right place."""
+        response = self.client().get('/code/source/deeper/deeper_file')
+        ok_('/rev/5e2b2b554eb86f90e189217fa9dc2eba66259910/deeper/deeper_file" title="Permalink" class="permalink icon">Permalink</a>' in response.data)
+        response = self.client().get('/code/rev/5e2b2b554eb86f90e189217fa9dc2eba66259910/deeper/deeper_file')
+        eq_(response.status_code, 200)
+        ok_("This file tests" in response.data)
```
erikrose: It's good that you test this and not just the 200 code.

tests/test_vcs_git/test_vcs_git.py
```diff
@@ -13,27 +9,40 @@ class GitTests(DxrInstanceTestCaseMakeFirst):
...
-        ok_('/rev/cb339834998124cb8165aa35ed4635c51b6ac5c2/main.c" title="Permalink" class="permalink icon">Permalink</a>' in response.data)
+        ok_('/rev/5e2b2b554eb86f90e189217fa9dc2eba66259910/main.c" title="Permalink" class="permalink icon">Permalink</a>' in response.data)
+        # Test that it works for this revision and the last one.
+        response = self.client().get('/code/rev/5e2b2b554eb86f90e189217fa9dc2eba66259910/main.c')
+        eq_(response.status_code, 200)
+        response = self.client().get('/code/rev/cb339834998124cb8165aa35ed4635c51b6ac5c2/main.c')
+        eq_(response.status_code, 200)
+
+    def test_deep_permalink(self):
+        """Make sure the permalink link exists and goes to the right place."""
```
erikrose: Can you say a word about why this test is needed in addition to the test_permalink one?

tests/test_vcs_git/test_vcs_git.py
```diff
@@ -13,33 +11,50 @@ class GitTests(DxrInstanceTestCaseMakeFirst):
...
+        ok_('/rev/%s/main.c" title="Permalink" class="permalink icon">Permalink</a>' % LATEST_REVISION in response.data)
+        # Test that it works for this revision and the last one.
+        response = self.client().get('/code/rev/%s/main.c' % LATEST_REVISION)
+        eq_(response.status_code, 200)
+        response = self.client().get('/code/rev/%s/main.c' % PREVIOUS_REVISION)
+        eq_(response.status_code, 200)
+
+    def test_deep_permalink(self):
+        """Make sure the permalink link exists and goes to the right place for files not in the
+        top-level directory. This test makes sure that the permalink works even for files not in
```
erikrose: Looks good. I think we can remove "This test makes sure that the permalink works even for files not in the top level git root directory, since" without any loss of meaning. (And if you're going to skip a line before the function body, do it inside the quotes; that's the convention we mostly follow.)

erikrose:
With that, r+.

PR #448, opened 2015-07-01
Fix bug 1007293, blame link scrolls to selected line.

pelmers:
Add an extra field to the FileToSkim.links() return tuple to indicate
whether to update to the selected line. Allow the code highlighter to
handle #l{number} in addition to #{number} so that we can re-use the
same code for links to Github, hg web, and permalinks.


dxr/plugins/clang/indexers.py
```diff
@@ -178,7 +178,7 @@ def get_scopes_to_members():
                 # First, link to the type definition itself:
                 links = [(type['kind'],
                           type['name'],
-                          '#%s' % type['span'].start.row)]
+                          '#%s' % type['span'].start.row, False)]
```
erikrose: It's not your fault; it's mine for getting behind in reviews. But regions and refs now use actual classes for their return values, so let's take the opportunity to do so for links as well. Then, the next time we add something, we won't need to touch every line that deals with the data structure. :-)

dxr/static/js/code-highlighter.js
```diff
@@ -115,21 +115,25 @@ $(function () {
...
+    //update the link href based on the windowHash.
+    function updateLink(anchor, windowHash) {
+        var href = anchor.attr('href'),
+            hash_loc = href.indexOf('#');
+        // If the link already has #, then cut it off.
         if (hash_loc >= 0)
-            permalink_href = permalink_href.substring(0, hash_loc);
-        permalink.attr('href', permalink_href + windowHash);
+            href = href.substring(0, hash_loc);
+        // hg web uses #l and GitHub uses #L but supports #l, so we use #l here.
```
erikrose: That's cool. If, instead of just tacking the anchor on at the end, we did some sort of token substitution, we could support sending entire highlight *ranges* to GitHub. For example, we might use something like "{start_highlight}" and "{end_highlight}", which would let us express URLs to systems that support just a single line (like hgweb) or multiple (like GH) without a lot of special casing. For example, GH would be "/...#L{start_highlight}-L{end_highlight}".

erikrose: (We'd stick the template containing the {} tokens in a data attribute.)

dxr/static/js/code-highlighter.js
```diff
@@ -7,13 +7,13 @@
...
  * 3) Highlight lines when page loads, if window.location.hash exists
- * In addition, we update the permalink link to keep it synchronized with window.location.
+ * In addition, we update links to keep them synchronized with window.location.
  */
 
 $(function () {
     'use strict';
     var container = $('#line-numbers'),
-        permalink = $('.permalink'), // whenever we update window.location, update this href too
+        syncLinks = $('.sync-link'), // whenever we update window.location, update these too
```
erikrose: If we go down the token replacement road, we'll want to look up how efficient a[data-template] selectors are. Maybe we don't need a marker class.

dxr/static/templates/text_file.html
```diff
@@ -17,7 +17,7 @@
           <ul>
             {% for item in section['items'] %}
               <li>
-                <a href="{{ item.href }}" title="{{ item.title }}" class="{{ item.icon }} icon">{{ item.title }}</a>
+                <a href="{{ item.href }}" title="{{ item.title }}" class="{{ item.icon }} icon {% if item.sync %} sync-link {% endif %}">{{ item.title }}</a>
```
erikrose: There are some extra spaces here that will result in needless spaces in the final output.

dxr/static/js/code-highlighter.js
pelmers: I like the idea, but I worry that it's more complicated than we expect, since DXR also has multi-ranges highlighting, e.g. https://dxr.allizom.org/mozilla-central/source/moz.build#7-13,22,32-43

erikrose: I was thinking we'd stop short of supporting multiple ranges, since nothing else can make use of them atm. That's how I still lean.

But you got me thinking again about the problem of constructing multi-region ranges for outside systems, when we can't predict how they'd format them. After a brief detour through simple sets of tokens we could provide to do fake boolean logic, it occurred to me that we already have a safe, client-side language available for building strings that can do conditionals and everything: nunjucks. So, if we do any sort of text substitution at all, I think we should spell the tokens in a Jinja-compatible way: `{% smoo %}`. Then at least we'll be forward compatible if we do support multiple ranges someday. (Can nunjucks parse a template client-side? I think it can.)

pelmers:
On a double check, I don't think that hg web supports region highlight at all. It seems to only have jump to line. I will change the template correspondingly.
Example: http://hg.mozilla.org/mozilla-central/file/9340658848d1/xpcom/idl-parser/xpidl.py#l10-l30

PR #449, opened 2015-07-01
Restore same-symbol highlighting

erikrose:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1123877

While we're at it, clean up the API and implementation. regions() and refs() now return Region and Ref classes which will be extensible in the future as we add things (like we've added `hover` and `qualname_hash` to refs).

### DIFF
--- a/dxr/indexers.py
+++ b/dxr/indexers.py
@@ -225,7 +225,7 @@ class FileToSkim(PluginConfig):
 
         Yield an ordered list of extents and menu items::
 
-            (start, end, (menu, hover))
+            (start, end, (menu, hover, qualname))
 
         ``start`` and ``end`` are the bounds of a slice of a Unicode string
         holding the contents of the file. (``refs()`` will not be called for
--- a/dxr/indexers.py
+++ b/dxr/indexers.py
@@ -225,23 +225,11 @@ class FileToSkim(PluginConfig):
 
         Yield an ordered list of extents and menu items::
 
-            (start, end, (menu, hover, qualname))
+            (start, end, Ref)
 
         ``start`` and ``end`` are the bounds of a slice of a Unicode string
         holding the contents of the file. (``refs()`` will not be called for
         binary files.)
-
-        ``hover`` is the contents of the <a> tag's title attribute. (The first
-        one wins.)
-
-        ``menu`` is a list of mappings, each representing an item of the
-        context menu::
-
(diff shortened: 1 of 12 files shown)


dxr/indexers.py
```diff
@@ -406,6 +394,101 @@ def needles_by_line(self):
...
+
+    def __init__(self, css_class):
+        self.css_class = css_class
+
+    def es(self):
+        return self.css_class
+
+    @classmethod
+    def es_to_triple(cls, es_region):
+        """Convert ES-dwelling refs representation to a (start, end, Region)
```
pelmers: "refs repr" should be "region representation" here.

dxr/lines.py
```diff
@@ -331,7 +296,7 @@ def finished_tags(lines, refs, regions):
     # Plugins return unicode offsets, not byte ones.
 
     # Get start and endpoints of intervals:
-    tags = list(tag_boundaries(refs, regions))
+    tags = list(tag_boundaries(list(chain(refs, regions))))
```
pelmers: Is it necessary to call list on the chained sequence? (it looks like it's only iterated over once)

dxr/lines.py
```diff
@@ -427,17 +374,9 @@ def segments(text, tags, bof_offset):
...
-                menu, hover = payload.payload
-                menu = cgi.escape(json.dumps(menu), True)
-                if hover:
-                    title = ' title="' + cgi.escape(hover, True) + '"'
-                else:
-                    title = ''
-                yield u'<a data-menu="%s"%s>' % (menu, title)
+                yield payload.closer()
+            else:
+                yield payload.opener()
```
pelmers: This reads really nicely now!

dxr/lines.py
erikrose: Good catch! That was just for debugging.

dxr/lines.py
```diff
@@ -340,7 +305,7 @@ def finished_tags(lines, refs, regions):
     tags.sort(key=nesting_order)  # balanced_tags undoes this, but we tolerate
                                   # that in html_lines().
     remove_overlapping_refs(tags)
-    return balanced_tags(tags)
+    return list(balanced_tags(tags))
```
pelmers: Also making a list here, which doesn't seem necessary

pelmers:
r+

PR #450, opened 2015-07-02
Update most of our requirements.

erikrose:
If a newer Flask optimizes url_for() at all, I want it; it's a hot spot for us. Update most of the rest while we're at it.

Maybe futures will stop hanging on Vagrant. It looks like it at least improves exception handling; maybe we can get rid of our custom stuff.

Didn't upgrade funcy or elasticsearch because I don't trust their quality in general.


pelmers:
Looks good to me. r+

However I wouldn't expect much improvement from url_for; it looks like the code for it hasn't changed in a while. http://git.io/vqfRT

erikrose:
That's okay; I've been meaning to upgrade Flask anyway.

PR #451, opened 2015-07-03
Bug 1052572 - Check if Vcs can make links before trying

djc:


djc:
I do see the "Permalink" from Git twice; not sure if/how that's related. I looked into it a bit, but the code path for that didn't make a lot of sense to me.

pelmers:
In case it helps, I'll point you to where the links are generated.
Permalink is created in core.py: https://github.com/mozilla/dxr/blob/1fb8bc80e3c3be59c286bbc9f5c30491fe169d77/dxr/plugins/core.py#L456-L462
Other links are in omniglot.py: https://github.com/mozilla/dxr/blob/1fb8bc80e3c3be59c286bbc9f5c30491fe169d77/dxr/plugins/omniglot.py#L27-L30

djc:
I found those, but that doesn't tell me why the `links()` method from the `core` plugin ostensibly gets called twice.

erikrose:
1. If something is showing up twice, could you show us how to reproduce it?
2. We should be able to solve this without expanding the API: we can just define a links() method on Git that returns [] unless self.upstream is populated (and otherwise delegates to the superclass).

PR #453, opened 2015-07-07
Optimize indexing by deferring menu generation to request time.

erikrose:
Based on profiles, this should make clang indexing 25-50% faster.

Reviewing one commit at a time will probably be easiest for this.


pelmers:
I really like the sound of not having to store all the menu strings in ES. I want to see if I'm understanding this design correctly: instead of storing the entire menu in ES, we store a plugin and id field which identifies the responsible class, and a data field which the MenuMaker defines and later uses to reconstruct the full menu option.

Have you been able to benchmark how big the impact of the change is (on index and request times and the weight of the index)?

erikrose:
> responsible class, and a data field which the MenuMaker defines and later uses to reconstruct the full menu option.

Correct, except it constructs potentially several menu options. Many types of entity, like clang functions, spawn many menu items from the same data (callers, references, overrides, etc.). Why store the data more than once?

> Have you been able to benchmark how big the impact of the change is (on index and request times and the weight of the index)?

I haven't got all the pieces in place yet, but I can estimate based on the profile, which shows url_for taking about 1/3 of the total index time. If I don't accidentally do something expensive, this should shave that much off. We might also get a little speed boost because we're sending less to ES, if this suddenly becomes un-CPU-bound and that begins to matter.

I can't say yet how much this will affect the index size. I consider it a bonus, but I wouldn't be surprised if it's substantial: {"clang", "Function", "some::qual::name"} is a lot smaller than {"icon":"reference","title":"Find references to macros with this name","html":"Find references","href":"/code/search?q=%2Bmacro-ref%3AVERY_EXTERNAL"}, and that's in the fairly rare case that there's a single menu item generated.

dxr/plugins/python/menus.py
```diff
@@ -1,15 +1,23 @@
...
-        {'html': 'Find subclasses',
-         'title': 'Find subclasses of this class',
-         'href': search_url(tree, '+derived:' + qualname),
-         'icon': 'type'},
-        {'html': 'Find base classes',
-         'title': 'Find base classes of this class',
-         'href': search_url(tree, '+bases:' + qualname),
-         'icon': 'type'},
-    ]
+class PythonPluginAttr(object):
```
erikrose: Maybe make this private for consistency with the other plugins' such classes.

dxr/plugins/rust/menu.py
```diff
@@ -34,106 +35,187 @@ def truncate_value(value, typ=""):
...
+class JumpToTypeDeclarationMenuMaker(_JumpToTarget, _RustPluginAttr):
+    target_name = 'type declaration'
+
+
+class JumpToVariableDeclarationMenuMaker(_JumpToTarget, _RustPluginAttr):
+    target_name = 'variable declaration'
+
+
+class JumpToFunctionDeclarationMenuMaker(_JumpToTarget, _RustPluginAttr):
+    target_name = 'function declaration'
```
nrc: Why all these classes? Seems like it would be nicer to make target_name a regular field set in the ctor.

nrc:
lgtm

dxr/plugins/rust/menu.py
erikrose: The idea was to get the constant strings out of ES for size and so we could change them without reindexing. Was this the right tradeoff? I'm not sure. The size didn't go down as much as I hoped: 400MB for the 9GB moz-central index.

dxr/lines.py
```diff
@@ -35,6 +36,133 @@ def __repr__(self):
...
+
+    Carries with it a context menu and other metadata.
+
+    """
+    sort_order = 1
+    __slots__ = ['menu', 'hover', 'qualname_hash']
+
+    def __init__(self, menu, hover=None, qualname=None, qualname_hash=None):
+        """
+        :arg menu: a list of MenuMakers that generate the context menu
```
pelmers: In the spirit of Name it What You Call it, would it make more sense to name this arg to menu_makers?

dxr/lines.py
```diff
@@ -35,6 +36,133 @@ def __repr__(self):
...
+        :class:`~dxr.lines.Ref`) triple.
+
+        :arg es_ref: An item from the array under the 'refs' key of an ES LINE
+            document
+        :arg tree: The :class:`~dxr.config.TreeConfig` representing the tree
+            from which the ``es_ref`` was pulled
+
+        """
+        def rehydrate_menu_makers(dried):
+            """Given the contents of the 'menu' key from ES, return a list of
```
pelmers: nit: rather than returning a list, it looks like this is a generator.

dxr/plugins/rust/menu.py
```diff
@@ -1,6 +1,7 @@
 import os
 
-from dxr.indexers import Ref
+from dxr.lines import Ref
+from dxr.menus import MenuMaker, MultiDatumMenuMaker, SingleDatumMenuMaker
```
pelmers: MenuMaker is an unused import.

dxr/plugins/rust/menu.py
```diff
@@ -34,106 +35,187 @@ def truncate_value(value, typ=""):
...
-                     "Go to DXR index of this crate on " + get_domain(dxrurl))
-    add_link_to_menu(menu, srcurl,
-                     "Go to source" + extra_text,
-                     "Go to source code for this crate on " + get_domain(srcurl))
-    add_link_to_menu(menu, docurl,
-                     "Go to docs" + extra_text,
-                     "Go to documentation for this crate on " + get_domain(docurl))
 
+def std_lib_links(tree_config, menu, (docurl, srcurl, dxrurl), extra_text = ""):
+    menu.insert(0, StdLibMenuMaker(tree_config, (doc_url, src_url, dxr_url, extra_text)))
```
pelmers: docurl, doc_url
srcurl, src_url
dxrurl, dxr_url

dxr/plugins/rust/menu.py
```diff
@@ -34,106 +35,187 @@ def truncate_value(value, typ=""):
...
+
+        """
+        path = decl['file_name']
+        if path:
+            return cls(tree, path, decl['file_line'])
+        else:
+            warn("Can't add jump to empty path.")  # Can this happen?
+
+    def menu_items(self):
+        yield {'html': 'Jump to %s' % self.target_name,
```
pelmers: It looks like we don't define target_name anywhere.

dxr/plugins/rust/menu.py
```diff
@@ -34,106 +35,187 @@ def truncate_value(value, typ=""):
...
+    target_name = 'variable declaration'
+
+
+class JumpToFunctionDeclarationMenuMaker(_JumpToTarget, _RustPluginAttr):
+    target_name = 'function declaration'
+
+
+class TraitImplMenuMaker(SingleDatumMenuMaker):
+    def menu_items(self):
+        qualname, count = self.data
```
pelmers: We pull out qualname, count here, but use self.count and self.qualname in the yielded value.

dxr/plugins/rust/menu.py
```diff
@@ -34,106 +35,187 @@ def truncate_value(value, typ=""):
...
+def function_menu(tree, datum, tree_config):
+    makers = []
     if 'declid' in datum and datum['declid'] in tree.data.functions:
-        # it's an implementation, find the decl
+        # It's an implementation; find the decl:
         decl = tree.data.functions[datum['declid']]
-        add_jump_definition(tree, tree_config, menu, decl, "Jump to trait method")
+        makers.append(JumpToTraitMethodMenuMaker.from_decl(tree_config, decl))
     else:
         # it's a decl, find implementations
```
pelmers: For symmetry, may as well also make this "It's a declaration; find implementations:"

dxr/plugins/rust/menu.py
```diff
@@ -1,6 +1,7 @@
 import os
 
-from dxr.indexers import Ref
+from dxr.lines import Ref
+from dxr.menus import MenuMaker, MultiDatumMenuMaker, SingleDatumMenuMaker
 from dxr.utils import search_url
```
pelmers: A few things we're trying to use here without import: url_for, warn, BROWSE.

dxr/plugins/rust/menu.py
```diff
@@ -173,93 +255,100 @@ def variable_menu(tree, datum, tree_config):
...
-    return None
+
+
+class TypeMenuMaker(SingleDatumMenuMaker, _RustPluginAttr):
+    def menu_items(self):
+        kind, qualname = self.data
+        if kind == 'trait':
+            yield {'html': "Find sub-traits",
+                   'title': "Find sub-traits of this trait",
+                   'href': search_url(tree_config, "+derived:%s" % quote(qualname)),
```
pelmers: I think we want self.tree, rather than tree_config.

dxr/vcs.py
```diff
@@ -173,8 +175,7 @@ def _construct_upstream_url(self):
                 if repo.startswith("git@github.com:"):
                     return "https://github.com/" + repo[len("git@github.com:"):]
                 elif repo.startswith(("git://github.com/", "https://github.com/")):
-                    if repo.endswith(".git"):
-                        repo = repo[:-len(".git")]
+                    repo = without_ending('.git', repo)
```
pelmers: I like this refactor!

pelmers:
I notice that there's a number of runtime errors in the Rust plugin that the smoke test catches. I think we should remove the SkipTest flag from it. The makefiles call on $RUSTC, but since we have rustc in the path I think that we can say RUSTC ?= rustc in the makefiles.

pelmers:
I also want to follow up on the note described in https://github.com/erikrose/dxr/blob/2834a6ceca850ec4e1176bae83fcc26fa334e61c/dxr/lines.py#L60-L66.

I haven't fleshed out how it would look, but it seems that we could simplify the implementation of on-demand menu generation by moving menu_items into the Ref class and let plugins themselves decide if they need to factor out further if many menus are repeated between different data items (like the Rust plugin) or not (most of the other plugins define MenuMaker classes that are only used once).

dxr/plugins/rust/menu.py
erikrose: Yikes. Apparently the test coverage is worse than I thought.

erikrose:
Not only can we unskip test_smoke, but we don't have to mess with the makefiles, since the rust DXR plugin sets the $RUSTC env var to "rustc" if it's not already set. :-D

dxr/lines.py
```diff
@@ -35,6 +37,167 @@ def __repr__(self):
...
+            plugins = all_plugins()
+            try:
+                return plugins[plugin].refs[id]
+            except KeyError:
+                warn('Ref subclass from plugin %s with ID %s was referenced '
+                     'in the index but not found in the current '
+                     'implementation. Ignored.' % (plugin, id))
+
+        payload = es_data['payload']
+        cls = ref_class(payload['plugin'], payload['id'])
```
pelmers: I think we need a None check here, in case ref_class fails, we want to ignore it rather than error out.

dxr/lines.py
```diff
@@ -35,6 +37,167 @@ def __repr__(self):
...
+                icon: the icon to show next to the menu item: the name of a PNG
+                    from the ``icons`` folder, without the .png extension
+            }
+
+        Typically, this pulls data out of ``self.menu_data``.
+
+        """
+        raise NotImplementedError
+
+    def opener(self):
```
pelmers: I think a docstring would be nice here, perhaps to describe how we put menu items into the HTML so we can pull them out on clicks.

dxr/menus.py
```diff
@@ -0,0 +1,117 @@
+from dxr.utils import without_ending
```
pelmers: Hopefully we'll say bye to this file once the Rust refs are ported.

dxr/lines.py
erikrose: Now that we have fewer units in play that the index cares about (just one Ref subclass rather than a class for almost every menu item), I think we can just have this fail hard when it fails. When you mess with a Ref subclass in a non-backward-compatible way, bump the format version.

dxr/lines.py
erikrose: Done. Good idea.

erikrose:
Superseded by https://github.com/mozilla/dxr/pull/465

PR #454, opened 2015-07-07
hash() results in CPython are 64 bits on 64-bit archs.

erikrose:


pelmers:
r+

I'm curious what was happening previously; were 32 bits being chopped off the hash?

PR #455, opened 2015-07-15
Do not request results when viewing files, fixes bug 1183885.

pelmers:
Previously we were requesting the results of q=“” when viewing files
and scrolling around. Now only request more results on search pages.

example: https://dxr.allizom.org/mozilla-central/source/moz.build

### DIFF
--- a/dxr/static/js/dxr.js
+++ b/dxr/static/js/dxr.js
@@ -195,31 +195,8 @@ $(function() {
      * Starts or restarts the scroll position poller.
      */
     function pollScrollPosition() {
+        clearInterval(scrollPoll);
         scrollPoll = setInterval(infiniteScroll, 250);
-    }
-
-    // On document ready start the scroll pos poller.
-    pollScrollPosition();
-
-    /**
-     * Updates the window's history entry to not break the back button with
-     * infinite scroll.
-     * @param {int} offset - The offset to store in the URL
-     */
-    function setHistoryState(offset) {
-        var state = {},
-            re = /offset=\d+/,
-            locationSearch = '';
-
-        if (location.search.indexOf('offset') > -1) {
-            locationSearch = location.search.replace(re, 'offset=' + offset);
-        } else {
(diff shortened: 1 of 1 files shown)


erikrose:
It's hard to test this atm, because infinite scroll is broken for everything, always. Somehow, the "limit=" querystring param gets into the URL bar, and then nothing will ever load past that limit.

pelmers:
Can you point me to an example of the infinite scroll breaking? It works fine for me on search results of staging. I could see it breaking if we resize the window (make it bigger) since the limit is set based on the size of the window.

erikrose:
1. Create a file like this, with at least 400 lines:

```
1 thing
2 thing
3 thing
...
```

2. Load up http://33.33.33.77:8000/code/source/.
3. Type "thing" in the search box.
4. Wait 300ms.
5. Scroll the page. Not all the results will load. The URL becomes "http://33.33.33.77:8000/code/search?q=thing&redirect=false&case=false&limit=102&offset=0" (note the "limit=").

pelmers:
cannot reproduce
I tried to reproduce the bug with these steps, but I have all 900+ results loading (in case the gif is too blurry). Maybe because my limit=78 rather than 102?

erikrose:
Still can't get it to work. Tried Safari. Trying a new venv now. In any case, `limit` and `offset` are not supposed to get into the URL.

erikrose:
New venv does not avail. :-(

erikrose:
I am sad to say that I still get a spurious `q=` request while scrolling through my 391-line "numbers copy.txt" file:
```
33.33.33.1 - - [05/Aug/2015 16:40:15] "GET /code/source/numbers%20copy.txt HTTP/1.1" 200 -
33.33.33.1 - - [05/Aug/2015 16:40:20] "GET /code/search?q=&redirect=false&case=false&limit=100&offset=0 HTTP/1.1" 200 -
```

erikrose:
r+

PR #456, opened 2015-07-16
XPIDL support

pelmers:
Add plugin for XPIDL files.

Riding along is syntax highlighting for IDL, XUL, and moz.build files.

### DIFF
--- a/dxr/plugins/xpidl/__init__.py
+++ b/dxr/plugins/xpidl/__init__.py
@@ -1,0 +1,28 @@
+from functools import partial
+from os.path import abspath
+
+from schema import Optional, Use, And
+
+from dxr.config import AbsPath
+from dxr.plugins import Plugin, AdHocTreeToIndex, filters_from_namespace
+from dxr.plugins.xpidl import filters
+from dxr.plugins.xpidl.mappings import mappings
+from dxr.plugins.xpidl.indexers import FileToIndex
+
+
+def split_on_colon_into_abspaths(value):
+    return map(abspath, value.strip().split(':'))
+
+
+ColonPathList = And(basestring,
+                    Use(split_on_colon_into_abspaths),
+                    error='This should be a colon-separated list of paths.')
+
+plugin = Plugin(
+    tree_to_index=partial(AdHocTreeToIndex,
+                          file_to_index_class=FileToIndex),
+    filters=filters_from_namespace(filters.__dict__),
+    mappings=mappings,
+    config_schema={
+        'header_bucket': AbsPath,
+        Optional('include_folders', default=[]): ColonPathList})
--- a/dxr/plugins/xpidl/__init__.py
+++ b/dxr/plugins/xpidl/__init__.py
@@ -36,5 +36,5 @@ plugin = Plugin(
     filters=filters_from_namespace(filters.__dict__),
     mappings=mappings,
     config_schema={
-        'header_bucket': AbsPath,
+        'header_path': AbsPath,
         Optional('include_folders', default=[]): ColonPathList})
--- a/dxr/plugins/xpidl/__init__.py
+++ b/dxr/plugins/xpidl/__init__.py
@@ -1,3 +1,15 @@
+"""XPIDL plugin: analyze XPIDL files using the mozilla-central xpidl parser.
+
+This plugin analyzes XPIDL files by attempting to parse all files ending in '.idl' using the
+xpidl parser. Then it delicately walks the productions in the AST to pull out relevant
+information, such as interface and member declarations, and shoves them into refs and needles to
+feed to ES. The plugin borrows the xpidl C++ header generating code to find direct hops from
+interface constructs to their corresponding output in header land. It also donates Filters which
+use the discovered needles to perform structural queries.
+
+For further reference, see https://developer.mozilla.org/en-US/docs/Mozilla/XPIDL.
+"""
+
 from functools import partial
 from os.path import abspath
 
@@ -10,13 +22,13 @@ from dxr.plugins.xpidl.mappings import mappings
 from dxr.plugins.xpidl.indexers import FileToIndex
 
 
-def split_on_colon_into_abspaths(value):
-    return map(abspath, value.strip().split(':'))
+def split_on_space_into_abspaths(value):
+    return map(abspath, value.strip().split())
 
 
 ColonPathList = And(basestring,
-                    Use(split_on_colon_into_abspaths),
-                    error='This should be a colon-separated list of paths.')
+                    Use(split_on_space_into_abspaths),
+                    error='This should be a space-separated list of paths.')
 
 plugin = Plugin(
     tree_to_index=partial(AdHocTreeToIndex,
--- a/dxr/plugins/xpidl/indexers.py
+++ b/dxr/plugins/xpidl/indexers.py
@@ -1,0 +1,45 @@
+from os.path import join
+
+from xpidl.xpidl import IDLParser, IDLError
+
+import dxr.indexers
+from dxr.indexers import iterable_per_line, with_start_and_end, split_into_lines
+from dxr.plugins.xpidl.filters import PLUGIN_NAME
+from dxr.plugins.xpidl.visitor import IdlVisitor
+
+
+class FileToIndex(dxr.indexers.FileToIndex):
+    def __init__(self, path, contents, plugin_name, tree):
+        super(FileToIndex, self).__init__(path, contents, plugin_name, tree)
+        self.temp_folder = join(self.tree.temp_folder, 'plugins', PLUGIN_NAME)
+        self.parser = IDLParser(self.temp_folder)
+        self._idl = None
+        self._idl_exception = False
+
+    @property
+    def idl(self):
+        """Parse the IDL file and resolve deps."""
+        # Don't try again if we already excepted.
+        if not self._idl and not self._idl_exception:
+            try:
+                self._idl = IdlVisitor(self.parser, self.contents, self.path, self.absolute_path(),
+                                       self.plugin_config.include_folders,
+                                       self.plugin_config.header_bucket, self.tree)
+            except IDLError:
+                self._idl_exception = True
+        return self._idl
+
+    def is_interesting(self):
+        # TODO: consider adding a link from generated headers back to the idl
+        return self.path.endswith('.idl')
+
+    def links(self):
+        if self.idl:
+            yield (3, 'IDL', [('idl-header', self.idl.header_filename, self.idl.generated_url)])
+
+    def refs(self):
+        return self.idl.refs if self.idl else []
+
+    def needles_by_line(self):
+        return iterable_per_line(
+            with_start_and_end(split_into_lines(self.idl.needles if self.idl else [])))
--- a/dxr/plugins/xpidl/indexers.py
+++ b/dxr/plugins/xpidl/indexers.py
@@ -24,9 +24,10 @@ class FileToIndex(dxr.indexers.FileToIndex):
         # Don't try again if we already excepted.
         if not self._idl and not self._had_idl_exception:
             try:
-                self._idl = IdlVisitor(self.parser, self.contents, self.path, self.absolute_path(),
+                self._idl = IdlVisitor(self.parser, self.contents, self.contents.splitlines(True),
+                                       self.path, self.absolute_path(),
                                        self.plugin_config.include_folders,
-                                       self.plugin_config.header_bucket, self.tree)
+                                       self.plugin_config.header_path, self.tree)
             except IDLError:
                 self._had_idl_exception = True
         return self._idl
--- a/dxr/plugins/xpidl/indexers.py
+++ b/dxr/plugins/xpidl/indexers.py
@@ -1,5 +1,4 @@
 from os.path import join
-
 from xpidl.xpidl import IDLParser, IDLError
 
 import dxr.indexers
@@ -14,19 +13,22 @@ class FileToIndex(dxr.indexers.FileToIndex):
         self.temp_folder = join(self.tree.temp_folder, 'plugins', PLUGIN_NAME)
         self.parser = IDLParser(self.temp_folder)
         self._idl = None
-        self._idl_exception = False
+        self._had_idl_exception = False
 
     @property
     def idl(self):
-        """Parse the IDL file and resolve deps."""
+        """Parse the IDL file and resolve dependencies. If successful, return an IdlVisitor
+        object which has visited the AST and has refs and needles ready for ES. Otherwise,
+        on exception, return None."""
+
         # Don't try again if we already excepted.
-        if not self._idl and not self._idl_exception:
+        if not self._idl and not self._had_idl_exception:
             try:
                 self._idl = IdlVisitor(self.parser, self.contents, self.path, self.absolute_path(),
                                        self.plugin_config.include_folders,
                                        self.plugin_config.header_bucket, self.tree)
             except IDLError:
-                self._idl_exception = True
+                self._had_idl_exception = True
         return self._idl
 
     def is_interesting(self):
@@ -35,7 +37,7 @@ class FileToIndex(dxr.indexers.FileToIndex):
 
     def links(self):
         if self.idl:
-            yield (3, 'IDL', [('idl-header', self.idl.header_filename, self.idl.generated_url)])
+            yield 3, 'IDL', [('idl-header', self.idl.header_filename, self.idl.generated_url)]
 
     def refs(self):
         return self.idl.refs if self.idl else []
--- a/dxr/plugins/xpidl/mappings.py
+++ b/dxr/plugins/xpidl/mappings.py
@@ -1,0 +1,29 @@
+from dxr.filters import LINE
+from dxr.indexers import STRING_PROPERTY
+
+# Like qualified line needle, but has no qualname
+LINE_NEEDLE = {
+    'type': 'object',
+    'properties': {
+        'name': STRING_PROPERTY,
+        'start': {
+            'type': 'integer',
+            'index': 'no'  # just for highlighting
+        },
+        'end': {
+            'type': 'integer',
+            'index': 'no'
+        }
+    }
+}
+
+mappings = {
+    LINE: {
+        'properties': {
+            'xpidl_var_decl': LINE_NEEDLE,
+            'xpidl_function_decl': LINE_NEEDLE,
+            'xpidl_derived': LINE_NEEDLE,
+            'xpidl_type_decl': LINE_NEEDLE
+        }
+    }
+}
--- a/dxr/plugins/xpidl/tests/test_menus_and_filters/code/Makefile
+++ b/dxr/plugins/xpidl/tests/test_menus_and_filters/code/Makefile
@@ -1,0 +1,6 @@
+all:
+	mkdir -p obj
+	header.py -o obj/nsrootidl.h --cachedir=obj nsrootidl.idl
+	header.py -o obj/nsIDownloadHistory.h --cachedir=obj nsIDownloadHistory.idl
+clean:
+	rm -rf obj
--- a/dxr/plugins/xpidl/tests/test_menus_and_filters/code/Makefile
+++ b/dxr/plugins/xpidl/tests/test_menus_and_filters/code/Makefile
@@ -1,6 +1,12 @@
-all:
-	mkdir -p obj
-	header.py -o obj/nsrootidl.h --cachedir=obj nsrootidl.idl
-	header.py -o obj/nsIDownloadHistory.h --cachedir=obj nsIDownloadHistory.idl
+all: obj/nsrootidl.h obj/nsIDownloadHistory.h
+
+obj/%.h: %.idl | obj
+	header.py -o $@ --cachedir=obj $^
+
+obj:
+	mkdir $@
+
 clean:
-	rm -rf obj
+	$(RM) -r obj
+
+.PHONY: all clean
--- a/dxr/plugins/xpidl/tests/test_menus_and_filters/code/nsIDownloadHistory.idl
+++ b/dxr/plugins/xpidl/tests/test_menus_and_filters/code/nsIDownloadHistory.idl
@@ -1,0 +1,65 @@
+/* -*- Mode: C++; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*-
+ * vim: sw=2 ts=2 sts=2
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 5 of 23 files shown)


setup.py
```diff
@@ -40,6 +41,7 @@
                       'pyelasticsearch>=1.1,<2.0',
                       'Pygments>=2.0.1,<3.0',
                       'python-hglib>=1.6,<2.0',
+                      'xpidl>=1.0',
```
erikrose: 1. This isn't in PyPI yet, so listing it here will just cause 404s. Let's upload it to PyPI.
2. We'll also need to add this to requirements.txt so it will be pinned to a hash.

dxr/plugins/xpidl/tests/test_menus_and_filters/code/Makefile
```diff
@@ -0,0 +1,6 @@
+all:
+	mkdir -p obj
```
erikrose: No need for -p since it's just making 1 folder

dxr/plugins/xpidl/tests/test_menus_and_filters/test_filters.py
```diff
@@ -0,0 +1,28 @@
+from dxr.testing import DxrInstanceTestCase
+
+
+class IdlFilterQueryTest(DxrInstanceTestCase):
+    """Test query filters for Idl structure."""
+
+    def test_derived(self):
+        self.found_line_eq('+derived:nsISupports',
```
erikrose: Any particular reason these are qualified ("+") queries?

dxr/plugins/xpidl/tests/test_smoke/test_smoke.py
```diff
@@ -0,0 +1,10 @@
+# Index a big idl file.
```
erikrose: Why not a module-level docstring?

dxr/plugins/xpidl/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+from dxr.plugins.xpidl import filters
+from dxr.plugins.xpidl.mappings import mappings
+from dxr.plugins.xpidl.indexers import FileToIndex
+
+
+def split_on_colon_into_abspaths(value):
+    return map(abspath, value.strip().split(':'))
+
+
+ColonPathList = And(basestring,
```
erikrose: Why make up a new thing when you can just do what ignore_patterns does? (See the explanation in configuration.rst.)

dxr/plugins/xpidl/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+                    Use(split_on_colon_into_abspaths),
+                    error='This should be a colon-separated list of paths.')
+
+plugin = Plugin(
+    tree_to_index=partial(AdHocTreeToIndex,
```
erikrose: I think you should be able to leave this out altogether, and it'll find FileToIndex by name. I never meant people to explicitly use AdHocFileToIndex.

dxr/plugins/xpidl/indexers.py
```diff
@@ -0,0 +1,45 @@
...
+                self._idl_exception = True
+        return self._idl
+
+    def is_interesting(self):
+        # TODO: consider adding a link from generated headers back to the idl
+        return self.path.endswith('.idl')
+
+    def links(self):
+        if self.idl:
+            yield (3, 'IDL', [('idl-header', self.idl.header_filename, self.idl.generated_url)])
```
erikrose: If you're not going to wrap the line, you can omit the outer parentheses.

dxr/plugins/xpidl/mappings.py
```diff
@@ -0,0 +1,29 @@
...
+            'type': 'integer',
+            'index': 'no'
+        }
+    }
+}
+
+mappings = {
+    LINE: {
+        'properties': {
+            'xpidl_var_decl': LINE_NEEDLE,
```
erikrose: Our convention until now has been to use file extensions as the language prefix. Is "idl" perhaps shared with some of the other IDL variants?

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,244 @@
+from cStringIO import StringIO
+from itertools import ifilter
+from os.path import relpath, join, basename, dirname, exists
+
+from flask import url_for
+from xpidl.header import idl_basename, header, include, jsvalue_include, \
```
erikrose: The backslash has fallen out of fashion. Everybody uses parens to wrap these days.

dxr/plugins/xpidl/__init__.py
```diff
@@ -0,0 +1,28 @@
+from functools import partial
```
erikrose: Mark this day on your calendar: I'm going to recommend you do something like the clang plugin. ;-) I could really use a little docstring here describing the general approach to analyzing IDL.

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,244 @@
...
+            line += number_lines("typedef %s %s;\n\n" % (p.realtype.nativeType('in'),
+                                                         p.name))
+
+    return production_map
+
+
+class IdlVisitor(object):
+    """Traverse an IDL syntax tree and collect refs and needles."""
+
+    def __init__(self, parser, contents, rel_path, abs_path, include_folders, header_bucket,
```
erikrose: I could use a docstring. What are rel_path and abs_path of, especially?

dxr/plugins/xpidl/indexers.py
```diff
@@ -0,0 +1,45 @@
...
+from dxr.plugins.xpidl.visitor import IdlVisitor
+
+
+class FileToIndex(dxr.indexers.FileToIndex):
+    def __init__(self, path, contents, plugin_name, tree):
+        super(FileToIndex, self).__init__(path, contents, plugin_name, tree)
+        self.temp_folder = join(self.tree.temp_folder, 'plugins', PLUGIN_NAME)
+        self.parser = IDLParser(self.temp_folder)
+        self._idl = None
+        self._idl_exception = False
```
erikrose: I expected this to be an actual exception based on the name. Maybe `_had_idl_exception`?

dxr/plugins/xpidl/indexers.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def idl(self):
+        """Parse the IDL file and resolve deps."""
```
erikrose: What does it return? What does it return if it has an exception?

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,244 @@
...
+                 tree_config):
+        self.tree = tree_config
+        # Hold on to the URL so we do not have to regenerate it everywhere.
+        self.header_filename = basename(rel_path.replace('.idl', '.h'))
+        header_path = relpath(join(header_bucket, self.header_filename),
+                              self.tree.source_folder)
+        self.generated_url = url_for('.browse',
+                                     tree=self.tree.name,
+                                     path=header_path)
+        self.ast = parser.parse(contents, basename(rel_path))
```
erikrose: Why an instance var? It doesn't look like it's used outside this function.

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,244 @@
...
+        self.ast.resolve(self.search_paths, parser)
+        self.line_map = header_line_numbers(self.ast, self.header_filename)
+        self.line_list = contents.splitlines()
+        # List of (start, end, Ref) where start and end are byte offsets into the file.
+        self.refs = []
+        self.needles = []
+
+        # Initiate visitations.
+        for item in self.ast.productions:
+            if item.kind == 'include':
```
erikrose: When I see a big stack of elifs like this, I wonder: could this be a dict?

dxr/plugins/xpidl/tests/test_menus_and_filters/code/nsIDownloadHistory.idl
```diff
@@ -0,0 +1,65 @@
+/* -*- Mode: C++; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*-
```
erikrose: How do you feel about the size and sharedness of this test harness? Generally, except for smoke tests, I've tried to keep test cases minimal and unshared, so we know we aren't invalidating tests if we have to change the code later.

erikrose:
To be continued, starting with the IdlVisitor…

dxr/plugins/xpidl/tests/test_menus_and_filters/test_filters.py
pelmers: No particular reason; this was just an example of doing it that way on the first case and then copying it to the others. I'll remove them, since they're not qualified.

dxr/plugins/xpidl/__init__.py
pelmers: Preferably I would like some copy of WhitespaceList and AbsPath, is there a way to combine the two?

dxr/plugins/xpidl/tests/test_menus_and_filters/code/nsIDownloadHistory.idl
pelmers: I think the file looks bigger than it actually is because of the doc comments; it has 2 methods, 1 attribute, and 1 const member. I'll go ahead and shorten the file while keeping the same interesting points.

dxr/plugins/xpidl/tests/test_menus_and_filters/code/Makefile
pelmers: mkdir will fail if obj already exists, however.

setup.py
pelmers: Here we go: https://pypi.python.org/pypi/xpidl/1.0

dxr/plugins/xpidl/mappings.py
pelmers: Yes, not all .idl files are xpidl files -- for example, some are MSIDL (?), Microsoft's IDL-ish variant: https://dxr.allizom.org/mozilla-central/search?q=%22%3A+IUnknown%22+ext%3Aidl

dxr/plugins/xpidl/tests/test_menus_and_filters/code/Makefile
erikrose: Good point.

dxr/plugins/xpidl/__init__.py
erikrose: Never mind; I misremembered. It doesn't look for things by name unless the whole Plugin is being constructed automatically.

dxr/plugins/xpidl/mappings.py
erikrose: Then this is perfect. Carry on. :-)

dxr/plugins/xpidl/__init__.py
erikrose: Sure, it'd be something along these lines:
```python

AbsPathWhitespaceList = And(basestring,
                     Use(lambda value: map(abspath, value.strip().split())),
                     error='This should be a whitespace-separated list of paths.')
```
You might even be able to repeat yourself less than that.

pelmers: I actually had this at first, but cPython cannot pickle anonymous functions, so this failed on using multiple workers, so I made it into the very_verbose_function_name.

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+        Parse an IDL file and visit the productions of its AST.
+
+        :arg parser: IdlParser to use to parse the file
+        :arg contents: the contents of the file, as a string
+        :arg rel_path: relative path to the file from tree root
+        :arg abs_path: absolute path to the file
+        :arg include_folders: list of folders to use to resolve include directives
+        :arg header_bucket: folder in which the generated header would be placed to make links
+        :arg tree_config: A :class:`TreeConfig` object for the current tree
+        """
```
erikrose: one mo' space

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+        # Hold on to the URL so we do not have to regenerate it everywhere.
+        self.header_filename = basename(rel_path.replace('.idl', '.h'))
+        header_path = relpath(join(header_bucket, self.header_filename),
+                              self.tree.source_folder)
+        self.generated_url = url_for('.browse',
+                                     tree=self.tree.name,
+                                     path=header_path)
+        ast = parser.parse(contents, basename(rel_path))
+        # Might raise IdlError
+        self.search_paths = [dirname(abs_path)] + include_folders
```
erikrose: It'd be nice to have comments here about the assumptions we're making. For example, here we seem to say "first we'll search the directory where an XPIDL file sits, and then we'll look at the explicitly provided include folders".

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+    """Return the last byte position on which we can find name in the
+    location's line, resolving if necessary."""
+
+    location.resolve()
+    return location._line.rfind(name) - location._colno + location._lexpos
+
+
+def header_line_numbers(idl, filename):
+    """Map each production to its line number in the header."""
+    # Basically I replace all fd.write in print_header with line increments
```
erikrose: Can you make more clear that you ripped this off from elsewhere, where that is, and that it's not your fault? ;-)

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+                              self.tree.source_folder)
+        self.generated_url = url_for('.browse',
+                                     tree=self.tree.name,
+                                     path=header_path)
+        ast = parser.parse(contents, basename(rel_path))
+        # Might raise IdlError
+        self.search_paths = [dirname(abs_path)] + include_folders
+        ast.resolve(self.search_paths, parser)
+        self.line_map = header_line_numbers(ast, self.header_filename)
+        self.line_list = contents.splitlines()
```
erikrose: Maybe we can just pass in the already-split lines we have floating around in the FileToIndex. (That splits using universal newlines, IIRC, and this doesn't.)

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+        # List of (start, end, Ref) where start and end are byte offsets into the file.
+        self.refs = []
+        self.needles = []
+
+        # Initiate visitations.
```
erikrose: Or "visit" :-)

dxr/plugins/xpidl/visitor.py
erikrose: It would also be worth mentioning how you get data back *out* of this thing.

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+        for item in ast.productions:
+            try:
+                getattr(self, 'visit_' + item.kind)(item)
+            except AttributeError:
```
erikrose: Scary as hell. We'll catch AttributeErrors that the methods themselves throw. Let's choose the method inside the try and then run it in the `else` or something.

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+        start_col = location._line.rfind(name)
+        lineno = self.check_lineno(name, location._lineno)
+        if lineno:
+            return Extent(Position(lineno + 1, start_col),
+                          Position(lineno + 1, start_col + len(name)))
+
+    def yield_needle(self, name, mapping, extent):
+        self.needles.append((PLUGIN_NAME + '_' + name, mapping, extent))
+
+    def yield_name_needle(self, filter_name, name, location):
```
erikrose: I wouldn't mind a docstring.

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+        extent = self.make_extent(name, location)
+        if extent:
+            self.yield_needle(filter_name, {'name': name}, extent)
+
+    def yield_ref(self, start, end, menus):
+        self.refs.append((start, end, Ref(menus)))
+
+    def generated_menu(self, production):
+        # Return a menu for jumping to corresponding C++ source using the line map.
```
erikrose: Why not a docstring?

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+from dxr.plugins.xpidl.filters import PLUGIN_NAME
+from dxr.utils import search_url
+
+
+def start_pos(name, location):
+    """Return the last byte position on which we can find name in the
+    location's line, resolving if necessary."""
+
+    location.resolve()
+    return location._line.rfind(name) - location._colno + location._lexpos
```
erikrose: So, does this really work, or will it get fouled up by occurrences of the name in a comment, for instance?

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+                                      Extent(Position(lineno + 1, member.location._colno),
+                                             Position(lineno + 1,
+                                                      member.location._colno + len(member.name))))
+
+    def visit_include(self, item):
+        filename = item.filename
+        start = start_pos(filename, item.location)
+        # Remark: it would be nice if the Include class held on to this data after it does the
+        # same work.
+        # We know that the parser resolved the include in the same way, so we assume we can
```
erikrose: Yep, this is a likely place to optimize in the future, if profiles bear it out. It's a lot of repeated IO.

dxr/plugins/xpidl/visitor.py
pelmers: Badly, but good enough 95% of the time. 😔

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+        # too and take the first element.
+        resolved_path = next(ifilter(exists, (join(path, filename) for path in self.search_paths)))
+        self.yield_ref(start, start + len(filename), [self.include_menu(resolved_path)])
+
+    def visit_typedef(self, item):
+        start = start_pos(item.name, item.location)
+        self.yield_ref(start, start + len(item.name), [self.generated_menu(item)])
```
erikrose: All these things functions end up doing len(thing) and adding that to `start`. Would we be better off having start_pos() return both a start pos and an end pos?

dxr/plugins/xpidl/visitor.py
```diff
@@ -0,0 +1,252 @@
...
+            'html': 'Jump to file',
+            'title': 'Go to the target of the include statement',
+            'href': url_for('.browse', tree=self.tree.name,
+                            path=relpath(item_path, self.tree.source_folder)),
+            'icon': 'jump'
+        }
+
+    def visit_interface(self, interface):
+        # Yield refs for the members, methods, etc. of an interface (and the interface itself).
+        start = start_pos(interface.name, interface.location)
```
erikrose: Unread?

erikrose: No, but we might as well move it down closer to where it is read.

dxr/plugins/xpidl/visitor.py
erikrose: And once we do that, it may make sense to have yield_ref() call it—we'll see. But right now, every time we call start_pos(), it's pretty much to pass the result into yield_ref().

tests/test_config.py
```diff
@@ -24,6 +24,9 @@ def test_enabled_star():
 
             [[python]]
             python_path = /some/path
+
+            [[xpidl]]
+            header_bucket = /somewhere
```
erikrose: Is "bucket" a term commonly used with XPIDLs, or should we change this to "path" for consistency with other plugins (like "python_path")?

erikrose:
Let's add a menu item for implementations of interfaces.

tests/test_config.py
pelmers: Ah, probably makes sense to say path. My mental imagery is that there's a bucket and we toss headers in.

dxr/plugins/xpidl/visitor.py
pelmers: I'll factor all the line splitting of file contents into a util function, so we can be sure we call the same method each time.

erikrose:
Let's add docs for the config option in configuration.rst.

dxr/plugins/xpidl/indexers.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # Don't try again if we already excepted.
+        if not self._idl and not self._had_idl_exception:
+            try:
+                self._idl = IdlVisitor(self.parser, self.contents, self.contents.splitlines(True),
+                                       self.path, self.absolute_path(),
+                                       self.plugin_config.include_folders,
+                                       self.plugin_config.header_path, self.tree)
+            except IDLError:
+                self._had_idl_exception = True
```
erikrose: It looks like nobody ever reads this. Shouldn't we at least emit a warning so we know to fix the parser?

requirements.txt
```diff
@@ -66,9 +66,15 @@ Werkzeug==0.10.4
 # sha256: uyTo_df6vGRca0giqfa20IF1End-YHo_-ko4rSI_kgY
 funcy==1.1
 
+# sha256: rUar6kB6nbKhCxgUh53YtF7Zxb5XqGPRnKv2Od6rIWE
+https://bitbucket.org/pelmers/python-hglib/get/issue3924-version-fix.tar.gz#egg=python-hglib
```
erikrose: Make sure this merged properly to give us the new upstream version.

erikrose:
With those 3 things, r+

dxr/plugins/xpidl/indexers.py
pelmers: There are files on moz-central ending in .idl that are not xpidl files. As far as I understand, these are mostly MSIDL. A search that reveals most of these: [dxr.mozilla.org](https://dxr.mozilla.org/mozilla-central/search?q=ext%3Aidl+-path%3Aobj-x86_64-unknown-linux-gnu%2Fdist%2Fidl%2F+path%3Aobj-x86_64-unknown-linux-gnu%2F&redirect=true&case=true&limit=78&offset=0)

pelmers: Do you think it would be better to add a warning message when this happens?
warn("IDL parser raised an exception on {path}: {message}")

erikrose: Meh, if we can't get our extensions right, let's leave it alone. Later, maybe we can come back and have it actually look at the contents of the file to determine what kind of IDL file it is.

PR #458, opened 2015-07-21
Fix bug 1185687, stop showing symlinks in search results.

pelmers:
New behaviors:
We do not call the by lines processing for symlinks at index time.
Symlinks do not appear in search results (lines and files).
A symlink in folder browsing will have its href set to the real target.
A direct link to symlink will redirect to its real target.

### DIFF
--- a/dxr/app.py
+++ b/dxr/app.py
@@ -209,15 +209,18 @@ def browse(tree, path=''):
         return _browse_folder(tree, path.rstrip('/'), config)
     except NotFound:
         frozen = frozen_config(tree)
-        # Grab the FILE doc, just for the sidebar nav links:
+        # Grab the FILE doc, just for the sidebar nav links and the symlink target:
         files = filtered_query(
             frozen['es_alias'],
             FILE,
             filter={'path': path},
             size=1,
-            include=['links'])
+            include=['link', 'links'])
         if not files:
             raise NotFound
+        if 'link' in files[0]:
+            # Then this path is a symlink, so redirect to the real thing.
+            return redirect(url_for('.browse', tree=tree, path=files[0]['link'][0]))
 
         lines = filtered_query(
             frozen['es_alias'],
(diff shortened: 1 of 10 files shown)


dxr/app.py
```diff
@@ -209,15 +209,18 @@ def browse(tree, path=''):
         return _browse_folder(tree, path.rstrip('/'), config)
```
erikrose: I believe we'll want to do this for folders as well: after all, a symlink's target can be a folder.

erikrose: oops, nm

dxr/build.py
```diff
@@ -475,7 +475,7 @@ def index_file(tree, tree_indexers, path, es, index):
             linkses.append(file_to_index.links())
 
             # Per-line stuff:
-            if is_text:
+            if is_text and not islink(path):
```
erikrose: Let's pull the islink() call out of the `for` loop and save some `stat` calls. Also, did you look at the other places in this function where we say `if is_text`? If we don't want to insert LINE needles, regions, and refs for symlinks, we might want to add the islink() condition there, too (and probably factor `islink(…) and is_text` up).

erikrose:
Remember to explain that we delegate islink() caring to is_interesting() of the plugins so they can individually decide whether to care about links. No per-line needles should be emitted for symlinks, since we're maintaining the illusion that they have no contents of their own. But plugins should be able to decide whether links get per-file needles. They should get path and ext ones, since, after all, they have paths and extensions, but they should not get, say, "size" needles from a theoretical image-size plugin: only the original image should.

erikrose:
Let's actually squash these 2 commits, since the 2nd does more than its commit msg says.

erikrose:
Also, go through and audit any implementations of is_interesting() and make sure they do the right thing with links: the Python plugin, for example.

erikrose:
And let's not emit links() either.

dxr/plugins/python/indexers.py
```diff
@@ -1,7 +1,8 @@
 import ast
 import token
 import tokenize
-from StringIO import StringIO
+from os.path import islink
+from cStringIO import StringIO
```
pelmers: I switched out the import here since I happened to see it. Should I back this out since it's not related?

erikrose: No, this is small enough to sneak in.

erikrose:
Let's factor up the `islink(self.absolute_path())` occurrences to make a FileToIndex.is_link(). It saves a little code, prods plugin implementors to consider links, and gives us the option to easily start caching the result later.

erikrose:
Then, r+.

PR #459, opened 2015-07-27
Include more menu refs in the Python plugin.

pelmers:
Playing around with the Python plugin, I didn't even realize that we had many function filters until I checked the ES mapping. I think that putting ref menus on things makes them more discoverable, so here I add refs for function definitions and function calls.

Add refs to find function usages and definitions to call sites and definitions.
Use id: and ref: filters in the calls because Python’s AST does not
differentiate between class instantiation and function calls — both are Calls.

### DIFF
--- a/dxr/plugins/python/filters.py
+++ b/dxr/plugins/python/filters.py
@@ -30,6 +30,12 @@ class FunctionFilter(_PyFilter):
     description = Markup('Function or method definition: <code>function:foo</code>')
 
 
+class FunctionRefFilter(_PyFilter):
+    name = 'function-ref'
+    is_reference = True
+    description = Markup('Function or method reference: <code>function-ref:foo</code>')
+
+
 class DerivedFilter(_QualifiedPyFilter):
     name = 'derived'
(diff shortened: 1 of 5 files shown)


pelmers:
Remark: we should probably hold off on this until we're happy with the design of request-time menu creation.

erikrose:
Remember that there's further inaccuracy introduced because structural search are also cross-language. So ref:baz will turn up Python identifiers of all kinds as well as C++ and Rust ones. Hardly ideal. We should give this some thought. Perhaps we need language-specific filters like `py-ref:baz` or that language metafilter I've had in mind (`ref:baz lang:py`). Or maybe we should just go with `function-ref:` rather than `ref:`. Was your aim with `ref:` to not leave out non-call mentions of functions?

pelmers:
Good point on how we could pick up too many results using id and ref filters. My intention in using 'ref' and 'id' is that we see class constructions as function calls, but classes and functions are differentiated at definition time.

I think I'll give this PR some more massage, and make an update when I can avoid these issues.

PR #460, opened 2015-07-29
Remove case sensitivity checkbox.

pelmers:
If any text is uppercase, then assume case sensitive.
Otherwise, assume it’s insensitive. Use an @ sigil
in front of a text term to force case sensitive.

This also gives us per-term case sensitivity.

I've been thinking about using showBubble to remark when we guess case sensitive.

Also, it collides slightly with path promotion, and it may be better to finish that first.


dxr/query.py
```diff
@@ -326,7 +326,9 @@ def visit_text(self, text, ((some_text,), _)):
         ``visit_filtered_term`` will overrule us later.
 
         """
-        return {'name': 'text', 'arg': some_text}
+        # Case-sensitive if there's any uppercase characters in the term.
+        case_sensitive = any((c.isupper() for c in some_text))
```
erikrose: Right choice! If you had required "all lowercase", unicode would have caused a disaster.

dxr/static/js/context_menu.js
```diff
@@ -120,7 +120,7 @@ $(function() {
 
             // If the regex did not find a start index, start from index 0
             if (startIndex === -1) {
-                start = 0;
+                startIndex = 0;
```
erikrose: Good catch!

dxr/query.py
```diff
@@ -313,12 +307,18 @@ def visit_not_term(self, not_term, (not_, term_dict)):
...
+    def visit_filtered_term(self, filtered_term, (plus, filter, colon, (term_dict,))):
         """Add fully-qualified indicator and the filter name to the term_dict."""
         term_dict['qualified'] = plus.text == '+'
         term_dict['name'] = filter.text
         return term_dict
 
+    def visit_cased_text(self, cased_text, (at_, term_dict)):
+        """Flip case_sensitive bit in the term_dict."""
+        # Negate our case sensitivity guess if we prepend an @.
+        term_dict['case_sensitive'] = not term_dict['case_sensitive']
```
erikrose: It's strictly less powerful but (a few of us think) quite a bit easier to keep in your head if we make @ mean "case sensitive" rather than "the opposite of what I guessed". What do you think?

(We could add a "case insensitive" operator later if necessary and/or some UI to ignore case on various terms.)

erikrose:
Note to self: consider whether to use quotes instead of @ for case sensitivity.

PR #462, opened 2015-07-31
Make es_index per tree, fixes bug 1177113.

pelmers:
If es_index is provided for a tree, use that;
otherwise default to the original format.

https://bugzilla.mozilla.org/show_bug.cgi?id=1177113

### DIFF
--- a/dxr/config.py
+++ b/dxr/config.py
@@ -114,6 +114,7 @@ class Config(DotSection):
                 Optional('google_analytics_key', default=''): basestring,
                 Optional('es_hosts', default='http://127.0.0.1:9200/'):
                     WhitespaceList,
+                # A semi-random name, having the tree name and format version in it.
                 Optional('es_index', default='dxr_{format}_{tree}_{unique}'):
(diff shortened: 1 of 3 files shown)


dxr/config.py
```diff
@@ -204,6 +205,7 @@ def __init__(self, name, unvalidated_tree, sections, config):
             Optional('description', default=''): basestring,
             Optional('disabled_plugins', default=plugin_list('')): Plugins,
             Optional('enabled_plugins', default=plugin_list('*')): Plugins,
+            Optional('es_index', default=config.es_index): basestring,
```
erikrose: Huh, I guess we can get away with referencing `config` directly in the `Optional`, if we assume the Config to be immutable (which it is, for now). So this is fine. (Note that we couldn't reference `self`, though, since a missing attribute would crash us during Schema instantiation, before we ever got the opportunity to raise a nice ConfigError.)

erikrose:
r+. Especially nice tests: really covered all the bases

PR #463, opened 2015-08-03
Add a SyntaxFilter to pygmentize plugin, bug 1190587

pelmers:
*Incomplete!* I'm putting this up as RFC.

Some notes/questions:
- Filtering by region filters for all the lines that contain the region. So if we want to search within a region, the filter needs two pieces of information: 1) the region and 2) the text. If we just did a region filter, then a text search, we would match the text on any part of the line that contains the region.
- What should we call the filter?
- How does the class.text filter syntax feel? Is there a nicer way to combine two pieces of information in the filter?

What's not done:
- Cleaning up the code.
- Adding tests, docstrings.
- Figuring out why all the search is lowercase (or something like that, I'm not really handling case right now).


pelmers:
<sup>I'll be back!</sup>

PR #464, opened 2015-08-11
Keep core plugin from double-contributing. Fixes bug 1191537.

erikrose:
core would get included in the frozen config's enabled_plugins, and then TreeConfig would happily re-add core again at request time. We no longer freeze 'core' as part of an index's catalog entry.

all_plugins() no longer returns core, and neither does plugins_named() (which has been renamed to provide a reminder). core is no longer part of Config's enabled_plugins, either. The only place core comes from now is a TreeConfig's enabled_plugins attr.

Also, make skimmers run in the order that plugins are specified in the config file.

### DIFF
--- a/dxr/app.py
+++ b/dxr/app.py
@@ -26,7 +26,7 @@ from dxr.filters import FILE, LINE
 from dxr.indexers import Ref, Region
 from dxr.lines import html_line, tags_per_line, finished_tags
 from dxr.mime import icon, is_image, is_text
-from dxr.plugins import plugins_named, all_plugins
+from dxr.plugins import plugins_named
 from dxr.query import Query, filter_menu_items
 from dxr.utils import (non_negative_int, decode_es_datetime, DXR_BLUEPRINT,
                        format_number, append_update, append_by_line, cumulative_sum)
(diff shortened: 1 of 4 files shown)


pelmers:
Good catch on the duplication of core in the plugin list; however I think that the explanation is incomplete. Since the permalink happens in a FileToIndex, duplicating the plugin at request time wouldn't cause two of them to appear.

Rather, the root of this issue was that we are using enabled_plugins = * (which I hadn't thought to test locally). Notice that we populate from all_plugins [here](https://github.com/mozilla/dxr/blob/b1f7c3d7e3336e43ef3d2113a52708216fd3d82e/dxr/config.py#L158-L164) first, then follow that up by inserting 'core' [here too](https://github.com/mozilla/dxr/blob/b1f7c3d7e3336e43ef3d2113a52708216fd3d82e/dxr/config.py#L242).

In conclusion, there's probably three copies of 'core' floating around in the cloud right now. :)

pelmers:
Besides adding a little note in the explanation, lgtm.
r+

pelmers:
Made up a quick regression test: https://gist.github.com/pelmers/ba4c2b082ac82a6ed55c

erikrose:
Great comments. Adding your test. Also caught another mistake in my commit message: I claim plugins_named() no longer returns core. *fix fix*

erikrose:
There. I took a second swing at it. It changes less, includes a variation of your test, and doesn't change the format version.

PR #465, opened 2015-08-17
Optimize indexing by deferring menu generation to request time.

erikrose:
The intent is to make clang indexing faster. Benchmarks on the earlier, menu-maker-based version of this showed a 25-50% speed boost.


dxr/plugins/rust/refs.py
```diff
@@ -0,0 +1,339 @@
...
+    def prepare_menu_data(self, tree_index, datum):
+        return trim_dict(datum, self.keys)
+
+
+class FunctionRef(_RustRef):
+    def prepare_menu_data(self, tree_index, datum):
+        if 'declid' in datum and datum['declid'] in tree_index.data.functions:
+            # It's an implementation; find the decl:
+            decl = tree_index.data.functions[datum['declid']]
+            return [trim_dict(datum, ['qualname']), trim_dict(decl, ['file_name', 'file_line']), 0]
```
erikrose: (Porting this comment from the older PR:) Why use a dict to hold the qualname here? Why not just spit out the string?

pelmers: Good point, it's an easy adjustment to only save qualname. generic_function_menu should really only take qualname as an argument, rather than a datum dict where it only uses qualname.

erikrose: Done. :-)

PR #466, opened 2015-08-27
Bugs 1183885 and 1191979 regarding offset and limit.

pelmers:
The first commit of this pair was already reviewed in https://github.com/mozilla/dxr/pull/455, but it was dropped from the master branch most likely when I accidentally temporarily merged to path promotion.

To fix 1191979 (the second commit here), I string.replace offset= and limit= by '' from the ajax URL before updating the history state.

### DIFF
--- a/dxr/static/js/dxr.js
+++ b/dxr/static/js/dxr.js
@@ -195,31 +195,8 @@ $(function() {
      * Starts or restarts the scroll position poller.
      */
     function pollScrollPosition() {
+        clearInterval(scrollPoll);
         scrollPoll = setInterval(infiniteScroll, 250);
-    }
-
-    // On document ready start the scroll pos poller.
-    pollScrollPosition();
-
-    /**
-     * Updates the window's history entry to not break the back button with
-     * infinite scroll.
-     * @param {int} offset - The offset to store in the URL
-     */
-    function setHistoryState(offset) {
-        var state = {},
-            re = /offset=\d+/,
-            locationSearch = '';
-
-        if (location.search.indexOf('offset') > -1) {
-            locationSearch = location.search.replace(re, 'offset=' + offset);
-        } else {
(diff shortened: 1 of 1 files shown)


dxr/static/js/dxr.js
```diff
@@ -426,23 +391,30 @@ $(function() {
...
                 data.query = query;
-                // New results, overwrite
+                // New results, display them.
                 if (myRequestNumber > displayedRequestNumber) {
                     displayedRequestNumber = myRequestNumber;
-                    populateResults(data, false);
-                    var pushHistory = function() {
-                        history.pushState({}, '', queryString);
+                    populateResults(data, appendResults);
+                    var pushHistory = function () {
```
erikrose: Can you think of any reason we use this cumbersome function definition syntax rather than just `function pushHistory`?

pelmers: They behave the same in this case, so I kept what we had originally. I shall change to the other syntax.

dxr/static/js/dxr.js
```diff
@@ -426,23 +391,30 @@ $(function() {
...
+                        // Strip off offset= and limit= when updating.
+                        var displayURL = queryString.replace(/&?offset=\d+/, '').replace(/&?limit=\d+/, '');
```
erikrose: The regex should insist on a ? or a & before each param. Otherwise, we'll run into trouble when we inevitably add a foo_offset param later. (Better still might be always building up a query URL from scratch in doQuery, but this is good enough for a quick fix of a user-visible problem.)

dxr/static/js/dxr.js
erikrose: I didn't mean you had to change it as part of this PR; I just wondered if you saw anything I didn't. It's better not to change it (or at least to keep it as a separate commit).

erikrose:
Looking good. :-)

PR #467, opened 2015-08-30
Better makefile

erikrose:
f? I haven't tested the JS yet.

Makefile is now sane. You can run "make" with impunity.

* It will no longer run `npm install` if it doesn't need to, so it will be fast and not hit the network.
* It will rebuild the nunjucks templates if necessary and leave them alone otherwise. You don't need to call `make templates` manually anymore (though it continues to exist so the deploy process can call it and not bother compiling the clang plugin).
* `make clean` now removes the installed node modules.

Also upgrade nunjucks to 2.0 and npm-lockdown to 0.0.6.

Also also remove Grunt, so we only need 5MB of node modules rather than 30MB.


erikrose:
(Confirmed: nunjucks is broken atm. Working on it.)

PR #471, opened 2015-10-13
Fix crashes in Rust plugin.

nrc:
This is a stop gap fix, the real fix is in the compiler and I'm working on it, in the meantime this stops crashes during indexing so at least we'll have an index, even if some links are incomplete.


erikrose:
Oh hooray. I thought it was something I did, and I was just about to have a look. :-)

PR #472, opened 2015-10-21
Display SVGs as XML, and add link to view as image.

pelmers:
Fixes bug 1202598. https://bugzilla.mozilla.org/show_bug.cgi?id=1202598

Also normalize extension to lowercase when looking up in the extension map, so that .svg and .SVG are treated the same.

### DIFF
--- a/dxr/mime.py
+++ b/dxr/mime.py
@@ -4,17 +4,18 @@ from os.path import splitext
 def icon(path):
     """Return the basename (no extension) of the icon file to use for a path."""
     root, ext = splitext(path)
-    return ext_map.get(ext[1:], 'unknown')
+    return ext_map.get(ext[1:].lower(), 'unknown')
 
 
 def is_text(data):
     # Simple stupid test that apparently works rather well :)
     return '\0' not in data
 
+
 def is_image(path):
     """Determine whether the path is an image."""
     _, ext = splitext(path)
-    return ext_map.get(ext[1:], False) == 'image'
+    return ext_map.get(ext[1:].lower(), False) == 'image'
 
 
 # File extension known as this point
@@ -66,6 +67,5 @@ ext_map = {
     "jpg":        'image',
     "jpeg":       'image',
     "png":        'image',
-    "gif":        'image',
-    "svg":        'image'
+    "gif":        'image'
 }
--- a/dxr/mime.py
+++ b/dxr/mime.py
@@ -4,7 +4,7 @@ from os.path import splitext
 def icon(path):
     """Return the basename (no extension) of the icon file to use for a path."""
     root, ext = splitext(path)
-    return ext_map.get(ext[1:].lower(), 'unknown')
+    return ext_map.get(ext[1:], 'unknown')
 
 
 def is_text(data):
@@ -12,10 +12,14 @@ def is_text(data):
     return '\0' not in data
 
 
-def is_image(path):
-    """Determine whether the path is an image."""
-    _, ext = splitext(path)
-    return ext_map.get(ext[1:].lower(), False) == 'image'
+def is_binary_image(path):
+    """Determine whether the path is a binary image."""
+    return icon(path) == 'image'
+
+
+def is_indexable_image(path):
+    """Determine whether the path is an image with indexable (text) contents."""
+    return icon(path) in {'svg'}
 
 
 # File extension known as this point
@@ -32,6 +36,7 @@ ext_map = {
     "c":          'c',
     "xul":        'ui',
     "svg":        'svg',
+    "SVG":        'svg',
     "in":         'build',
     "idl":        'conf',
     "java":       'java',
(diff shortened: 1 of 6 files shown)


dxr/plugins/core.py
```diff
@@ -429,6 +429,11 @@ class FileToIndex(dxr.indexers.FileToIndex):
     def __init__(self, path, contents, plugin_name, tree, vcs):
         super(FileToIndex, self).__init__(path, contents, plugin_name, tree)
         self.vcs = vcs
+        self.extension = splitext(self.path)[1].lower()
```
erikrose: The dropping everything to lowercase doesn't thrill me, because sometimes extensions actually are case-sensitive. The worst of these is .C, which means C++ to gcc.

dxr/plugins/pygmentize.py
```diff
@@ -83,6 +83,7 @@ def _lexer_for_filename(filename):
                                or filename.endswith('.idl')
                                or filename.endswith('.webidl')
                 else 'dummy.xml' if filename.endswith('.xul')
+                                    or filename.endswith('.svg')
```
erikrose: Did you know endswith() can take a tuple of suffixes? Woo!

tests/test_binary_files/test_binary_files.py
```diff
@@ -31,20 +31,13 @@ def test_too_fat(self):
         ok_('href="/code/source/Green%20circle.jpg" class="icon image too_fat"' in response.data)
 
     def test_svg(self):
-        """Make sure we show the svg."""
+        """Make sure we show the source of the svg, but allow way to reach its image."""
```
erikrose: "a way"

dxr/mime.py
```diff
@@ -4,17 +4,18 @@
...
+    return ext_map.get(ext[1:].lower(), 'unknown')
 
 
 def is_text(data):
     # Simple stupid test that apparently works rather well :)
     return '\0' not in data
 
+
 def is_image(path):
     """Determine whether the path is an image."""
```
erikrose: We should probably make the meaning of this more specific, since SVG has shown a use case for said specificity.

dxr/plugins/core.py
erikrose: Maybe we can just get away without it, as https://dxr.mozilla.org/mozilla-central/source/addon-sdk/source/test/fixtures/mofo_logo.SVG is the only uppercase one in moz-central.

erikrose: https://en.wikipedia.org/wiki/Scalable_Vector_Graphics suggests that "SVG" is not an acceptable extension, but maybe there's a more official spec somewhere else.

tests/test_binary_files/test_binary_files.py
```diff
@@ -31,20 +31,13 @@ def test_too_fat(self):
...
         response = self.client().get('/code/source/yellow_circle.svg')
-        ok_('src="/code/raw/yellow_circle.svg"' in response.data)
+        ok_('href="/code/raw/yellow_circle.svg"' in response.data)
         response = self.client().get('/code/raw/yellow_circle.svg')
         ok_(response.status_code, 200)
 
     def test_some_bytes(self):
-        """We want some_bytes to show in the folder index, but without its own page."""
```
erikrose: These tests should both exist, so let's just rename one.

erikrose:
I don't think this needs a format-version bump, right?

dxr/mime.py
erikrose: For example, we might have `is_image()` and `is_image_with_indexable_text()` and just call those from various points below rather than hard-coding 'svg' specifically. Then we make the meaning super-clear at all the callsites and, as a bonus, have a hook for any other image-but-text formats that crop up.

dxr/plugins/core.py
pelmers: Good points, for now I will add the case to the table. If I get the chance I'll file a bug on the usage of SVG instead of svg.

dxr/plugins/pygmentize.py
pelmers: Did not know! \o/

pelmers:
I don't think we need to bump format, since the new is_binary_image is a subset of previous is_image.

tests/test_binary_files/test_binary_files.py
pelmers: Brought it back, renamed both

dxr/mime.py
pelmers: Good points; I've factored the check into separate functions: is_binary_image and is_indexable_image to differentiate the two kinds.

dxr/plugins/core.py
```diff
@@ -436,10 +436,13 @@ def needles(self):
             # realpath will keep following symlinks until it gets to the 'real' thing.
             yield 'link', relpath(realpath(self.absolute_path()), self.tree.source_folder)
         yield 'path', self.path
-        extension = splitext(self.path)[1]
+        extension = splitext(self.path)[1].lower()
```
erikrose: Are we keeping this lower() for a reason?

pelmers: Ah, I missed this line. Updated!

erikrose:
Made just one little naming tweak and merged. Thanks! :-D

PR #475, opened 2015-10-29
Fix bug 1195855, prevent error on some Rust ref: searches.

pelmers:
https://bugzilla.mozilla.org/show_bug.cgi?id=1195855.
It looks like for crates in rustc, the 'name' field of the needle is an integer, which breaks maybe_lower of filter highlighting. For example, consider the query https://dxr.mozilla.org/rust/search?q=ref:std

It also seems that in some cases the 'name' field of the needle is not even present, which seems to be the case for searches such as https://dxr.mozilla.org/rust/search?q=ref:test
However I haven't had any luck constructing a minimal test case for this part of the bug.
@nrc any ideas?

The fix simply makes sure that 'name' exists and is a string.

### DIFF
--- a/dxr/filters.py
+++ b/dxr/filters.py
@@ -196,7 +196,9 @@ class NameFilterBase(Filter):
         """
         maybe_lower = (identity if self._term['case_sensitive'] else
                        unicode.lower)
-        return maybe_lower(entity['name']) == maybe_lower(self._term['arg'])
+        return ('name' in entity
+                and isinstance(entity['name'], unicode)
+                and maybe_lower(entity['name']) == maybe_lower(self._term['arg']))
 
     def highlight_content(self, result):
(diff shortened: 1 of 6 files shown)


nrc:
@pelmers I think you get an id for a name (or no name at all) for `impl`s, e.g., `impl Foo for Bar { ... }`. Where `Foo` is a trait and `Bar` some types. You might also try an impl where `Bar` is generic, e.g., `impl<T> Foo for T { ... }`. You'll need to experiment a bit with tests, I'm not sure exactly what is going on here.

erikrose:
I lean toward making the rust plugin return the right type rather than armoring the highlighter against bad data. Incidentally, nrc just landed a RustSingleFileTestCase that may or may not be helpful in this case.

PR #476, opened 2015-10-31
Fix problems with Python encoding declarations and column offsets

jbradberry:
Due to https://bugs.python.org/issue22221, attempts to index Python files that contain encoding declarations at the top (e.g. `# -*- coding: utf-8 -*-`) will raise a SyntaxError, causing those files to be skipped.

Additionally, the ast.parse function will yield nodes with column offsets calculated in utf-8 bytes, not unicode characters.


erikrose:
Otherwise, lgtm!

jbradberry:
Pre-emptive rebase on top of #487.

All of the review comments should be addressed now.

jbradberry:
@erikrose r?

PR #478, opened 2015-11-03
Get Escape key handling working again.

kleintom:
Re-fix Bug 1153726 by updating the event key value for Escape for
recent versions of gecko: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key


erikrose:
Thanks!

PR #480, opened 2015-11-03
Adjust pop-up event handling.

kleintom:
This changes three things:
1) A right or middle click on the tree selector (or filter) menu
   no longer dismisses the menu;
2) As a side effect, Bug 965666 is fixed (opening a menu or help or
   the context menu closes any other open pop-up).
3) The Help menu now closes on Escape.


erikrose:
Thanks for the fixes (and the better JS style)! :-D I also like the modularity—how every menu is now responsible for its own hiding—rather than lumping all the responsibility into dxr.js.

PR #481, opened 2015-11-03
Keep results from blanking on scroll. Fixes bug #1212956.

erikrose:
The query run by infinite scroll is built from, among other things, the state of the Case-sensitive checkbox. Make sure that's synced up with the case param of the querystring on load.


erikrose:
Not bothering to clean up the general DRY ickiness around the case-sensitivity state, since we're ripping it out soon.

erikrose:
Whee, fixes bug #1072546 as well.

PR #482, opened 2015-11-04
Remove case sensitivity checkbox.

pelmers:
Remark: the previous PR, #460 targeted the 'es' branch, but that has since merged into 'master.' So I have to open a new PR. The contents should be the same.

If any text is uppercase, then assume case sensitive. Otherwise, assume it’s insensitive. Use an @ sigil in front of a text term to force case sensitive. This also gives us per-term case sensitivity.

Notes from 11/3:
- Create a 'case' operator modifier, and consider turning '+' and '-' into operators too (qual, not).
My initial idea is to expand the grammar to differentiate between "operators" (var, ref, id, etc.) which affect the *filter query* and "modifiers" (not, qual, case) which affect the *search term* itself.

### DIFF
--- a/dxr/app.py
+++ b/dxr/app.py
@@ -135,21 +135,19 @@ def search(tree):
     query_text = req.get('q', '')
     offset = non_negative_int(req.get('offset'), 0)
     limit = min(non_negative_int(req.get('limit'), 100), 1000)
-    is_case_sensitive = req.get('case') == 'true'
 
     # Make a Query:
     query = Query(partial(current_app.es.search,
                           index=frozen['es_alias']),
                   query_text,
-                  plugins_named(frozen['enabled_plugins']),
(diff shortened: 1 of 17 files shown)


pelmers:
I move 'not', 'case', and 'qual' into a new category of "Modifier," which can change the term dictionary constructed by the parse tree. Modifier provides some default which is merged with all others, and a 'modify' operation which triggers when we see it on the query term.
We refer to modifier in the query either by a "name," just like operators, or a shorthand, which is what we currently have: -,+,@ etc.

thoughts? @erikrose

pelmers:
Side effect: can no longer parse input such as "----+++" which consists of only modifers, since now
```
term = shorthand* (modifier ":")* positive_term
```

pelmers:
Superseded by #539

PR #483, opened 2015-11-04
Fix a bug which causes field and variable references to be self-links

nrc:
Can we test this? It only manifests when you click 'jump to definition' on a variable's menu.

### DIFF
--- a/dxr/plugins/rust/tests/__init__.py
+++ b/dxr/plugins/rust/tests/__init__.py
@@ -1,0 +1,14 @@
+from dxr.testing import SingleFileTestCase
+
+class RustSingleFileTestCase(SingleFileTestCase):
+    """Test case suited to testing the Rust plugin."""
+    source_filename = 'mod.rs'
+
+    @classmethod
+    def config_input(cls, config_dir_path):
+        config = super(RustSingleFileTestCase, cls).config_input(config_dir_path)
+
+        config['DXR']['enabled_plugins'] = 'pygmentize rust'
+        config['code']['build_command'] = '$RUSTC mod.rs'
+
+        return config
--- a/dxr/plugins/rust/tests/test_var_menus.py
+++ b/dxr/plugins/rust/tests/test_var_menus.py
@@ -1,0 +1,17 @@
+from dxr.plugins.rust.tests import RustSingleFileTestCase
+from dxr.testing import menu_on
+
+class VarMenuTests(RustSingleFileTestCase):
+    source = """
+    fn main() {
+        let _ = FOO;
+    }
+    const FOO: i32 = 42;
+    """
+
+    def test_const_link(self):
+        """ test that the link to defintion works properly """
+        menu_on(self.source_page('mod.rs'),
+                'FOO',
+                {'html': 'Jump to definition',
+                         'href': '/code/source/mod.rs#5'})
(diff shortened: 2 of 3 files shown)


erikrose:
Yep! See test_menus.py. Basically, call menu_on() in your test, and say what you expect to see.

nrc:
now with a test \o/

dxr/plugins/rust/tests/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+class RustSingleFileTestCase(SingleFileTestCase):
+    """Test case suited to testing the Rust plugin."""
+    source_filename = 'mod.rs'
+
+    @classmethod
+    def config_input(cls, config_dir_path):
+        config = super(RustSingleFileTestCase, cls).config_input(config_dir_path)
+
+        config['DXR']['enabled_plugins'] = 'pygmentize rust'
+        config['DXR']['disabled_plugins'] = 'clang'
```
erikrose: Shouldn't need this; `enabled_plugins` is comprehensive. `disabled` is useful only when you're doing an "all but these" config with `enabled` equal to `*`.

dxr/plugins/rust/tests/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+        config['code']['build_command'] = '$RUSTC mod.rs'
```
erikrose: On the Vagrant box, $RUSTC is empty, so we'll need to rig a way to fall back to "rustc" like the indexing plugin does.

dxr/plugins/rust/tests/test_var_menus.py
```diff
@@ -0,0 +1,17 @@
...
+    fn main() {
+        let _ = FOO;
+    }
+    const FOO: i32 = 42;
+    """
+
+    def test_const_link(self):
+        """ test that the link to defintion works properly """
+        menu_on(self.source_page('mod.rs'),
+        'FOO',
```
erikrose: Let's indent the rest of these lines under `self`.

dxr/plugins/rust/tests/__init__.py
erikrose: If you can't think of anything (and nothing comes to my mind), we can just rig it so $RUSTC *is* defined on the vagrant box. There's precedent for appending stuff to .bashrc in vagrant_provision.sh.

erikrose: Wait! Never mind; I see the rust plugin exports $RUSTC itself. Happy day!

nrc:
With comments addressed

erikrose:
lgtm! If tests pass, I'll merge.

erikrose:
@klibby Can you cause the indexer Docker image to rebuild so this gets in? Thanks!

klibby:
a7daaaa changes what 'make' does, so the docker build fails:

```bash
env CC=clang CXX=clang++ make -C dxr
make: Entering directory `/builds/dxr-build-env/dxr'
$VIRTUAL_ENV/bin/python peep.py install -r requirements.txt
/bin/sh: 1: /bin/python: not found
make: *** [.peep_installed] Error 127
```

Requiring a venv before running make seems to conflict with the current documentation, but is an easy fix for me: https://github.com/klibby/dxr-docker/commit/0f301cf524edd673d5fd804004622355ae073657

New docker image built and pushed so new builds will pick it up!

erikrose:
Ugh, sorry about that. Looks like another reason to merge the repos. I'm having a little look today at moving to Docker for the dev env, which is a first step toward that.

Thanks for absorbing the shock and fixing it!

PR #487, opened 2015-11-07
Change the Python callers needle to point to the call site

jbradberry:
This is desirable for consistency with the clang plugin.

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1221137


jbradberry:
This should be it, unless I go nuts adding more tests.  @erikrose r?

PR #488, opened 2015-11-09
Simple docker dev environment

pelmers:
Add dockerfiles for dxr and elasticsearch development, and makefile targets.

Taking a different approach from #415, I build upon fubar's dxr-docker image to make a dev image by replacing the Github DXR pull with the local copy.

Usage:
Prereq: Have docker running.
1. Build elasticsearch image and run it. `make es`
2. Build DXR image. `make image`
3. Run tests/index some code. `make tests` or `make index clean serve`
Remark: can specify tests via `make tests TESTS=tests/test_<...>.py`

TODO:
update docs.
maybe use docker-compose to setup the two docker images at once.

Some questions:
Would it be better to bring dxr-docker in first?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM quay.io/fubar/ubuntu-indexer
+MAINTAINER Peter
+
+# Install Rust.
+RUN curl -s https://static.rust-lang.org/rustup.sh | sh -s -- --channel=nightly --date=2015-06-14 --yes
+# It's okay Mercurial, you can trust staff group.
+RUN echo [trusted] >> /home/jenkins/.hgrc && echo groups = staff >> /home/jenkins/.hgrc
+# Replace the distributed dxr with the one from here (working copy).
+ADD . /builds/dxr-build-env/dxr
+# Run make.
+RUN . venv/bin/activate && \
+    dxr/peep.py install -r dxr/requirements.txt && \
+    cd dxr && \
+    python setup.py install && \
+    make && \
+    cd - && \
+    deactivate
+RUN pip install nose-progressive
+
+WORKDIR /builds/dxr-build-env/dxr
--- a/docker/README
+++ b/docker/README
@@ -4,16 +4,11 @@ So if you run `dxr index`, that happens here.
 Usage of the docker environment:
 Put a dxr.config and source code directory here.
 
-Run an elasticsearch docker container, and expose port 8000.
-docker run --name es -p 80:8000 elasticsearch
-
-Remark: By default the elasticsearch container does not support javascript
-queries, which DXR uses -- regexp searches will fail unless you rebuild the
-elasticsearch container with an updated config file.
-
 In the parent (repository root):
-docker build -t dxr .
-docker run --net container:es dxr index
-docker run --net container:es dxr serve -a
+
+Run an elasticsearch docker container, and expose port 8000.
+make es
+make image
+make index clean serve
 
 Visit $DOCKER_IP in the browser.
--- a/docker/README
+++ b/docker/README
@@ -1,0 +1,19 @@
+This directory is copied up to the container, and is set as the WORKDIR.
+So if you run `dxr index`, that happens here.
+
+Usage of the docker environment:
+Put a dxr.config and source code directory here.
+
+Run an elasticsearch docker container, and expose port 8000.
+docker run --name es -p 80:8000 elasticsearch
+
+Remark: By default the elasticsearch container does not support javascript
+queries, which DXR uses -- regexp searches will fail unless you rebuild the
+elasticsearch container with an updated config file.
+
+In the parent (repository root):
+docker build -t dxr .
+docker run --net container:es dxr index
+docker run --net container:es dxr serve -a
+
+Visit $DOCKER_IP in the browser.
--- a/makefile
+++ b/makefile
@@ -2,8 +2,20 @@
 
 all: requirements static plugins
 
-test: all
-	python setup.py test
+es:
+	cd docker_es; docker build -t es_dxr .; docker rm -f es_dxr || true; docker run --name es_dxr -p 80:8000 es_dxr
+
+image:
+	docker build -t dxr .
+
+test:
+	docker run --net container:es_dxr --tty --name dxr_test dxr nosetests --with-progressive ${TESTS}
+
+index:
+	docker run --net container:es_dxr --tty --name dxr_index -w "/builds/dxr-build-env/dxr/docker" dxr dxr index
+
+serve:
+	docker run --net container:es_dxr --tty --name dxr_serve -w "/builds/dxr-build-env/dxr/docker" dxr dxr serve -a
 
 clean: static_clean
 	rm -rf node_modules/.bin/nunjucks-precompile \
@@ -11,6 +23,7 @@ clean: static_clean
 	       .npm_installed \
 	       .peep_installed
 	find . -name "*.pyc" -exec rm -f {} \;
+	docker rm -f dxr_index dxr_serve dxr_test || true
 	$(MAKE) -C dxr/plugins/clang clean
 
 static_clean:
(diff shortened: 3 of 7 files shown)


erikrose:
And just for my reference, here's how I got docker going on OS X:
```
brew update
brew install xhyve
git clone https://github.com/ailispaw/boot2docker-xhyve
cd boot2docker-xhyve
make
sudo ./xhyverun.sh
# boot2docker login: docker
```
We can add that as a readme or a handy script eventually.

makefile
```diff
@@ -2,8 +2,20 @@
 
 all: requirements static plugins
 
-test: all
-	python setup.py test
+es:
+	cd docker_es; docker build -t es_dxr .; docker rm -f es_dxr || true; docker run --name es_dxr -p 80:8000 es_dxr
```
erikrose: I assume these are all on one line because docker twiddles some env vars? If we do have to keep them on one line, we should at least separate the commands with `&&` so they don't plow onward if something breaks.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM quay.io/fubar/ubuntu-indexer
```
erikrose: I think I'd like to flip this upside-down. Since the dev env is a subset of what's needed on the indexers, it seems to me that the indexer should pull from this, rather than the other way around. We should move things like [virtualenv setup](https://github.com/klibby/dxr-docker/blob/master/docker/ubuntu-indexer/bin/dxr-setup.sh#L35-L36) over to this, and the Jenkins-specific stuff can stay over on the indexer image. How does that strike you?

makefile
pelmers: Yes, also it strikes me that we would find it more maintainable to put the line into a makefile under docker_es and call that. 
On Nov 9, 2015 2:24 PM, Erik Rose <notifications@github.com> wrote:In makefile:
> @@ -2,8 +2,20 @@
>  
>  all: requirements static plugins
>  
> -test: all
> -	python setup.py test
> +es:
> +	cd docker_es; docker build -t es_dxr .; docker rm -f es_dxr || true; docker run --name es_dxr -p 80:8000 es_dxr

I assume these are all on one line because docker twiddles some env vars? If we do have to keep them on one line, we should at least separate the commands with && so they don't plow onward if something breaks.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Install Rust.
+RUN curl -s https://static.rust-lang.org/rustup.sh | sh -s -- --channel=nightly --date=2015-06-14 --yes
+# It's okay Mercurial, you can trust staff group.
+RUN echo [trusted] >> /home/jenkins/.hgrc && echo groups = staff >> /home/jenkins/.hgrc
+# Replace the distributed dxr with the one from here (working copy).
+ADD . /builds/dxr-build-env/dxr
+# Run make.
+RUN . venv/bin/activate && \
+    dxr/peep.py install -r dxr/requirements.txt && \
```
erikrose: There's a make target for this now: `make requirements`.

makefile
```diff
@@ -11,6 +23,7 @@ clean: static_clean
 	       .npm_installed \
 	       .peep_installed
 	find . -name "*.pyc" -exec rm -f {} \;
+	docker rm -f dxr_index dxr_serve dxr_test || true
```
erikrose: Under what conditions does the error-eating `true` need to kick in?

makefile
pelmers: Ick, not replying to email comment from phone again.

Dockerfile
pelmers: Ah, perfect!

makefile
pelmers: If any of the containers are not present, then it goes to `true` branch.

I've been using `make clean` after running tests, or after making an index-serve of code in docker/ and want to clear out all of my containers because I have changed DXR locally and need to rebuild the image anyway.

Dockerfile
pelmers: sounds right to me -- we should keep the DXR building script in the DXR repo and the deploy specifics over in the deploy repo. @klibby opinion?

erikrose: Right. In the design taking shape in my mind, this repo would get DXR to the point where it's able to index code of all languages it supports. (So, for example, rust would be on the image.) Ideally, we could add the dev-specific stuff (Sphinx, nose, …) as a separate layer or make target or env var switch (whatever makes sense) so it's suitable both for dev and to base deployment images on.

docker/README
```diff
@@ -0,0 +1,14 @@
+This directory is copied up to the container, and is set as the WORKDIR.
```
erikrose: As per IRC, maybe we can use a [docker volume](https://docs.docker.com/engine/userguide/dockervolumes/) instead to save copying. Source trees can get really big. If we use `-v` rather than a hard-coded path in the Dockerfile, we can even serve prod's needs, giving us a decoupled way to provide source to the container after the host-dwelling Jenkins updates it.

klibby:
I'm happy to have the docker image building bits live in the dxr repo if that works best for you. I'm really only just adding some stuff on top of it, including the shell script called to do the indexing.

The indexers mount a host directory as a volume. I recall that there was something about using a data volume (ie docker volume, not a mounted host directory) that didn't play nicely at all with thing, but I am unable to remember the specifics at the moment. Perhaps it was just the sheer size and number of repos that I have to contend with, which is unlikely to be the case for a dev system.

erikrose:
@klibby Cool. We're going to have to build an image anyway in this repo, so we might as well make it generically useful. If you're not using docker data volumes, how are you doing the mount? NFS?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    cd dxr && \
+    python setup.py install && \
+    make && \
+    cd - && \
+    deactivate
```
erikrose: What's this for?

erikrose:
I'm also having a look at https://www.vagrantup.com/blog/feature-preview-vagrant-1-6-docker-dev-environments.html as a possible friendly, cross-platform frontend to the Docker-based core. Then we wouldn't have to explain docker-machine or how to set up a chain of shared-folder mounts through a boot2docker host VM.

pelmers:
That idea sounds good to me!

Dockerfile
pelmers: Hmm, I borrowed this from https://github.com/klibby/dxr-docker/blob/00f7b4fdf86420d9029fdac445c67fb7f98d6ecc/docker/ubuntu-indexer/bin/dxr-setup.sh#L40-L43

pelmers:
Also, please do pick at this patch; I realize that I've several open and languishing pull requests and can't do much about them at the moment thanks to the backloaded courses I take.

klibby:
@erikrose it's an NFS mount that's mounted as a docker volume (e.g. docker run -v /mnt/nfs_data:/local/src).

Also, releng cleaned up their old docker images and it turns out I've been behind the times. Looks like m-c based builds are now using centos as a base, but I'm running into weird build problems trying to use it. Currently testing a mostly bare (ie without releng/taskcluster stuff) ubuntu image. It would be nice to use TC images with an eye to shifting the builds to them, but I just don't have the cycles to make it go atm.

erikrose:
@klibby It sounds like factoring most of the DXR setup stuff out into a dxr_setup.sh script (much like you've done) or the makefile is a good move: we can reuse it on Ubuntu, CentOS, or whatever the next flavor of the month is. (We can't necessarily let the moz-central image's OS choice drive DXR's, since there will probably be various releng images with various OS bases.) I lean toward putting even OS-specific setup steps into shell scripts (ubuntu_setup.sh, etc.) for reuse, since Docker doesn't support multiple inheritance, and we may want to base our indexer images on releng's rather than on DXR's. To some degree, I'm coming to the same conclusions you did in https://github.com/klibby/dxr-docker/tree/master/docker/ubuntu-indexer/bin. Does this sound sane to you?

PR #489, opened 2015-11-11
Get clang plugin overrides and overriddens menu entries working again.

kleintom:
Partially fixes bug 1191562.  The remaining cases where the
overrides and overriddens menu options still don't appear are for a
pure virtual function and its override.

### DIFF
--- a/dxr/plugins/clang/condense.py
+++ b/dxr/plugins/clang/condense.py
@@ -57,6 +57,14 @@ def process_function(props):
     props['type'] = c_type_sig(input_args, props['type'])
     return props
 
+def process_maybe_override(overrides, overriddens, props):
+    qualname = props.get('qualname')
+    if qualname in overrides or qualname in overriddens:
+        # Mark this function as virtual.
+        props['override'] = True
+
+    return props
+
 
 def process_override(overrides, overriddens, props):
     """Note overrides of methods, and organize them so we can emit
@@ -233,9 +241,7 @@ def lines_from_csvs(folder, file_glob):
     return chain.from_iterable(lines_from_csv(p) for p in paths)
 
 
-DISPATCH_TABLE = {'call': process_call,
-                  'function': process_function}
-def condense_file(csv_folder, file_path):
+def condense_file(csv_folder, file_path, overrides, overriddens):
     """Return a dict representing an analysis of one source file.
 
     This is phase 2: the file-at-a-time phase.
@@ -248,11 +254,33 @@ def condense_file(csv_folder, file_path):
         compiler plugin
     :arg file_path: A path to the file to analyze, relative to the tree's
         source folder
+    :arg overrides: A dict whose keys are function qualnames that are overrides
+    :arg overriddens: A dict whose keys are function qualnames that are
+        overriddens
 
     """
+    def process_function_for_override(overrides, overriddens, props):
+        override_props = process_maybe_override(overrides, overriddens, props)
+        return process_function(override_props)
+
+    def process_maybe_function(overrides, overriddens, props):
+        if props.get('kind') == 'function':
+            return process_maybe_override(overrides, overriddens, props)
+        else:
+            return props
+
+    process_maybe_function_for_override = partial(process_maybe_function,
+                                                  overrides, overriddens)
+
+    dispatch_table = {'call': process_call,
+                      'function': partial(process_function_for_override,
+                                          overrides, overriddens),
+                      'ref': process_maybe_function_for_override,
+                      'decldef': process_maybe_function_for_override}
+
     return condense(lines_from_csvs(csv_folder,
                                     '{0}.*.csv'.format(sha1(file_path).hexdigest())),
-                    DISPATCH_TABLE)
+                    dispatch_table)
 
 
 def condense_global(csv_folder):
--- a/dxr/testing.py
+++ b/dxr/testing.py
@@ -391,37 +391,50 @@ def _make_file(path, filename, contents):
         file.write(contents.encode('utf-8'))
 
 
-def _decoded_menu_on(haystack, text):
-    """Return the JSON-decoded menu found around the source code ``text`` in
-    the HTML ``haystack``.
+def _decoded_menu_on(haystack, text, text_instance=1):
+    """Return the JSON-decoded menu found around the ``text_instance``th source
+    code occurrence of ``text`` in the HTML ``haystack``.
 
     Raise an AssertionError if there is no menu there.
 
     """
     # We just use cheap-and-cheesy regexes for now, to avoid pulling in and
     # compiling the entirety of lxml to run pyquery.
-    match = re.search(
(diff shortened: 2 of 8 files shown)


kleintom:
It would actually be quite easy at this point to add a "Find overrides" or a "Find overridden" menu item -only- when overrides respectively overriddens exists (right now we just always add both if the function is virtual).  Then people could just look at the menu to see if overrides or overriddens existed instead of having to run the search.  Would you want to do that instead?  (Note also that we don't display any override/overridden menu options for a function marked virtual but which has no overrides, so the suggested behavior would be more consistent with the way that currently works.)

kleintom:
Also, it seems (at least so far) like the pure virtual issue resolves itself if the if check at
https://github.com/mozilla/dxr/blob/master/dxr/plugins/clang/dxr-index.cpp#L494
is removed (just assign overriddenDef = *iter instead).  I'm not familiar enough to know if that would break other things though - any thoughts on that?  FWIW all the current tests and the tests added in this patch pass with that change.

dxr/testing.py
```diff
@@ -391,37 +391,50 @@ def _make_file(path, filename, contents):
...
     Raise an AssertionError if there is no menu there.
 
     """
     # We just use cheap-and-cheesy regexes for now, to avoid pulling in and
     # compiling the entirety of lxml to run pyquery.
-    match = re.search(
-            '<a data-menu="([^"]+)"[^>]*>' + re.escape(cgi.escape(text)) + '</a>',
-            haystack)
+    pattern = '<a data-menu="([^"]+)"[^>]*>' + re.escape(cgi.escape(text)) + '</a>'
+    if text_instance == 1:
```
erikrose: Let's not even special-case this; I'd rather have fewer branches to go wrong.

dxr/testing.py
```diff
@@ -391,37 +391,50 @@ def _make_file(path, filename, contents):
...
+        match = re.search(pattern, haystack)
+    else:
+        matches = re.finditer(pattern, haystack)
+        for _ in range(text_instance):
```
erikrose: xrange

dxr/plugins/clang/condense.py
```diff
@@ -248,11 +254,33 @@ def condense_file(csv_folder, file_path):
         compiler plugin
     :arg file_path: A path to the file to analyze, relative to the tree's
         source folder
+    :arg overrides: A dict whose keys are function qualnames that are overrides
+    :arg overriddens: A dict whose keys are function qualnames that are
+        overriddens
 
     """
+    def process_function_for_override(overrides, overriddens, props):
```
erikrose: Is there any reason for these to be inner functions? Most of the rest aren't. I see 2 alternatives:
1. Make `dispatch_table` local, as you've done, and use partials.
2. Keep `dispatch_table` global, use inner functions as you have, but rely on closing over `overrides` and `overriddens` rather than passing them in.
I don't think there's a need to both make `dispatch_table` *and* the functions local.

erikrose: Is there any reason for these to be inner functions? Most of the rest aren't, and I loathe inconsistency. :-) I see 2 alternatives:
 1. Make the functions module-level, and use partials.
 2. Keep the functions local, but rely on closing over `overrides` and `overriddens` rather than passing them in, so we have an excuse to keep them local.

In any case, `dispatch_table` would remain local, as you've made it.

I don't think there's a need to both make `dispatch_table` *and* the functions local.

erikrose: Sorry for all the edits; I wasn't thinking straight. :-)

dxr/plugins/clang/condense.py
```diff
@@ -57,6 +57,14 @@ def process_function(props):
     props['type'] = c_type_sig(input_args, props['type'])
     return props
 
+def process_maybe_override(overrides, overriddens, props):
+    qualname = props.get('qualname')
+    if qualname in overrides or qualname in overriddens:
+        # Mark this function as virtual.
```
erikrose: I don't think a function's presence in `overrides` or `overriddens` implies that it's virtual. After all, you can override a nonvirtual function; dispatch is just not as fancy.

erikrose:
> It would actually be quite easy at this point to add a "Find overrides" or a "Find overridden" menu item -only- when overrides respectively overriddens exists

Yes please! :-D

erikrose:
> Also, it seems (at least so far) like the pure virtual issue resolves itself if the if check at
https://github.com/mozilla/dxr/blob/master/dxr/plugins/clang/dxr-index.cpp#L494
is removed (just assign overriddenDef = *iter instead). I'm not familiar enough to know if that would break other things though - any thoughts on that? FWIW all the current tests and the tests added in this patch pass with that change.

That's tempting. I'm not sure either, without going back and reading through the clang API docs (and maybe even with). I'd want to try it and index something huge, like mozilla-central, and see if any crashes surface.

erikrose:
Okay, finally got through all this; thanks for the patch! Aside from the little style nits I picked out, I'm mostly interested in understanding why you seem to be entirely concerned with virtual functions. I don't spend all day in C++, but it seems to me one can override nonvirtual functions just as well. It could well be that your patch handles this just fine, but the comments and tests stick entirely to virtual ones. Can you enlighten me?

kleintom:
First, thanks for the review!  (And let me know if you'd prefer a squashed commit here.)
> I'm mostly interested in understanding why you seem to be entirely concerned with virtual functions. I don't spend all day in C++, but it seems to me one can override nonvirtual functions just as well. It could well be that your patch handles this just fine, but the comments and tests stick entirely to virtual ones. Can you enlighten me?

Mostly I was only thinking of virtual overrides out of habit - I used to use those menu options on a regular basis on the m-c code and it was always for virtual overrides, so I got used to thinking that's what they were returning.  Also, the dxr code if-check for whether or not to display those menu items used to be (prior to this patch) a check on "is_virtual", and none of the tests in clang/tests/test_overrides.py test nonvirtual overrides :-)

(Also nonvirtual overrides seem to be somewhat discouraged in C++ (though I don't take a stand on that), for example:
https://www.cs.rit.edu/~mjh/docs/c++-faq/strange-inheritance.html#faq-23.5)

I just ran a test using this patch with a nonvirtual override and no override menu options appear - I didn't change what gets put in FileToIndex.overrides/overridens, so I wonder if the menus have ever appeared for nonvirtual overrides?  In any case I can look into it more and create a bug (or a github issue, whichever you prefer) to track that once this gets merged.

kleintom:
> That's tempting. I'm not sure either, without going back and reading through the clang API docs (and maybe even with). I'd want to try it and index something huge, like mozilla-central, and see if any crashes surface.

Sounds good; I think a better/safer approach might be to keep the current if check I suggested removing and instead add an extra else leg for the pure virtual case - I'll see if I can get mozilla-central to index and file a followup if it looks like that might work.

PR #491, opened 2015-11-17
Deal with Python node location lookups that fail to exist in the lookup tables

jbradberry:
Currently, this should only be method calls.  This happens because the
ast node for calls point to the first token in the chain, instead of
the method token.


jbradberry:
@erikrose r?

erikrose:
Thanks, Jeff. You're a superhero. :-)

PR #492, opened 2015-11-18
Catch UnicodeDecodeError when doing the initial analysis on Python files

jbradberry:
If a file is encoded in some other encoding and has a sequence of
bytes that prevents it from being interpreted as UTF-8, it'll get
slurped up as a bytestring instead of a unicode string.  And in that
case, it'll probably hit the ast_parse utility function and throw this
error.

### DIFF
--- a/dxr/plugins/python/analysis.py
+++ b/dxr/plugins/python/analysis.py
@@ -52,7 +52,7 @@ class TreeAnalysis(object):
         """
         try:
             syntax_tree = ast_parse(file_contents(path, encoding))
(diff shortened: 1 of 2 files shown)


jbradberry:
@erikrose swoop.

jbradberry:
This second commit covers something I noticed when I tried to index a real codebase after fixing the thing in the first commit.  It's just some cleanup of a couple of node locations that failed to get covered by the start is not None check.

erikrose:
lgtm. I'll let the tests pass and then merge it.

PR #493, opened 2015-11-19
Doc updates

kleintom:


kleintom:
Erik, it looks like the debian version of the /etc/sysconfig/elasticsearch config file that I mentioned getting on Fedora is /etc/default/elasticsearch, and that some of the config stuff can be done there instead of other places:
https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-service.html

Regarding the js plugin for elasticsearch, it looks like the new way to install that (without having to worry about figuring out which plugin version goes with your es version) is to do 
```sudo bin/plugin install lang-javascript```
but that fails for me (it looks for a download in the wrong place), so I didn't mention it.

Finally, I'm perfectly happy for you to change my wording or structuring, however you'd like to do that :-)

Thanks!

File: docs/source/deployment.rst [hunk 1]
```diff
@@ -79,25 +110,37 @@ but these pointers should start you off with reasonable performance:
...
-  because then the JVM can no longer use compressed pointers. Giving it one
-  big chunk of RAM up front will avoid heap fragmentation and costly
-  reallocations. The remaining memory will easily be filled by the OS's file
-  cache as it tussles with Lucene indices.
-* Set ``bootstrap.mlockall`` to ``true``. You don't want any swapping.
+* Configure the following in :file:`/etc/elasticsearch/elasticsearch.yml`:
+
+  * Set ``bootstrap.mlockall`` to ``true``. You don't want any swapping.
+  * Set ``script.disable_dynamic`` to ``false``.  This enables DXR's use of the
+    javascript plugin.
```
erikrose: javascript → JavaScript throughout

erikrose:
This looks great, Tom! Thank so much!

PR #495, opened 2015-11-28
Some changes to the Rust plugin

nrc:
These are needed to build Cargo projects such as Rustfmt and Servo. There are a couple of fixes outside the plugin too - I think these are just minor bug fixes.

### DIFF
--- a/dxr/app.py
+++ b/dxr/app.py
@@ -338,7 +338,7 @@ def _browse_folder(tree, path, config):
              f.get('size'),
              url_for('.browse', tree=tree, path=f.get('link', f['path'])[0]),
              f.get('is_binary', [False])[0])
-            for f in files_and_folders if 'path' in f])
+            for f in files_and_folders])
 
 
 def skim_file(skimmers, num_lines):
--- a/dxr/app.py
+++ b/dxr/app.py
@@ -338,7 +338,7 @@ def _browse_folder(tree, path, config):
              f.get('size'),
              url_for('.browse', tree=tree, path=f.get('link', f['path'])[0]),
              f.get('is_binary', [False])[0])
-            for f in files_and_folders])
+            for f in files_and_folders if 'path' in f])
 
 
 def skim_file(skimmers, num_lines):
--- a/dxr/plugins/rust/tests/__init__.py
+++ b/dxr/plugins/rust/tests/__init__.py
@@ -1,10 +1,12 @@
 from dxr.testing import SingleFileTestCase, DxrInstanceTestCase
 import os
 
+RUSTC_COMMAND = "rustc -Zsave-analysis"
+
 class RustDxrInstanceTestCase(DxrInstanceTestCase):
     @classmethod
     def setup_class(cls):
-        os.environ["RUSTC"] = "rustc -Zsave-analysis"
+        os.environ["RUSTC"] = RUSTC_COMMAND
         super(RustDxrInstanceTestCase, cls).setup_class()
 
 
@@ -13,12 +15,15 @@ class RustSingleFileTestCase(SingleFileTestCase):
     source_filename = 'mod.rs'
 
     @classmethod
+    def setup_class(cls):
+        os.environ["RUSTC"] = RUSTC_COMMAND
+        super(RustSingleFileTestCase, cls).setup_class()
+
+    @classmethod
     def config_input(cls, config_dir_path):
         config = super(RustSingleFileTestCase, cls).config_input(config_dir_path)
-        
+
         config['DXR']['enabled_plugins'] = 'pygmentize rust'
         config['code']['build_command'] = '$RUSTC mod.rs'
-        
-        os.environ["RUSTC"] = "rustc -Zsave-analysis"
 
         return config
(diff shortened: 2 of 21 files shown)


nrc:
The big change that affects users (though not the DXR index of Rust itself) is that you need to supply the `RUSTC` env var with the `-Zsave-analysis` flag, the plugin won't do it for you. This is because if we are using Cargo rather than rustc, the extra flag breaks things - note the hack via a script file in the test_cargo test.

nrc:
cc @erikrose

nrc:
ping @erikrose r?

dxr/plugins/rust/tests/test_data/test_fields.py
```diff
@@ -52,29 +52,29 @@ def test_qual_enum_field_ref(self):
         self.found_line_eq('+var-ref:test::EnumFields::Struct::field1', "EnumFields::Struct{ <b>field1</b>, field2 } =&gt; {}", 42)
         self.found_line_eq('+var-ref:test::EnumFields::Struct::field2', "EnumFields::Struct{ field1, <b>field2</b> } =&gt; {}", 42)
 
-    def test_field_ref(self):
```
erikrose: So, did we break some functionality, or what? If it does turn out to be a good idea to temporarily disable this test, we should `raise SkipTest('The reason for skipping them')` as the first line of the test instead of commenting them out. That way, we have the reason recorded, and we have the guilt of a high "skipped" count at the end of the test run to motivate us.

dxr/app.py
```diff
@@ -338,7 +338,7 @@ def _browse_folder(tree, path, config):
              f.get('size'),
              url_for('.browse', tree=tree, path=f.get('link', f['path'])[0]),
              f.get('is_binary', [False])[0])
-            for f in files_and_folders])
+            for f in files_and_folders if 'path' in f])
```
erikrose: When were you running into this problem? I don't see how the FileToIndex in core.py would avoid its `yield 'path', self.path` line.

dxr/plugins/rust/tests/__init__.py
```diff
@@ -8,8 +15,10 @@ class RustSingleFileTestCase(SingleFileTestCase):
     @classmethod
     def config_input(cls, config_dir_path):
         config = super(RustSingleFileTestCase, cls).config_input(config_dir_path)
-
+
```
erikrose: Extra whitespace. (I'm going to clean up the whole codebase in the near future and put a little tooling in place to keep it that way—though not enough to be annoying.)

dxr/plugins/rust/tests/test_cargo/dxr.config
```diff
@@ -0,0 +1,11 @@
+[DXR]
+enabled_plugins     = pygmentize rust
+disabled_plugins    = clang
```
erikrose: No need for this; it's off already since you explicitly listed what you want turned on.

dxr/plugins/rust/tests/test_cargo/test_cargo.py
```diff
@@ -0,0 +1,17 @@
+from dxr.testing import DxrInstanceTestCase
+import os
+from os.path import dirname, join
+import subprocess
+import sys
+
+class CargoTests(DxrInstanceTestCase):
+    @classmethod
+    def setup_class(cls):
+        os.environ["RUSTC"] = "./dxr-rustc.sh"
```
erikrose: (Per IRC) we can do this for now, but it would be great if we could get a RUSTC_ARGS into rustc or cargo or a RUST_SAVE_ANALYSIS into rustc at some point.

dxr/plugins/rust/tests/__init__.py
```diff
@@ -8,8 +15,10 @@ class RustSingleFileTestCase(SingleFileTestCase):
     @classmethod
     def config_input(cls, config_dir_path):
         config = super(RustSingleFileTestCase, cls).config_input(config_dir_path)
-
+        
         config['DXR']['enabled_plugins'] = 'pygmentize rust'
         config['code']['build_command'] = '$RUSTC mod.rs'
+        
+        os.environ["RUSTC"] = "rustc -Zsave-analysis"
```
erikrose: It's weird that we do this in config_input() here and in setup_class() everywhere else. Also, can we factor up at least the string "rustc -Zsave-analysis" or perhaps the whole operation?

dxr/plugins/rust/tests/test_data/test_fields.py
erikrose: (Per IRC) changes to rustc broke stuff, and we haven't noticed because our copy is pinned to a June release.

erikrose:
That's all I have to say about that. :-) I agree your changes are the expedient thing for the moment, so, if you can do those few little cleanups, we can get this merged. Thanks for keeping us current, and sorry for the delay in review. Btw, can you either rebase or merge master in? Otherwise, the tests won't run next time you push, since we switched to Docker and Travis.

dxr/plugins/rust/tests/test_cargo/test_cargo.py
nrc: (Also per IRC) it sounds like RUSTC_ARGS is coming to Cargo in the near-ish future. Therefore I'll leave the painful script approach for now (rather than shipping the script with DXR) with the expectation that I will move to the env var approach asap.

nrc:
Comments addressed and rebased (Unpinning Rust coming as a separate PR)

nrc:
And some more changes

PR #498, opened 2015-12-10
Unpin the version of rust.

erikrose:
Nightly seems to have been working in prod for some time, so let's track it for awhile.


erikrose:
The rust tests started failing.

PR #499, opened 2015-12-14
Move venv to a data volume container so it persists across rebuilds o…

erikrose:
…f the dev container.


erikrose:
I don't know where the commit went, but it's d8b8caa9e93516ef985df726de1d36a53443b99a.

PR #500, opened 2015-12-16
Get the clang plugin to recognize multiple directly overridden functions

kleintom:
If a virtual function overrides virtuals in multiple direct base classes, all of
the overriddens should be recognized.

### DIFF
--- a/dxr/plugins/clang/condense.py
+++ b/dxr/plugins/clang/condense.py
@@ -22,9 +22,9 @@ class UselessLine(Exception):
     """A CSV line isn't suitable for getting anything useful out of."""
 
 
-POSSIBLE_KINDS = set(['call', 'macro', 'function', 'variable', 'ref',
-                       'type', 'impl', 'decldef', 'typedef', 'warning',
-                       'namespace', 'namespace_alias', 'include'])
+POSSIBLE_KINDS = set(['call', 'macro', 'function', 'func_override', 'variable',
+                      'ref', 'type', 'impl', 'decldef', 'typedef', 'warning',
+                      'namespace', 'namespace_alias', 'include'])
 
 
 def c_type_sig(inputs, output, method=None):
@@ -103,7 +103,7 @@ def process_override(overrides, overriddens, props):
     """
     # It may not be necessary to have a list here. In multiple inheritance,
     # does clang ever consider a method to override multiple other methods, or
-    # is it at most one each?
+    # is it at most one each?  Answer: clang recognizes multiple overriddens.
     overrides.setdefault(props['qualname'], set()).add(
             (props['overriddenqualname'], props['overriddenname']))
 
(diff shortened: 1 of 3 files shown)


kleintom:
A couple notes:

1) As far as I can tell, the old "overriddenloc" condense value wasn't being used, so I removed it.

2) I created a new "func_override" condense POSSIBLE_KIND, and now each override gets its own func_override csv line; this makes it more similar to the way "impl" works, and seems cleaner to me than the other alternative I considered, which was to keep all of the override information under a "function" kind, but make "overriddenname" and "overriddenqualname" mutli-valued (via comma-separated values, or whatever).

dxr/plugins/clang/condense.py
```diff
@@ -315,10 +315,9 @@ def listify_keys(d):
     condense(
         lines_from_csvs(csv_folder, '*.csv'),
         {'impl': partial(process_impl, parents, children),
-         'function': partial(process_override, overrides, overriddens)},
-        predicate=lambda kind, fields: (kind == 'function' and
-                                        'overriddenname' in fields) or
-                                       kind == 'impl')
+         'func_override': partial(process_override, overrides, overriddens)},
```
erikrose: Kudos on a very good design decision. It's so much better to have a dedicated kind for overrides rather than trying to shoehorn a multi-valued key into `function`.

dxr/plugins/clang/dxr-index.cpp
```diff
@@ -461,9 +461,12 @@ class IndexConsumer : public ASTConsumer,
     if (d->isThisDeclarationADefinition() ||
         d->isPure())  // until we have better support for pure-virtual functions
     {
-      beginRecord("function", d->getLocation());
-      recordValue("name", d->getNameAsString());
-      recordValue("qualname", getQualifiedName(*d));
+      SourceLocation functionLocation = d->getLocation();
```
erikrose: Do these add anything, like clarity, in your opinion, or are they just debugging expansions that didn't get removed?

erikrose:
Looks good to me! Just let me know what's up with those new declarations I noted.

I'm not at all surprised you found unused things like `overriddenloc` in there. The clang plugin is pretty messy. I did one round of cleaning when we landed the switch to elasticsearch, but it could do with some loving care from someone who codes in C++ all day. That cache of FileInfos has always leaked, for one fun example.

kleintom:
The new declarations were just so that those values wouldn't need to be recomputed in each iteration of the overrides loop below ... but I doubt it makes much difference either way (and maybe the compiler can even figure out that it doesn't need to recompute them in any case).  I'd be fine with removing them, just let me know which way you prefer (or remove them when you merge if that's easier :).

erikrose:
Ah, I didn't see those uses. Looks great as is! Thanks for another great patch! :-D

PR #502, opened 2015-12-17
Show the "Find base classes"/"Find subclasses" clang menu option only…

kleintom:
… when base classes, respectively subclasses, exist.

Also added some new "Find overrides"/"Find overriddens" menu tests.


kleintom:
(Re-pushed with the format file updated this time.)

kleintom:
Didn't notice these two issues until after I submitted the first commit, but I think they fit with this pr...

dxr/plugins/clang/menus.py
```diff
@@ -118,31 +118,40 @@ def _more_menu_items(self, (macro_name,)):
...
 
-    def _more_menu_items(self, (qualname, kind)):
+    def _more_menu_items(self, (qualname, kind, has_subclass, has_base_class)):
         """Return menu for type reference."""
+        def kind_plural():
+            if kind == 'class':
+                return 'classes'
+            elif kind == 'struct':
+                return 'structs'
+            raise TypeError('Unexpected kind: %s' % kind)
```
erikrose: I'd rather move this to indexing time, as people using the web interface should never see exceptions. Also, does it need to be a fatal exception? It seems like we could fall back to a more generic description like "type" or "thing".

dxr/plugins/clang/menus.py
```diff
@@ -118,31 +118,40 @@ def _more_menu_items(self, (macro_name,)):
 class TypeRef(_QualnameRef):
     @classmethod
     def _condensed_menu_data(cls, tree, prop):
-        """Hang onto qualname *and* kind."""
         return (super(TypeRef, cls)._condensed_menu_data(tree, prop) +
-                (prop.get('kind', ''),))
+                (prop.get('kind', ''), 'has_subclass' in prop, 'has_base_class' in prop))
```
erikrose: For example, we could default to 'type' or something here, add a case to kind_plurals, and have done with it.

dxr/plugins/clang/condense.py
```diff
@@ -152,6 +152,16 @@ def _process_loc(locstring):
     return src, Position(row, col)
 
 
+def process_maybe_impl(parents, children, props):
```
erikrose: I'd find docstrings really helpful on these "maybe" functions, since the names aren't the clearest.

erikrose:
Looks good to me otherwise!

dxr/plugins/clang/menus.py
```diff
@@ -118,31 +118,41 @@ def _more_menu_items(self, (macro_name,)):
...
     def _condensed_menu_data(cls, tree, prop):
-        """Hang onto qualname *and* kind."""
         return (super(TypeRef, cls)._condensed_menu_data(tree, prop) +
-                (prop.get('kind', ''),))
+                (prop.get('kind', 'type'), 'has_subclass' in prop, 'has_base_class' in prop))
 
-    def _more_menu_items(self, (qualname, kind)):
+    def _more_menu_items(self, (qualname, kind, has_subclass, has_base_class)):
         """Return menu for type reference."""
+        def kind_plural():
```
erikrose: This is fine for now. Any more branches, and we'd probably want to switch to a dict and `.get(kind, 'types')`.

erikrose: Maybe someday we'll get around to internationalizing the text in DXR and use the full-powered gettext routines. :-)

erikrose:
Thanks for another good patch, Tom!

erikrose:
@kleintom It looks like something in here is causing segfaults when we try to index mozilla-central or comm-central. Any ideas? (We just noticed because, due to a misconfig, we hadn't deployed this until recently.)

erikrose:
Though, after glancing at the files changed, that seems unlikely. Perhaps it's a coincidence.

kleintom:
Nothing's jumping out at me...

PR #505, opened 2015-12-29
Don't show "Untracked file" in the Nav pane when a file is tracked.

kleintom:


kleintom:
Unless I'm mistaken, it looks like the (what is now) `if is_textual_image(self.path):` conditional got unintentionally inserted between an if/else here: 
https://github.com/mozilla/dxr/commit/f36c41e961ab411da8cb0e4d59f6b9257ee17a5c#diff-27050266ad848025893af5a3c2ffd185L457

kleintom:
Let me know if you'd like a test for this one.  Thanks!

erikrose:
Good find! This is certainly an improvement, as it no longer weirdly shows both "Permalink" and "Untracked" at the same time. However, you've made evident a deeper bug: I don't think we've ever displayed "Untracked file" at the right time. Here, we show it only when there is no VCS checkout at all: in other words, all files in an entire tree would say "Untracked", so there's no useful contrast. What we should do is call `vcs.is_tracked(...)` and emit a permalink only if it's true—"Untracked file" otherwise.

kleintom:
I think that's what it's doing - `self.vcs` on the FileToIndex comes from a call to `vcs_cache.vcs_for_path(path)` on the corresponding TreeToIndex, which internally does the `is_tracked(...)` check, so the `self.vcs` for a core FileToIndex object is specific to whether or not `self.path` is tracked.  And the testing looks good, unless I'm missing something. :)

Maybe we should call it `vcs_for_path` or something to be clearer?

pelmers:
@kleintom is right here -- self.vcs is associated with the specific file being indexed rather than whether the whole tree has a vcs root. This looks good to me.

erikrose:
Thanks for taking a look @pelmers!

PR #507, opened 2015-12-30
dxr-index.cpp updates

kleintom:
I indexed (a couple weeks old version of) firefox several times with and without the patch, and all runs were within a minute of each other, sometimes slower, sometimes faster with the patch.  (That's a little surprising to me, but I also ran a debug indexing version where I was writing to files from the plugin, and the writes were definitely happening, so I'm pretty sure those indexing runs actually used the patches...)

I tried not to be too draconian on the style "cleanup", so in the end maybe no one will be happy!  :P  Feel free to suggest more changes, or changes you prefer should be taken back.

### DIFF
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -60,10 +60,6 @@ const char *hash(std::string &str) {
   return hashstr;
 }
 
-std::string srcdir;
-std::string output;
-std::string tmpdir; // Place to save all the csv files to
-
 struct FileInfo {
   FileInfo(std::string &rname) : realname(rname) {
     interesting = rname.compare(0, srcdir.length(), srcdir) == 0;
@@ -81,6 +77,8 @@ struct FileInfo {
   std::string realname;
   std::ostringstream info;
   bool interesting;
+  static std::string srcdir;  // the project source directory
+  static std::string output;  // the project build directory
 };
 typedef std::shared_ptr<FileInfo> FileInfoPtr;
 
@@ -131,6 +129,7 @@ private:
   std::map<std::string, FileInfoPtr> relmap;
   LangOptions &features;
   DiagnosticConsumer *inner;
+  static std::string tmpdir;  // Place to save all the csv files to
 
   const FileInfoPtr &getFileInfo(const std::string &filename) {
     std::map<std::string, FileInfoPtr>::iterator it;
@@ -170,6 +169,8 @@ public:
     return new IndexConsumer(ci);
 
   }
+
+  static void setTmpDir(const std::string& dir) { tmpdir = dir; }
 
   // Helpers for processing declarations
   // Should we ignore this location?
@@ -1238,12 +1239,13 @@ protected:
       D.Report(DiagID) << args[0];
       return false;
     }
-    srcdir = abs_src;
+    FileInfo::srcdir = abs_src;
     const char *env = getenv("DXR_CXX_CLANG_OBJECT_FOLDER");
+    std::string output;
     if (env)
       output = env;
     else
-      output = srcdir;
+      output = abs_src;
     char *abs_output = realpath(output.c_str(), NULL);
     if (!abs_output) {
       DiagnosticsEngine &D = CI.getDiagnostics();
@@ -1252,10 +1254,12 @@ protected:
       D.Report(DiagID) << output;
       return false;
     }
-    output = realpath(output.c_str(), NULL);
+    output = abs_output;
     output += "/";
+    FileInfo::output = output;
 
     const char* tmp = getenv("DXR_CXX_CLANG_TEMP_FOLDER");
+    std::string tmpdir;
     if(tmp){
       tmpdir = tmp;
     }else
@@ -1269,8 +1273,9 @@ protected:
       D.Report(DiagID) << tmpdir;
       return false;
     }
-    tmpdir = realpath(tmpdir.c_str(), NULL);
+    tmpdir = abs_tmpdir;
     tmpdir += "/";
+    IndexConsumer::setTmpDir(tmpdir);
 
     free(abs_src);
     free(abs_output);
@@ -1284,6 +1289,10 @@ protected:
 
 };
 
+// define static members
+std::string FileInfo::srcdir;
+std::string FileInfo::output;
+std::string IndexConsumer::tmpdir;
 }
 
 static FrontendPluginRegistry::Add<DXRIndexAction>
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -13,6 +13,7 @@
 
 #include <iostream>
 #include <map>
+#include <memory>
 #include <sstream>
 #include <stdio.h>
 #include <stdlib.h>
@@ -81,6 +82,7 @@ struct FileInfo {
   std::ostringstream info;
   bool interesting;
 };
+typedef std::shared_ptr<FileInfo> FileInfoPtr;
 
 class IndexConsumer;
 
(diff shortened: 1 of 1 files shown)


erikrose:
This Christmas or Thanksgiving or whatever, I am thankful that you did this in well-demarcated multiple commits. :-)

kleintom:
Yup, Merry Thanksgivingmas, since it started out as one big bad patch. :-)

erikrose:
I'll get to reviewing this in depth in the new year, but, at a glance, it looks sane to me. I also have to bone up on modern C++ so I know what you're talking about. :-)

Say, did you have to crank up the ES_HEAP_SIZE on the es docker image in order to index moz-central?

kleintom:
Now that you mention it...  I haven't been using docker since I had already gotten things working in my own environment, and I'm getting a libstdc++ error everytime anything tries to index using the clang plugin in docker:
```
dxr@fda5ebb09ffb:~/dxr/tests/test_basic$ dxr index
Starting tree 'code'.
Running pre_build  [####################################] 100%
Building tree
Build command for 'code' failed, exited non-zero.
Log follows:
    | rm -rf code
    | clang++ -Xclang -load -Xclang /home/dxr/dxr/dxr/plugins/clang/libclang-index-plugin.so -Xclang -add-plugin -Xclang dxr-index -Xclang -plugin-arg-dxr-index -Xclang /home/dxr/dxr/tests/test_basic/code -o code main.c
    | clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated
    | error: unable to load plugin '/home/dxr/dxr/dxr/plugins/clang/libclang-index-plugin.so': '/usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /home/dxr/dxr/dxr/plugins/clang/libclang-index-plugin.so)'
    | make: *** [code] Error 1
```
I had planned on getting around to looking at it, but never did.
Shall I open a bug or do you already know what this is?  I'm happy to try things out...

erikrose:
I'm not sure, but it sounds like the version of glibc++ that the clang compiler plugin was built against isn't around. Try rebuilding it?

dxr/plugins/clang/dxr-index.cpp
```diff
@@ -1245,34 +1280,41 @@ class DXRIndexAction : public PluginASTAction {
...
         "Temporary directory '%0' does not exist");
       D.Report(DiagID) << tmpdir;
       return false;
     }
-    tmpdir = realpath(tmpdir.c_str(), NULL);
+    tmpdir = abs_tmpdir;
     tmpdir += "/";
+    IndexConsumer::setTmpDir(tmpdir);
+
+    free(abs_src);
```
erikrose: We can't wait till here to free them all, because there are several early returns, right?

kleintom: My thinking was that if we took any of the early returns then the compile was about to end and the freeing would be a moot point...  I guess it does look bad though - what do you think?  I could also move the frees up to as soon as each var is no longer needed, which would reduce the amount of freeing needed in each early return.

kleintom: Actually if the frees are moved up as far as possible then no frees are needed in the early returns, so that's probably better.

kleintom:
> I'm not sure, but it sounds like the version of glibc++ that the clang compiler plugin was built against isn't around. Try rebuilding it?

Of course, yeah - I got fooled when I saw make in docker installing all new python packages into thinking the plugin had been newly built as well, but in fact it was using the plugin I had compiled on my host fedora system.  Thanks for the nudge!

erikrose:
Sure! Glad it was easy. Too bad make isn't smart enough to realize what's going on.

dxr/plugins/clang/dxr-index.cpp
kleintom: Or we could stick that memory in a unique_ptr with a custom deleter and let it worry about when to delete:
`std::unique_ptr<char, decltype(&free)> abs_src(realpath(args[0].c_str(), nullptr), free);`
Yeah, I like that better (maybe with a typedef for that long type).  The concern I have with freeing as early as possible is that someone could edit later to use the freed var after the free, which is undefined behavior that may or may not get caught in testing with a given compiler.  Unlikely I think, but not impossible I guess.

dxr/plugins/clang/dxr-index.cpp
```diff
@@ -80,45 +79,57 @@ struct FileInfo {
   std::string realname;
   std::ostringstream info;
   bool interesting;
+  static std::string srcdir;  // the project source directory
```
erikrose: Why persist the strange decision of making these singletons? Conceptually, are they not one-per-DXRIndexAction-instantiation? Perhaps in a later PR, we could move them into IndexConsumer. We probably also have to put IndexConsumer in charge of creating FileInfos, since that's what's interested in most of those vars.

dxr/plugins/clang/dxr-index.cpp
erikrose: Actually, I don't think we should free any of those long-lived vars in ParseArgs at all: the memory they point to must remain unmolested the whole time the AST walker runs, and ParseArgs is long exited by the time the AST walker even starts.

dxr/plugins/clang/dxr-index.cpp
```diff
@@ -1115,29 +1135,31 @@ class IndexConsumer : public ASTConsumer,
       StringRef searchPath,
       StringRef relativePath,
       const Module *imported) {
-    PresumedLoc presumedHashLoc;
-    FileInfo *target, *source;
-    SourceLocation targetBegin, targetEnd;
+    PresumedLoc presumedHashLoc = sm.getPresumedLoc(hashLoc);
```
erikrose: Why move all this up into the declarations? Does it have a memory leak consequence? It was pushed into the ifs as a speed optimization.

erikrose:
In an effort to get moz-central indexing again, I'm having a try at "rolling forward", so I made some of my suggested changes (see https://github.com/erikrose/dxr/tree/kleintom-dxr-index_updates) and am seeing if that crashes while indexing.

dxr/plugins/clang/dxr-index.cpp
erikrose: (I now see that some of this was likely due to switching to returning refs from getFileInfo().

dxr/plugins/clang/dxr-index.cpp
kleintom: > Actually, I don't think we should free any of those long-lived vars in ParseArgs at all (even when they go out of scope): the memory they point to must remain unmolested the whole time the AST walker runs, since we're just copying the ptrs into the static vars, and ParseArgs is long exited by the time the AST walker even starts.

Thanks for looking all this over!  I probably won't get to your other feedback until tomorrow or Friday, but I think on this one that the assignment actually copies out the underlying char data, not just the pointer:
http://en.cppreference.com/w/cpp/string/basic_string/operator%3D
I did a little stub test with the same setup as in dxr-index.cpp that seems to confirm that: after the `FileInfo::srcdir = abs_src;` assignment, `abs_src` and `FileInfo::srcdir.data()` point to different data.

erikrose: Ah, right you are. I didn't see that the destinations were std::strings.

dxr/plugins/clang/dxr-index.cpp
kleintom: > Why persist the strange decision of making these singletons? Conceptually, are they not one-per-DXRIndexAction-instantiation?

Are you thinking that more than one DXRIndexAction could exist at the same time, in a single `clang++` call?  And that the different `DXRIndexActions` might have different `srcdir`, `outdir`, or `tmpdir` values?  If so then I agree that those values can't be static.  But I didn't think either of those things did/could happen, in which case I think it makes sense for at least `FileInfo::srcdir` and `FileInfo::outdir` to be static, since every one of the many `FileInfo`s created will be using the same value for `srcdir` and `outdir`.

It's never been entirely clear to me, but the test runs I've looked at seem to also indicate that only one `IndexConsumer` gets created per `DXRIndexAction`, so, if that's the case, then there's no need to make `IndexConsumer::tmpdir` static, except for the fact that if it isn't static, then `DXRIndexAction` will have to store its value from when it's created in `ParseArgs` to when it can be passed at `IndexConsumer` creation time in `CreateASTConsumer`.  But in this case (even if only one `IndexConsumer` is ever created) I still think it makes logical sense for `IndexConsumer::tmpdir` to be static, since every `indexConsumer` instance created by a given `DXRIndexAction` is going to use the same value (unless of course that value can change over the course of a given compile!).

kleintom: > We probably also have to put IndexConsumer in charge of creating FileInfos, since that's what's interested in most of those vars.

I think that's already true: `IndexConsumer::getFileInfo` seems to be the only creator of `FileInfo`s :-)

dxr/plugins/clang/dxr-index.cpp
kleintom: > (I now see that some of this was likely due to switching to returning refs from getFileInfo().)

Yeah - thanks for moving what didn't need to be moved back.

PR #508, opened 2016-01-03
Only display "VCS Links" when there's a useable upstream link.

kleintom:
This also prevents exceptions when indexing a mercurial repository with no
default path or a git repository with the omniglot plugin and no github
upstream.


pelmers:
r+, thanks!

PR #515, opened 2016-02-04
Added support for Rust macros and macro calls

DanielJCampbell:
r? @nrc 

Also I've removed the std hack stuff - with the changes to save-analysis re: generated-code, the std crate is already being written out to the CSV file.

### DIFF
--- a/dxr/plugins/rust/__init__.py
+++ b/dxr/plugins/rust/__init__.py
@@ -70,6 +70,8 @@ class FileToIndex(indexers.FileToIndex):
     def refs(self):
         classes_and_tables = [(refs.FunctionRef, 'functions'),
                               (refs.FunctionRefRef, 'function_refs'),
+                              (refs.MacroRef, "macros"),
+                              (refs.MacroRefRef, "macro_refs"),
                               (refs.VariableRef, 'variables'),
                               (refs.VariableRefRef, 'variable_refs'),
                               (refs.TypeRef, 'types'),
@@ -105,6 +107,8 @@ class FileToIndex(indexers.FileToIndex):
         return iterable_per_line(with_start_and_end(split_into_lines(chain(
             self.file_needles('function', 'functions'),
             self.file_needles('function_ref', 'function_refs'),
+            self.file_needles('macro', 'macros'),
+            self.file_needles('macro_ref', 'macro_refs'),
             self.file_needles('var', 'variables'),
             self.file_needles('var_ref', 'variable_refs'),
             self.file_needles('type', 'types'),
@@ -262,6 +266,8 @@ class TreeData:
         self.variable_refs = []
         self.functions = {}
         self.function_refs = []
+        self.macros = {}
+        self.macro_refs = []
         self.types = {}
         self.type_refs = []
         self.impl_defs = {}
@@ -402,7 +408,6 @@ class TreeToIndex(indexers.TreeToIndex):
 
         # We need to do this once per crate whilst the current crate is still current
         self.generate_scopes()
-        self.std_hack()
 
     def process_csv(self, file_name, header_only):
         try:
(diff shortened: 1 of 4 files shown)


dxr/plugins/rust/refs.py
```diff
@@ -126,6 +126,27 @@ def menu_items(self):
         return menu
 
 
+class MacroRef(_RustRef):
+    def prepare_menu_data(self, tree_index, datum):
+        self.hover = datum['qualname']
+        return trim_dict(datum, ['qualname'])
+
+    def menu_items(self):
```
nrc: I think here and below we also want to add the `generic_variable_menu` items

nrc:
LGTM, modulo the `generic_variable_menu` thing.

r? @erikrose

dxr/plugins/rust/refs.py
DanielJCampbell: I'm not quite following - that shouldn't have any impact on macros.

Unless you mean you want to be able to jump to macro references from a macro reference?

nrc: Oh yeah, sorry, should have looked that up, it wasn't what I was expecting. Anyway, yeah, I guess you should add `find_references_...` to RefRef

DanielJCampbell: Done

erikrose:
Do the mentioned "changes to save-analysis re: generated-code" kick in by the rustc nightly of 2016-01-25? That's where DXR is stuck atm due to DXR's rust tests not passing on later ones.

dxr/plugins/rust/__init__.py
```diff
@@ -857,6 +838,11 @@ def process_function(args, tree):
     process_function_impl(args, tree)
 
 
+def process_macro(args, tree):
+    tree.data.macros[args['qualname']] = args
+    tree.add_to_lines(args, ('macros', args))
+
```
erikrose: (Just a nit: 1 skipped line between methods of a class)

dxr/plugins/rust/refs.py
```diff
@@ -126,6 +126,29 @@ def menu_items(self):
...
+        return [find_references_menu_item(self.tree, self.menu_data["qualname"], "macro-ref", "macro")]
+
+class MacroRefRef(_RustRef):
+    def prepare_menu_data(self, tree_index, datum):
+        self.hover = datum['qualname']
```
erikrose: Another, slightly less imperative way to narrow down the `datum` dict to just the "qualname" key would be to subclass `_KeysFromDatum` rather than `_RustRef` and say `keys = ['qualname']` out at the class level. The nice thing about that is that it becomes obvious from examining the MacroRefRef class what it stores in elasticsearch. However, most other things still call trim_dict, so it's not a blocker.

erikrose:
Looks okay otherwise. I just want to wait for an answer to the "changes to save-analysis re: generated-code" question before merging.

DanielJCampbell:
https://github.com/rust-lang/rust/pull/31097 Is the pull that changed save analysis so that it exported the standard crates (when included). Not sure when it made it into nightly, but the timeframe seems pretty close to when it started breaking.

Also, before this commits, I'll implement the suggested changes, and I have a question. Using this build, searching for macros and macro references from the search bar has some weird behaviour: mainly, you can't find anything unless you use the exact qualified name. `macro:foo` returns nothing, `macro:foo::12345678` returns a result (assuming that macro exists). Do you have any idea why this would happen? As far as I'm aware I just copied the rust implementation for functions, but I must have missed something.

erikrose:
> you can't find anything unless you use the exact qualified name

When you search for macro:blah, DXR looks in elasticsearch for lines tagged with a `name` or `qualname` of "blah". (That's what a `QUALIFIED_LINE_NEEDLE` is: a needle in the elasticsearch haystack that has both a short name and a qualified name.) I would guess that you're storing "macro:foo::12345678" as the `qualname` and perhaps also failing to store "foo" as the `name`.

erikrose:
Since, as you say, the code you added looks just like the processing code for functions, I suspect the wrong info is getting into `tree` earlier in the process.

DanielJCampbell:
Thanks for that, I'm pretty sure I know where the problem is now. I'll have the recommended changes up by the end of today.

DanielJCampbell:
Should be good to go now. One issue is I can't write tests, as nightly rust hasn't caught up yet.

PR #517, opened 2016-02-11
Fix bugzilla 1223442

jlmarr10824:
Had lifetime problems with the module scope variable so switched to
using function scope.

### DIFF
--- a/dxr/plugins/clang/condense.py
+++ b/dxr/plugins/clang/condense.py
@@ -17,6 +17,13 @@ from funcy import decorator, identity, select_keys, imap, ifilter, remove
 from dxr.indexers import FuncSig, Position, Extent
 from dxr.utils import frozendict
 
+import re
+from os import listdir
+from os.path import isfile, join
+from collections import defaultdict
+
+db = defaultdict(list)
+csvfiles = None
 
 class UselessLine(Exception):
     """A CSV line isn't suitable for getting anything useful out of."""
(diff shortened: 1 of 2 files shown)


dxr/plugins/clang/condense.py
```diff
@@ -17,6 +17,10 @@
 from dxr.indexers import FuncSig, Position, Extent
```
erikrose: I notice that both files have had their permissions changed (see "100644 → 100755") on the "Files changed" tab. I'm sure you didn't mean to do that, so have a look at your setup and see what might be changing permissions behind your back. If Windows is involved, that's one place I'd look.

dxr/plugins/clang/condense.py
```diff
@@ -283,10 +287,36 @@ def lines_from_csv(path):
...
+       start with the specified file path digest.
+
+    All lines are lists of strings.
+
+    :arg folder: The folder in which to look for CSVs
+    :arg csvdict: A dict of sha1 hex digest keys
+    :arg path_digest: A string produced by sha1(file_path).hexdigest()
+
+    """
+    def lines_from_csv(path):
```
erikrose: This looks like a copy-and-paste of the same inner function above; let's factor it out and not repeat ourselves.

erikrose:
Did you happen to try the radix method mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1223442#c0? I wonder how it compares for speed with this; it would certainly use less RAM, though it would do more IO in calling listdir() all the time.

erikrose:
I'm having a little swing at refactoring this. It's a good start. More tomorrow!

jlmarr10824:
- I believe I changed the permissions while I was thrashing about trying to figure out why nosetest was not working for me.  Sorry that made it into the submission.  I assume you can rectify that.
- Thanks for refactoring the copy/paste.
- I didn't try the radix approach.  I've had bad experiences with performance of file system globs in perl, and assumed the same would apply in python.

erikrose:
Here's the refactoring I did: https://github.com/mozilla/dxr/compare/master...erikrose:no-globbing?expand=1. I'm benching it now.

jlmarr10824:
Ok, great.   Since my previous implementation was buggy I'm no longer expecting an order-of-magnitude improvement, but still something significant.  Test still running on my end.

erikrose:
And now I can't get Firefox to build: https://bugzilla.mozilla.org/show_bug.cgi?id=1249405.

erikrose:
@pelmers Could you take a look at this?

dxr/plugins/clang/condense.py
```diff
@@ -342,6 +371,15 @@ def listify_keys(d):
     parents = {}
     children = {}
 
+    # create a list of csvfiles then populate dict 'db' with a list of csv's for each filepath digest
+    csvfiles = None
```
pelmers: This line looks unnecessary

pelmers:
I ran Erik's no-globbing branch on a compile of Python and find that it's finishing some 25% faster (3 minute index vs 4 minute index). I'll try to see what the difference is on mozilla-central overnight, but current master completed in 3:17:33 for me.

pelmers:
Tried to run it overnight, but the process hangs at 0% of the "indexing files" stage.

erikrose:
Can you try one more time? There used to be a transient hang with concurrent.futures on Vagrant; it may also exist on Docker.

PR #518, opened 2016-03-02
Place clang macro definition tooltip on macro refs instead of macro def.

kleintom:


dxr/plugins/clang/dxr-index.cpp
```diff
@@ -1087,7 +1095,7 @@ class IndexConsumer : public ASTConsumer,
             text[i] != '\t' && text[i] != '\n')
           text[i] = '?';
       }
-      recordValue("text", text, true);
+      macromap.insert(make_pair(nameStart, text));
```
erikrose: It seems like this will keep the full text of any macro body in memory forever. I'm not sure how that's going to go on something as big as Firefox, which can barely be compiled in a 7GB VM as is. What do you think?

erikrose: If it is indeed a problem, one alternative would be to build the map in the post-processing global phase, when memory is more plentiful.

dxr/plugins/clang/tests/test_macros.py
```diff
@@ -136,3 +137,27 @@ def test_decls(self):
...
+    def test_macro_titles(self):
+        """Test that a ref to a macro gets a title tooltip containing the
+        definition of the macro, and that the macro def doesn't get a title.
+        Also make sure the macro tooltip skips the macro's leading whitespace
+        up to and including an initial backslash newline.
+
+        """
+        markup = self.source_page('main.cpp')
+        # The SD ref link is followed by the '<' of the comment markup.
+        ok_('title=" int s;\\\n int d;">SD</a> <' in markup, msg=markup)
```
erikrose: I wonder if there's a more obvious way to express the tests. The differences are mighty small here, and I suspect that they're tightly coupled to syntax highlighting we happen to be using.

dxr/plugins/clang/dxr-index.cpp
kleintom: > It seems like this will keep the full text of any macro body in memory forever.

Isn't there one IndexConsumer per translation unit compilation, and that IndexConsumer gets freed as soon as that compilation completes?  That's what I've been thinking at least.

I did an m-c compile with some logging just to make sure - I never see the logging in the destructor for IndexConsumer that I added appear (I assume it's just getting quelched or appears too late - I do see IndexConsumer creation for each compilation unit), but it gave some other data that might be helpful here:
of 3028 translation units that had a macro def, the average final length of macromap was 1,692, the median was 1,440, and the max was 13,051, i.e. in m-c there are at most 13,051 macro defs in any given translation unit.  Those numbers are bigger than I was expecting, but still, if you take an average macro-def string of 800 bytes (10 lines average? I would guess that's high?), say, then with 13,051 entries, the largest macromap would be roughly 13,051 * (4 byte key (SourceLocation) + 800 byte macro-def-string + 10(?) bytes overhead) = 10,623,514 bytes or about 10.5M extra memory use, and the average compile would use about 1.4M extra memory for macromap.  And then you might have a bunch of compiles running at once, so maybe, I don't know, an extra 10-50 Mb?

Also on my machine with 8G ram I didn't have any problems compiling m-c under dxr (no swapping that I ever noticed, not even close actually), and my running firefox actually always had more memory use than any of the compiles going on as far as I saw, so I would be inclined to say it seems okay to me (if those numbers are anywhere close...).  But if there's more room to spare in the indexing step (is there? the impression I'd had in the past was that parts of the indexing step were using more memory than the compilation step), I could move it there.

dxr/plugins/clang/tests/test_macros.py
kleintom: > I wonder if there's a more obvious way to express the tests.

Agreed, I'll try to come up with something else.  Also, I noticed I left a debugging `msg=markup` in that `ok_`.

dxr/plugins/clang/dxr-index.cpp
erikrose: Nope! Sold. I didn't know you had compiled m-c under it, nor did I notice that it was in an object that's discarded per translation unit. Your back-of-the-envelope calculations look good as well; that's exactly what I wanted to see. :-) It'll be a tad faster in C++ anyway, although we'll pay a little more IO since we're repetitively writing the macro definition out to the CSVs on every reference. We'll go with this! Thanks!

dxr/plugins/clang/tests/test_macros.py
erikrose: Actually, I didn't dislike the msg=markup, since that's the first thing you'd have to do to debug a failure anyway.

kleintom:
I finally noticed that the tooltip wasn't displaying macro args, and that the macro args we were recording on the macro def were never being used, so I moved the args list out of the def and into the ref tooltips.  The way it currently works is we remove any "#define NAME" to create the tooltip if there are args - so "#define DO(x) stuff(x)" would get a tooltip of "(x) stuff(x)" (what do you think?) - and if there are no args then we just remove all first-line leading whitespace like we did before.

FYI, I indexed m-c with this last night and got an es index of size 17,571,180,714 bytes, which compares with an m-c index from December 28 that's somehow survived with a size of 17,088,800,911 bytes - so looks like this adds about half a gig of index...

erikrose:
Thanks, Tom! Sorry it took so long to get it merged in!

PR #521, opened 2016-03-18
Browser history processing

kleintom:
Handle browser history moves better:

* Don't push history every time we run a doQuery since that leads to duplicate history entries (one on the page load, then another a couple seconds later when the doQuery pushHistory timer pops).
* Run a reload on popstate when we go from a query url to a non-query url.
* Now that non-query views can be reloaded in history, make sure we're handling highlighting correctly on re-views.
* Correctly handle (with regard to history) the case where a context menu entry points to an anchor on the current page.
* Dismiss the context menu when we change history and when we click a link on the context menu.

### DIFF
--- a/dxr/static_unhashed/js/dxr.js
+++ b/dxr/static_unhashed/js/dxr.js
@@ -60,7 +60,9 @@ $(function() {
 
     // We also need to cater for the above scenario when a user clicks on in page links.
     window.onhashchange = function() {
-        scrollIntoView(window.location.hash.substr(1));
+        if (window.location.hash) {
+            scrollIntoView(window.location.hash.substr(1));
+        }
     };
 
     /**
@@ -140,7 +142,8 @@ $(function() {
         resultsLineCount = 0,
         dataOffset = 0,
         previousDataLimit = 0,
-        defaultDataLimit = 100;
+        defaultDataLimit = 100,
+        lastURLWasSearch = false;  // Remember if the previous history URL was for a search (for popState).
 
     // Has the user been redirected to a direct result?
     var fromQuery = /[?&]?from=([^&]+)/.exec(location.search);
@@ -419,6 +422,7 @@ $(function() {
                                                          replace(/([&?])limit=\d+/, dropAmp).
                                                          replace('?&', '?');
                             history.pushState({}, '', displayURL);
+                            lastURLWasSearch = true;
                         };
                         if (redirect)
                             // Then the enter key was pressed and we want to update history state now.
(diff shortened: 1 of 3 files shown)


kleintom:
Rebased (no other changes); still looks good to me.  @pelmers r?

dxr/static_unhashed/js/dxr.js
```diff
@@ -411,7 +411,13 @@ $(function() {
...
                         var pushHistory = function () {
                             // Strip off offset= and limit= when updating.
-                            var displayURL = queryString.replace(/[&?]offset=\d+/, '').replace(/[&?]limit=\d+/, '');
+                            function dropAmp(fullMatch, ampOrQuestion) {
+                                // Drop a leading ampersand, keep a leading question mark.
+                                return (ampOrQuestion === '?') ? '?' : '';
+                            }
+                            var displayURL = queryString.replace(/([&?])offset=\d+/, dropAmp).
+                                                         replace(/([&?])limit=\d+/, dropAmp).
+                                                         replace('?&', '?');
```
kleintom: On second thought, maybe this is better:
```diff
- // Strip off offset= and limit= when updating.
- var displayURL = queryString.replace(/[&?]offset=\d+/, '').replace(/[&?]limit=\d+/, '');
+ // Strip off offset= and limit= when updating - leave a '?', drop an '&'.
+ var displayURL = queryString.replace(/(?:(\?)|&)offset=\d+/, '$1').
+                              replace(/(?:(\?)|&)limit=\d+/, '$1').
+                              replace('?&', '?');
```
It's a little harder to read (for me at least), but more professional-looking and concise I guess :P  Any preference?

dxr/static_unhashed/js/dxr.js
```diff
@@ -557,13 +580,16 @@ $(function() {
     // If on load of the search endpoint we have a query string then we need to
     // load the results of the query and activate infinite scroll.
     window.addEventListener('load', function() {
-        if (/search$/.test(window.location.pathname) && window.location.search) {
+        if (locationIsSearch()) {
```
pelmers: Seems we can simplify it a little,
```
lastURLWasSearch = locationIsSearch()
if (lastU..W..S...) {
}
```

pelmers:
It seems that the address bar still does not always match the page. If I browse to a page anchored on a line, then type a search without hitting enter, the location does not change (and no history pushes) even after results appear.

kleintom:
> If I browse to a page anchored on a line, then type a search without hitting enter, the location does not change (and no history pushes) even after results appear.

I'm not seeing that, can you give some more specific steps?
I tried:
* index test_basic
* load http://127.0.0.1:8000/code/source/README.mkd#1 in a new tab
* type `main` without hitting return in the search field

After a couple seconds the search url appears as expected.

pelmers:
Hmm I might not have been waiting enough, I can't reproduce that anymore

kleintom:
Well now I can :-)  Looking into it...

kleintom:
STR the issue we were seeing:

1) Click on a result that loads a file with enough lines to need a scrollbar (just browsing to a large file doesn't have the same effect, and the hash is irrelevant).

2) Type in a search that will return enough results to need infinite scroll.

Step 1) triggers an onscroll event, even with no scrolling being done (I guess because we empty and then refill the page?), so we set didScroll = true, which primes us for doing infinite scroll queries.

In step 2) the first set of results returned after typing is designed to only fill a page, so when we display that set of results and turn on scroll polling, the first scroll poll already decides we're ready for more results, so does another query, which cancels the historyWaiter timer that otherwise would have updated the url for the search (and since the second query was append-only, it doesn't update history).

The fix is to not cancel the historyWaiter timer in doQuery if we're only appending results.

I guess it would be nice to not be doing that immediate infinite scroll query in step 2, but I suggest leaving that for another bug since it's not hurting anything (and already existed before this patch).

pelmers:
Got it, thanks for digging into this one, Tom! r+

PR #522, opened 2016-03-22
Single result behavior and clang "Jump to definition"

kleintom:
feedback?

Commit 1 jumps to a unique result when a redirect query returns a single result.  It doesn't give an indication bubble the way direct search does (though it could), but it does maintain the query string in both the url and in the query field, so the user can easily edit their query if they change their mind.

Commit 2 uses the commit 1 behavior to reimplement clang's "Jump to definition" menu for a function by making that link a query for the qualname of the function, and then relying on the commit 1 behavior to jump to the func def location when the query returns a single result.  This is based  on Erik's idea at https://bugzilla.mozilla.org/show_bug.cgi?id=986522#c13

The main issue I can see with doing it this way is that I don't see any way to know for sure whether the actual function def location lies in or outside of the source tree (you don't know until you do the search).  If the function def location clang returns is in the source tree and the actual def location is not, then the "Jump to definition" link will just return 0 results, while if the actual def location is in the search tree but the location clang returns is not then the "Jump to definition" menu option won't be shown at all (as currently written).

(The other (minor) issue is that if the def of a function is in the same file as the ref, then "Jump to definition" will reload the page, something the current code doesn't require.)

If this basically looks good let me know and I'll write some tests.


pelmers:
Commit 1 seems to address https://bugzilla.mozilla.org/show_bug.cgi?id=1072920 as well

kleintom:
Sorry for stealing part 1, Peter - I hadn't realized it was already assigned to you.

kleintom:
I went with having a single file result redirect to the file in order to match the core plugin `direct_path` direct search behavior described at https://bugzilla.mozilla.org/show_bug.cgi?id=1202603

erikrose:
> If the function def location clang returns is in the source tree and the actual def location is not, then the "Jump to definition" link will just return 0 results, while if the actual def location is in the search tree but the location clang returns is not then the "Jump to definition" menu option won't be shown at all (as currently written).

In what scenarios would those 2 eventualities occur?

pelmers:
No problem! I was not getting back around to it soon anyway.

dxr/static_unhashed/js/dxr.js
```diff
@@ -155,8 +155,13 @@ $(function() {
             searchUrl += '&case=' + isCaseSensitive;
         }
 
-        queryField.val(decodeURIComponent(fromQuery[1]));
+        queryField.val(decodeURIComponent(fromQuery[1].replace('+', ' ')));
         showBubble('info', viewResultsTxt.replace('{{ url }}', searchUrl));
+    } else {
+        var from_qQuery = /[?&]from_q=([^&]+)/.exec(location.search);
```
erikrose: Now that we render pages dynamically, we no longer need to do this through this fiddly JS regex claptrap. (It was a workaround from the days when we pre-rendered every HTML page and saved them to disk.) Now we can pass a "query=foo" and a "bubble=direct" or "bubble=single" querystring param and adjust the markup in the template.

erikrose: I think we should show a bubble, just to ward off any user surprise. Something along the lines of "There was only 1 search result, so I took you straight there."

dxr/app.py
```diff
@@ -170,6 +170,20 @@ def _search_json(query, tree, query_text, is_case_sensitive, offset, limit, conf
             return jsonify({'redirect': url_for('.browse', _anchor=line, **params)})
     try:
         count_and_results = query.results(offset, limit)
+        # If we're asked to redirect and there's a single result, redirect to the result.
+        if (request.values.get('redirect') == 'true' and
+            count_and_results['result_count'] == 1):
+            _, path, line, _ = next(count_and_results['results'])
+            line = line[0][0] if line else None
```
erikrose: It's worth considering whether it's desirable to redirect the user to a found FILE-domain result, e.g. when they're doing a `path:` search just to find out whether any such files exist. In those cases, they quite possibly aren't concerned with the contents of the file at all, and the slowness of loading a large file might be worse than what we have now. But I can likely be persuaded on this point. What do you think?

dxr/plugins/clang/menus.py
```diff
@@ -201,6 +203,33 @@ def _more_menu_items(self, (qualname,)):
...
+            # query below will return 0 results if the defloc is inside and the
+            # actual def is outside the source tree, and conversely this menu
+            # won't be displayed (even though it should be) if the defloc is
+            # outside and the actual def is inside the source tree.
+            menu = [{'html': "Jump to definition",
+                     'title': "Jump to definition",
+                     'href': url_for('.search', tree=self.tree.name,
+                                     q='+function:"' + self.menu_data[2] + '"',
+                                     redirect='true',
+                                     no_from='true'),
```
erikrose: We're starting to accumulate quite a few special-case querystring params for `search`. Do you see a way we could have fewer? If not, we should document them all in some central place.

dxr/plugins/clang/menus.py
```diff
@@ -201,6 +203,33 @@ def _more_menu_items(self, (qualname,)):
 class FunctionRef(_RefWithDefinition):
     """Ref for function definitions or references"""
 
+    def menu_items(self):
+        """Return a jump-to-definition menu item along with whatever others
+        _more_menu_items() returns.
+
+        """
+        path = self.menu_data[0]
```
erikrose: It doesn't look like `menu_data[0]` is really used here (beyond acting as a boolean), nor is `menu_data[1]`. So that's needlessly inflating the ES indices. There might not be good reason to subclass `_RefWithDefinition` anymore.

dxr/plugins/clang/menus.py
```diff
@@ -201,6 +203,33 @@ def _more_menu_items(self, (qualname,)):
...
+        if path is None:
+            menu = []
+        else:
+            # We get here when the defloc lies inside the source tree, but for
```
erikrose: Note to self: understand this next.

kleintom: > Note to self: understand this next.

I hadn't thought this through far enough, sorry.  There are two cases I was worried about, but now I think the `path is None` check is irrelevant to both of them, which is yet another reason to not derive from `_RefWithDefinition` anymore.

The first case is when you have, for example, a header file with a class that declares a method `C::foo()`.  If the header file is in the tree, and there's a ref in a file in the tree (which includes the header), but the actual implementation of `C::foo` is in a file outside the tree in a compilation unit different from the ref file, then when clang compiles the compilation unit that includes the function ref it will list the declaration in the header as the function def location for `foo` (because the actual implementation location is not in that compilation unit, and so the decl is the best it can do).  So the ref type is in the tree and so we show a context menu and a 'Jump to definition' option for that ref, but in actuality the def is not in the tree and so the query for it returns 0 results.

The other issue is if you have a ref to function `foo()` in the tree, and the implementation file for `foo` is in the tree in a compilation unit different than the ref file, but the header declaring `foo` (which is, for this discussion, the header which the ref in question includes to recognize `foo`), is not in the tree, then I was concerned that you should get a "Jump to definition" menu option because the definition does exist in the tree, but in fact since the header file lies outside the tree you don't get a context menu at all for that ref - its type isn't recognized by dxr.  I guess you could still argue that maybe you should, since the def is actually in the tree (at least maybe for functions if not methods), but I'm okay with not arguing that :-P

So you can get 0 results doing 'Jump to definition' this way (i.e. you can show the 'Jump to definition' option when you shouldn't, and I don't think there's any way around that), but I guess you're not really going to miss any definitions you should have found like I was thinking you could.

dxr/app.py
kleintom: > It's worth considering whether it's desirable to redirect the user to a found FILE-domain result

My first reaction when I considered this point was to not redirect a FILE-domain result.  On the other hand, I think Gijs makes a pretty good argument in https://bugzilla.mozilla.org/show_bug.cgi?id=1202603  At some point it was decided that it was a good idea to redirect to a FILE-domain result on direct search, presumably on the theory that you were searching on a path name.  If you explicitly specify that you're searching for a path name by using a `path:` filter, why should that change?  Also if you don't want the redirect you can just not hit return (though I have to say that behavior was confusing to me pretty much right up to the point where I started reading this code).

dxr/plugins/clang/menus.py
kleintom: > Do you see a way we could have fewer?

If you'd like to start using a 'bubble' param then I think we can at least change my new `from_q` back to the already existing `from`, since using a different bubble for `from_q` and `from` queries was the only reason I introduced the `from_q`.  I'm not sure about getting rid of the `no_from` though...I'll have to think more about that.

dxr/static_unhashed/js/dxr.js
kleintom: > I hate to lay the responsibility of fixing it on you, but I also hate to make it worse.

No problem, I'll work on making the changes.

kleintom:
I did end up removing _RefWithDefinition as a base of FunctionRef, and the upshot of my long discussion last night on `path is None` is: we need path as a bool to tell whether the ref in question is actually a def or if the def is at another location, but the path value doesn't provide any useful info on whether the actual def is in the tree or not, so we ignore it and just search for the def.

I did move the info bubble display stuff into the templates, but I'm not seeing any way to get rid of either the new `from_q` or `no_from`.  If anybody has any ideas though I'd be happy to change things up.  I added some doc on `from`, `from_q`, and `no_from` to `_search_json` since that's where all of those originate - let me know if that's now quite what you had in mind.

I'll work on some tests tomorrow.

dxr/plugins/clang/menus.py
```diff
@@ -198,33 +200,57 @@ def _more_menu_items(self, (qualname,)):
...
+                   (search_for_def, prop['qualname'],
+                    'has_overriddens' in prop, 'has_overrides' in prop),
+                   qualname=prop['qualname'])
+
+    def menu_items(self):
+        search_for_def, qualname = self.menu_data[:2]
+        if search_for_def:
+            menu = [{'html': "Jump to definition",
+                     'title': "Jump to definition",
+                     'href': url_for('.search', tree=self.tree.name,
```
pelmers: IIRC url_for is a big time sink to call repeatedly, we may want to hardcode this particular construction like search_url does.

dxr/plugins/clang/menus.py
```diff
@@ -198,33 +200,57 @@ def _more_menu_items(self, (qualname,)):
...
+                     'icon': 'jump'}]
+        else:
+            menu = []
+        menu.extend(self._more_menu_items())
+        return menu
 
-    def _more_menu_items(self, (qualname, has_overriddens, has_overrides)):
+    def _more_menu_items(self):
+        qualname, has_overriddens, has_overrides = self.menu_data[1:]
+        qualname = quote(qualname)
```
pelmers: 👍

dxr/testing.py
```diff
@@ -143,9 +143,34 @@ def search_response(self, query, is_case_sensitive=True):
...
+                         redirect='true',
+                         no_from='true'),
+            headers={'Accept': 'application/json'})
+        eq_(response.status_code, 200)
+        try:
+            location = json.loads(response.data)['redirect']
+        except KeyError:
+            self.fail("The query didn't return a direct result.")
+        if line:
+            ok_(location == '/code/source/%s#%s' % (path, line))
```
pelmers: We can use eq_ for these

dxr/templates/text_file.html
```diff
@@ -5,6 +5,22 @@
 
   {% set state_eof = True %}
 
+  {% block bubble %}
```
pelmers: I like having the bubble moved to the template

pelmers:
I'm a fan of the new behavior, but I find the distinction between from and from_q is too subtle: there should be just one parameter that has a query string.

Instead we could add a param such as "redirect_type={single, direct}" which reflects the intention more clearly.

Personally, I'd say we don't need to maintain the illusion of direct jump by hiding the bubble and remove no_from as well, since it's not very intrusive.

kleintom:
I rebased and added two new commits: one to replace `from` and `from_q` with a `redirect_type` parameter taking the value 'single' or 'direct', and one to remove the `no_from` parameter.  Thanks for the suggestions Peter!

Originally I was going to fight for `no_from`, but then I noticed something which I guess wasn't that remarkable to me back in March (when we had other similar issues): if you click on "Jump to definition", it loads a page with a redirect query (so that gets put on the history), and then the query runs and you get auto-redirected to the single result.  But then if you click back you go back to the redirect query page ... and the query runs and takes you back to the single result.  Of course you can get out of it by clicking back twice quickly, but instead I decided to just cut off any additional redirects after the initial load of the redirect query (on the initial load I remove the `redirect=true` parameter from the url so subsequent visits to that page don't redirect, but then run the query using the original redirect url).  Now when you go back from the def page you just load a query that returns one result without redirecting.  You still need to hit back again to get back to your original source page, but I don't see any way around that.  Which is why I'm now fine with being upfront about the fact that we're using a query :-)

[The only issue I can see with the `redirect=true` mucking on a page load is that if you type in a redirect query and don't get a redirect result, and then hit the reload button, the `redirect=true` parameter will be removed, but then it wasn't doing any good in that case anyway.  Am I missing anything else?]

pelmers:
Looks good to me!

kleintom:
I should have updated the format number with this, sorry.  I'll try another merge tomorrow if I can figure out how to remerge a reverted merge (there's apparently an issue, which I haven't tested yet, as a consequence of using `git revert -m` like I did - maybe just smash the merge commits together (since they're already in the history individually) and add a format bump?  That's what I'm thinking now; let me know if that's objectionable).

But should we even merge something that needs a format bump right now with all the failing builds?  It's still not clear to me if it's possible to just deploy the builds that are succeeding with a new format number if the others are expected to continue failing for a while?  @klibby?
In any case maybe I'll wait until the js plugin issue has cleared so that I'm not adding something else that might hold up updating m-c.  Thoughts?

PR #524, opened 2016-04-15
Cut off python name inheritance cycles.  Fixes bug 1264627.

kleintom:
The bug is caused by a class definition in
mozilla-central/testing/web-platform/tests/tools/py/py/_log/warning.py of
'class DeprecationWarning(DeprecationWarning):', where the base
DeprecationWarning is a built-in which is never explicitly imported.  We don't
distinguish between the new DeprecationWarning name and the built-in name, so
it looks to us like we have an inheritance cycle.  Similar bugs can occur
with longer cycles.

The fix here cuts those off so that we don't crash on indexing when we exceed
the allowed recursion depth.

### DIFF
--- a/dxr/plugins/python/analysis.py
+++ b/dxr/plugins/python/analysis.py
@@ -79,7 +79,7 @@ class TreeAnalysis(object):
         # Compute which functions override other functions.
         for class_name, functions in self.class_functions.iteritems():
             functions = set(functions)
-            base_classes = self.get_base_classes(class_name)
+            base_classes = self.get_base_classes(class_name, set([class_name]))
 
             # For each base class, find the union of functions within
             # the current class and functions in the base; those are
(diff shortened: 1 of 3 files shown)


kleintom:
Here's the class that's been causing the indexing failure:
https://hg.mozilla.org/mozilla-central/file/45661ca3b6b1/testing/web-platform/tests/tools/py/py/_log/warning.py#l3

dxr/plugins/python/analysis.py
```diff
@@ -100,26 +100,46 @@ class method).
             for overridden_function in overridden_functions:
                 self.overridden_functions[overridden_function].append(function)
 
-    def get_base_classes(self, absolute_class_name):
+    def get_base_classes(self, absolute_class_name, seen):
         """Return a list of all the classes that the given class
         inherits from in their canonical form.
 
+        :arg seen: The set of normalized base class names already known.  Python
+        doesn't permit actual inheritance cycles, but we don't currently
```
erikrose: Just indent the lines under ":arg seen:" for consistency with the Sphinx docs elsewhere in the project. (Also, do similarly below.)

dxr/plugins/python/tests/test_derived.py
```diff
@@ -47,3 +48,47 @@ def test_absolute_name(self):
             'class <b>Parent2</b>(Grandparent):',
             'class <b>Child</b>(Parent, Parent2):',
         ])
+
+
+class NameCycleTests(PythonSingleFileTestCase):
+    """In class inheritance processing, we currently don't distinguish between
```
erikrose: Great docs! Those will serve us well when we improve builtin handling someday.

dxr/plugins/python/tests/test_derived.py
```diff
@@ -1,4 +1,5 @@
 from textwrap import dedent
+import nose.tools
```
erikrose: You can go ahead and import more deeply here; that's what we do in most cases, except where names are ambiguous or undescriptive. (`from nose.tools import raises`)

dxr/plugins/python/tests/test_derived.py
```diff
@@ -47,3 +48,47 @@ def test_absolute_name(self):
...
+            'class <b>Oops</b>(UserWarning):'
+            ])
+
+    @nose.tools.raises(AssertionError)  # remove this line when fixed
+    def test_inheritance_name_cycle_lookup_looping(self):
+        """Make sure we don't find the wrong name (local vs. built-in) on
+        base: and derived: searches when we have an inheritance name cycle.
+
+        """
+        self.found_nothing('bases:main.Oops')
```
erikrose: Both great cases. Very tricky. :-)

erikrose:
Looks great. Just tweak the spelling stuff, then go ahead and merge.

PR #528, opened 2016-04-21
File operator

kleintom:
Add a 'file:' operator for matching on file names only. Fixes bug 1249325.


kleintom:
(There was some mention at one point of removing the 'ext:' operator once 'file:' was implemented:
https://github.com/erikrose/dxr/blob/query-parser/docs/queries.rst
I'm not sure we'd want to do that at this point (and I would do it in a different PR in any case), but bringing it up here just in case.)

dxr/plugins/core.py
```diff
@@ -308,6 +301,26 @@ def filter(self):
             raise BadTerm('Path globs need at least 3 literal characters in a row '
                           'for speed.')
```
pelmers: 2 blank lines

dxr/plugins/core.py
```diff
@@ -21,11 +22,11 @@
                             NoTrigrams, PythonRegexVisitor)
 from dxr.utils import glob_to_regex
 
-__all__ = ['mappings', 'analyzers', 'TextFilter', 'PathFilter', 'ExtFilter',
-           'RegexpFilter', 'IdFilter', 'RefFilter']
+__all__ = ['mappings', 'analyzers', 'TextFilter', 'PathFilter', 'FileFilter',
```
pelmers: Here it's FileFilter, but later it's defined FilenameFilter

dxr/plugins/core.py
```diff
@@ -308,6 +301,26 @@ def filter(self):
...
+                         '<code>?</code>, and <code>[...]</code> act '
+                         'as shell wildcards.')
+
+    @negatable
+    def filter(self):
+        glob = self._term['arg']
+        try:
+            return es_regex_filter(
+                regex_grammar.parse(glob_to_regex(glob)),
+                'file_name',
```
pelmers: Here's the only difference with PathFilter; would it be reasonable to refactor into subclasses?
Perhaps inherit of something like PathSegmentFilter.

dxr/es.py
```diff
@@ -13,6 +13,12 @@
 }
 
 
+UNANALYZED_STRING = {
```
pelmers: nice refactoring 🏆

dxr/plugins/core.py
```diff
@@ -2,15 +2,16 @@
 
 from base64 import b64encode
 from itertools import chain
-from os.path import relpath, splitext, islink, realpath
+from os.path import relpath, splitext, islink, realpath, basename
```
pelmers: while we touch this line, I note that the file never calls islink

dxr/plugins/core.py
kleintom: > Here it's FileFilter, but later it's defined FilenameFilter

Nice catch!

dxr/plugins/core.py
kleintom: I'm still on the fence... the `name` and `description` vars also differ, as does the `BadTerm` text.  Seems like you'd be factoring out basically just the call to `es_regex_filter` (with some parameter for the mapping name)...  I'm leaning toward saying it's not worth it - what do you think?

kleintom:
@pelmers I wound up factoring out a _PathSegmentFilterBase - can you take a look?

pelmers:
Looks great, r+

kleintom:
Thanks for the reviews Peter!

PR #529, opened 2016-04-26
Support clang 3.6

makotokato:
dxr doesn't support clang 3.6.  clang 3.6 have a few changes to use std::unique_ptr instead.

### DIFF
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -139,7 +139,11 @@ private:
   std::ostream *out;
   std::map<std::string, FileInfoPtr> relmap;
   LangOptions &features;
+#if CLANG_AT_LEAST(3, 6)
+  std::unique_ptr<DiagnosticConsumer> inner;
+#else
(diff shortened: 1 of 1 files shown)


dxr/plugins/clang/dxr-index.cpp
```diff
@@ -1241,9 +1249,15 @@ void PreprocThunk::InclusionDirective(  // same in 3.2 and 3.3
 // Our plugin entry point.
 class DXRIndexAction : public PluginASTAction {
 protected:
+#if CLANG_AT_LEAST(3, 6)
+  std::unique_ptr<ASTConsumer> CreateASTConsumer(CompilerInstance &CI,
+                                 llvm::StringRef f) override {
```
kleintom: Align this line with 'CompilerInstance' in the previous line please.

kleintom:
Thanks for the PR!  Looks good to me but I'm not on clang 3.6 yet, so can you confirm that `make test` passes with no errors?

If so and you can fix up that one little nit I'll go ahead and merge.

PR #530, opened 2016-04-27
Canonicalize clang function decls before pulling parameter names.

kleintom:
The current practice of using non-canonical decls leads to parameter names
depending on which ref to the function they're pulled from, which means
function qualnames for a given function vary across refs, and then search is
broken.

### DIFF
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -256,6 +256,9 @@ public:
     }
 
     if (const FunctionDecl *fd = dyn_cast<FunctionDecl>(&d)) {
+      // If you don't use the canonical decl you can get different param strings
+      // depending on which ref you're querying (which is bad).
+      fd = fd->getCanonicalDecl();
       // This is a function.  getQualifiedNameAsString will return a string
       // like "ANamespace::AFunction".  To this we append the list of parameters
       // so that we can distinguish correctly between
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -141,6 +141,7 @@ private:
   LangOptions &features;
   DiagnosticConsumer *inner;
   static std::string tmpdir;  // Place to save all the csv files to
+  PrintingPolicy printPolicy;
 
   const FileInfoPtr &getFileInfo(const std::string &filename) {
     std::map<std::string, FileInfoPtr>::iterator it;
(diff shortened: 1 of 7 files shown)


kleintom:
@pelmers Do you have some time to take a look at this one?

Both the first commit and the second commit fix all of the bugs blocked by bug 1268205 by normalizing qualnames for functions (and also, by extension, var-refs on local function variables).  The second commit further simplifies things using clang's PrintingPolicy to:
1) print 'bool' instead of '_Bool' - I think that one's a no-brainer.
2) drop type tags like 'class', 'struct', 'enum', etc. on types.  So, for example, instead of the current
`+callers:"js::Stringify(struct JSContext *, js::MutableHandleValue, class JSObject *, class JS::Value, class js::StringBuffer &, enum js::StringifyBehavior)"`
it will now be
`+callers:"js::Stringify(JSContext *, js::MutableHandleValue, JSObject *, JS::Value, js::StringBuffer &, js::StringifyBehavior)"`
That seems like a win to me - we never actually needed the tags for qualnames and this makes things shorter (so less ES space), closer to the actual function declaration, and easier to handcraft (if anyone actually does that).  I'm interested in what you and others think though.

You'll note that the tests didn't change along with change 2) - that's because I've never been able to get the tags to appear after I switched to using getCanonicalDecl.  They still appear with commit 1 only (not with commit 2) when I index m-c, so I'm guessing there's something about the m-c environment/compile options that's affecting the print options.  ?  I've also never been able to reproduce the "type 2" bug I mention in https://bugzilla.mozilla.org/show_bug.cgi?id=1268205 so there aren't any tests for that, though as I mentioned, both commits do fix it in local builds.

kleintom:
Head's up: I indexed again last night and using PrintingPolicy seems to have added 20 minutes to a full indexing run (yesterday's run was while I was working and doing other stuff, so I didn't pay attention to the time) ... All of that extra time should be in the build step, so I'll do another clean build-only run and see what I get.

kleintom:
Okay, the build-only run with PrintingPolicy took the same amount of time as a build-only run using current HEAD (which is also the same amount of time as a build-only run using commit one here), and I can't see any reason why indexing would be taking longer with these commits (if anything I would expect it to be faster with shorter strings/less I/O), so I'll stick with both commits for now and chalk last night's run up to some anomaly.  I'll do another clean run tonight though and if it's slow again I'll just revert commit 2 and we'll go with commit 1 for now (which I'll have to check for time as well ... Saturday night).

kleintom:
Yup, the time from last night on a full indexing was right where it should be.

pelmers:
Looks good! r+

(reading c++ still makes my head spin 🙃🙂🙃🙂)

kleintom:
:-D  And reading clang AST "documentation" makes my head hurt, but I think we're good.  Thanks for the reviews!

PR #531, opened 2016-04-29
Use binaryornot and chardet (as a backup) for text detection.

kleintom:
binaryornot is replacing our old simple is_text, and chardet is being used
as a backup when we think a file is text but we're unable to decode it using
the configured source_encoding.  It might be nice if we could just drop
source_encoding all together, but chardet is quite slow - running chardet over
every is_text file in mozilla-central leads to indexing times nearly twice
their current length.


kleintom:
This took about 6 minutes longer to index mozilla-central than current HEAD did, which I think is worth it.

pelmers:
Looks good to me, especially the commenting. r+

kleintom:
Merged in https://github.com/mozilla/dxr/commit/3453ad091a6d400d4d1b055c0a442becf8ec1550

requirements.txt
```diff
@@ -1,7 +1,13 @@
+# sha256: TIMmzrNyf0_3AR1_Ybe8oVzKGU4iiAE6c9XMr1bToYs
```
erikrose: We should also add the hash for the wheel (or sdist) here; there's a wheel and an sdist available, and which one is chosen is a matter of pip version and flags.

Btw, many applause on the work you've done while I'm away. Looks like DXR is in good hands!

PR #533, opened 2016-05-03
Add 'extmatch' plugin for matching files in a folder by extension.

kleintom:
Matched files get nav pane links to each other.
As a start we're just matching "header" files with "implementation" files, with
no configuration.


kleintom:
@pelmers feedback/r?

This felt like it wanted to be a configurable plugin to me, so I gave it its own plugin, but I'd prefer to wait until Erik's back to decide on configuration, so for now that's just hardcoded.

dxr/plugins/extmatch.py
```diff
@@ -0,0 +1,71 @@
...
+from collections import namedtuple
+
+from dxr.indexers import (FileToIndex as FileToIndexBase,
+                          TreeToIndex as TreeToIndexBase)
+from dxr.mime import icon
+from dxr.utils import browse_file_url
+
+# A tuple of extensions along with a title describing the type of extensions:
+_TitledExts = namedtuple('_TitledExts', ['exts', 'title'])
+
```
pelmers: whitespace nit: " Surround top-level function and class definitions with two blank lines. "

dxr/plugins/extmatch.py
```diff
@@ -0,0 +1,71 @@
...
+        matching extension, when such a file exists.
+        """
+        def dual_exts_for(ext):
+            if ext in self.ext_pairings[0].exts:
+                return self.ext_pairings[1]
+            if ext in self.ext_pairings[1].exts:
+                return self.ext_pairings[0]
+            return _TitledExts((), '')
+
+        def is_unignored(path):
```
pelmers: is there a better adjective than unignored that fits? this feels like a double-negative since False is "No it's not not-ignored"

dxr/plugins/extmatch.py
kleintom: > whitespace nit: " Surround top-level function and class definitions with two blank lines. "

Thanks for the reminder (again! :-)

dxr/plugins/extmatch.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        def dual_exts_for(ext):
+            if ext in self.ext_pairings[0].exts:
+                return self.ext_pairings[1]
+            if ext in self.ext_pairings[1].exts:
+                return self.ext_pairings[0]
+            return _TitledExts((), '')
+
+        def is_indexed(path):
+            if any(fnmatchcase(basename(path), e)
```
erikrose: We should expose to plugins a single utility method that answers the question "Is this ignored?" The logic is tricky enough (and likely enough to be revised) that we shouldn't repeat it.

dxr/plugins/extmatch.py
```diff
@@ -0,0 +1,73 @@
...
+from dxr.utils import browse_file_url
+
+# A tuple of extensions along with a title describing the type of extensions:
+_TitledExts = namedtuple('_TitledExts', ['exts', 'title'])
+
+
+class TreeToIndex(TreeToIndexBase):
+    def __init__(self, plugin_name, tree, vcs_cache):
+        super(TreeToIndex, self).__init__(plugin_name, tree, vcs_cache)
+        self.header_exts = _TitledExts(('.h', '.hpp', '.hxx'), 'Header')
```
erikrose: A nit: we should use lists rather than tuples for the `exts` value. In DXR, we've do far followed the convention that tuples are for anonymous structs and lists are for homogeneous sequences of things. Quite possibly we should use something more efficient instead, since we're doing a lot of linear scanning over them using `in`. But the data sizes are small, so we might not see a benefit.

kleintom: > Quite possibly we should use something more efficient instead, since we're doing a lot of linear scanning over them using in. But the data sizes are small, so we might not see a benefit.

This:
```python
impl_exts_list = ['.cpp', '.c', '.cc', '.cxx', '.mm']
h_exts_list = ['.h', '.hpp', '.hxx']

impl_exts_set = set(['.cpp', '.c', '.cc', '.cxx', '.mm'])
h_exts_set = set(['.h', '.hpp', '.hxx'])

for impl_exts, h_exts in [[impl_exts_list, h_exts_list],
                          [impl_exts_set, h_exts_set],
                          [impl_exts_list, h_exts_list],
                          [impl_exts_set, h_exts_set]]:
    
    start = timer()
    for dirpath, dirnames, filenames in os.walk('./mozilla-central'):
        for f in filenames:
            root, ext = splitext(f)
            if ext in impl_exts:
                exts = h_exts
            if ext in h_exts:
                exts = impl_exts
    print timer() - start
```
prints
```
4.68175315857
4.64918088913
4.69052195549
4.64750003815
```
(once the m-c listings are in memory), so in this case it indeed looks like it doesn't make much difference.  The other issue with using `set` in this case is that when we iterate over the matched (dual) extensions, we want to do the `is_file` check in the order the extensions were listed since they're in most-frequent-extension-first order, which (theoretically) need not happen if we iterate sets instead.  Or did you have another container in mind?  If not I think I'll just switch to lists.

(Thanks for bringing this up in any case :-)

dxr/plugins/extmatch.py
kleintom: Sounds good :-)

dxr/plugins/extmatch.py
erikrose: Benchmarks always win. :-) Lists should be fine.

PR #534, opened 2016-05-05
Binary browse and image rev

kleintom:
* Allow browsing to binary files.
* Display the nav pane for binary images and support viewing binary images on /rev/.
* Support viewing of textual images on /rev/.


kleintom:
(Still needs a little more work :-)

dxr/app.py
```diff
@@ -249,6 +251,23 @@ def raw(tree, path):
     return send_file(data_file, mimetype=guess_type(path)[0])
 
 
+@dxr_blueprint.route('/<tree>/rawrev/<revision>/<path:path>')
```
erikrose: I'm not in love with the user-visible name "rawrev", since nowhere else do we mash two words up against each other without any divider.

kleintom:
@pelmers Do you have some time to look this one over?

dxr/app.py
kleintom: raw_rev?

erikrose: So far we've managed to keep underscores out of the URL, which is a nice thing, I think (esp. when we're already using dashes). I haven't had a chance to look through your PR in detail yet, but I vaguely and nebulously wonder if we really need this extra endpoint. I suppose it's to distinguish a binary's containing page from its raw data?

kleintom: I have to admit I was unaware of the dashes-vs-underscores debate, but even had I been, we do use dashes in filters as you point out, so dashes are more consistent - thanks!

Yes, the new endpoint is to serve up just the plain raw binary of a file at a particular revision rather than embedded in the dxr layout.  Then the /rev/ page for a binary image can have a /raw-rev/ img href to the image at that rev (so the rev image gets displayed on its /rev/ page), and for svg /rev/ pages we can put a "view image" link in the nav pane to the svg served up as svg by a /raw-rev/.

kleintom:
Updated to use 'raw-rev'.

pelmers:
Looks and works good to me, r+

PR #535, opened 2016-05-06
Expose pieces of build.unignored so that plugins can do their own ver…

kleintom:
…sion.


kleintom:
I guess `unignored_files` and `unignored_folders` would be useful if a plugin wanted to do some whole-tree processing that involved working a folder at a time, and `path_is_indexed` (or `file_is_indexed`) would be more useful at the FileToIndex level?  Is that what you had in mind Erik?

PR #536, opened 2016-05-07
Correctly set 'limit' in JS doQuery when a queryString exists.

kleintom:
* On an initial (i.e. non-append) search-results load we were setting
  previousDataLimit to the computed 'limit' var, but the actual limit (explicit
  or implicit) in the case of a pre-existing queryString passed into doQuery can
  be larger than 'limit', in which case we wind up rerequesting results starting
  from the smaller 'limit' when we infinite scroll.  This was only an issue when
  loading a url with an existing query in it and enough search results to
  require infinite scroll.

* Also don't update global search result variables like 'previousDataLimit' and
  'dataOffset' when we receive old outdated results that we ignore.

* Stop infinite scroll when a query returns no results:
If previousDataLimit == resultsLineCount (e.g. there are 100 total results and
we use the default limit of 100) then we used to get into endless infinite
scroll queries returning 0 results everytime we scrolled.  Now we stop
infinite scroll when a query returns 0 results.

### DIFF
--- a/dxr/static_unhashed/js/dxr.js
+++ b/dxr/static_unhashed/js/dxr.js
@@ -343,14 +343,22 @@ $(function() {
      */
     function doQuery(redirect, queryString, appendResults) {
         query = $.trim(queryField.val());
-        var myRequestNumber = nextRequestNumber,
-            lineHeight = parseInt(contentContainer.css('line-height'), 10),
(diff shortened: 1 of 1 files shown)


kleintom:
@pelmers @erikrose  Please holler if you'd like more time to look it over, but if not I'll plan on landing this one tomorrow.

PR #538, opened 2016-05-09
Fix permalinks when the indexed tree uses Perforce.

abbeyj:


pelmers:
Thanks for the patch! Though I can't test it, I don't see any issues. r+ !

PR #539, opened 2016-05-09
Remove case sensitivity checkbox. (take 3)

pelmers:
The previous PR, #482, attempted to unify the symbols '-', '@', and '+' in the query language under a "modifiers" category, but I was not able to do so cleanly because then a query such as "--+" would no longer parse.

Here I bring back the original, which uses smart case, and '@' to force case-sensitive per term.

One suggestion was to show a bubble when the behavior happens, but to do so, the server must tell the page's javascript, and it seems unnecessary to pass that around on every request when it's only useful the first time.

r? @erikrose


pelmers:
[bug 1184747](https://bugzilla.mozilla.org/show_bug.cgi?id=1184747)

tests/test_query_parser.py
```diff
@@ -27,7 +27,7 @@ def test_overall(self):
     def test_tricksy_orphanses(self):
```
erikrose: What has it got in its grammarses? *hisssss*

erikrose:
This looks great and really simplifies the codebase. I was going to argue that we need a bubble, but this is such a good cleanup and implements a requested feature; let's just wait until the confusion tickets start rolling in and address whichever way those point. However, we can't implement a completely undocumented, potentially confusing feature. Can we squirrel away an "@" example and a "Hey, mixed-case ones are case-sensitive" one in the Operators menu examples, even if it's in an existing common filter like `id` or something?

pelmers:
Thanks for the review! I've added a note to the [IdFilter description](https://github.com/mozilla/dxr/pull/539/files#diff-27050266ad848025893af5a3c2ffd185R417)

dxr/plugins/core.py
```diff
@@ -414,9 +414,9 @@ class IdFilter(FilterAggregator):
     name = 'id'
     domain = LINE
     description = Markup('Definition of an identifier: '
-                         '<code>id:someFunction</code> <code>id:SomeClass</code> '
-                         '<code>id:@function</code> '
-                         '(@ forces case-sensitive search, otherwise automatic)')
+                         '<code>id:SomeClass</code> <code>id:@foofunction</code> '
+                         '(@ forces case-sensitive search, '
```
erikrose: Let's stretch it out and be more explicit, since we're already hiding this pretty sneakily: (<code>@</code> forces case-sensitive search. Otherwise, case-sensitive is assumed if there are capital letters in the query.)

dxr/plugins/core.py
```diff
@@ -414,7 +414,9 @@ class IdFilter(FilterAggregator):
     name = 'id'
     domain = LINE
     description = Markup('Definition of an identifier: '
-                         '<code>id:someFunction</code> <code>id:SomeClass</code>')
+                         '<code>id:SomeClass</code> <code>id:@foofunction</code> '
+                         '(@ forces case-sensitive search, '
```
erikrose: Let's stretch it out and be more explicit, since we're already hiding this pretty sneakily: (<code>@</code> forces case-sensitive search. Otherwise, case-sensitive is assumed if there are capital letters in the query.)

erikrose:
So, if you don't object, go ahead and make that one wording tweak to the `id` description, and then Merge This Puppy™!

PR #540, opened 2016-05-10
Use the original template pattern for function template qualnames.

kleintom:
Use the original template pattern rather than the concrete types of an
instantiated template.

This gives all non-specialized templated function decls the same qualname and
thereby fixes search for such functions.

I haven't been able to get clang to normalize template specializations in all
cases, so for now each specialization of a function template gets its own
qualname.


kleintom:
feedback?  I'll wait to land PR #522 until we make a decision on this one since without something like this, doing a '+function:' call on the call site of a templated function is going to return 0 results (whereas right now 'Jump to definition' does work on such functions).

As I mention above, I don't understand all template cases the way I'd hoped yet, but I think there's a fair amount of work left to do to fully support templates in any case, and I think this is a good first step. 

Pros: searching on templated functions is currently pretty broken - this gets it working entirely for templated functions/methods that don't have any specializations.  For those that do have specializations, search works within a given specialization, i.e. you can find callers of a particular specialization (finding callers of the original templated function finds all those calllers that aren't callers of a specialization).  (If/)When we switch to doing "Jump to definition" via a '+function:' search, this will cause us to jump to the definition of the specialization of the particular decl the query was performed at, which I think is what we would want (and what currently happens).

Cons: If you want to find all the callers of a templated function, you need to do 'Find callers' on the original function *and* on each of its specializations.  Also, I don't think there's an easy way to find all of the specializations of a given templated function (though clang may support adding that at some later point).  [One more note: if all specializations did use the same qualname then a '+function:' search for that qualname would turn up multiple results, one for each specialization, which could be considered good or bad I guess].

pelmers:
This patch looks solid to me. But I should bring @erikrose opinion on this too.

kleintom:
Thanks for taking a look Peter.  I should have added that on a local build of a slightly old m-c, this seems to fix the queries reported in bugs 1270246 and 1220865 (I didn't attempt to find all the callers by hand and then verify that the search returns them all, but it does now return results where it used to return none).

erikrose:
FWIW, returning multiple results from a fully-qualified search might surprise certain parts of the program; I think there might be spots that assume a single one.

erikrose:
It looks like you've thought through all the pros and cons and mapped out a sane direction for future improvements. Your tests look thorough and well-documented. Merge that puppy in.

And keep in mind that, if we want to more easily find all the callers of a templated function, we could introduce a new filter type—that's always an option.

PR #541, opened 2016-05-10
Path promotion, take 2

pelmers:
Replaces #443 

I update the previous work on path promotion to the latest head, passing all tests. I remove the path highlighting code because @kleintom has a much better one in #519.
I draw a box on the right side of search results that holds up to 5 path results for single unfiltered term searches.
Remark: the switch tree selector then shows up to the left of this box, rather than the far right of the screen. When the box does not appear then it's at the far right. (well, it's possible 'absolute' the thing to the right always...)


pelmers:
I've updated the merge to the commits of today. One other thing I do here is allow folders to appear as results. If we don't want it, that would be a simple toggle in query.py (the [previous comment](https://github.com/mozilla/dxr/blob/master/dxr/query.py#L137-L138) disabled it because the javascript would not render properly).
Thoughts? @erikrose @kleintom

kleintom:
Seems like `nunjucks/breadcrumbs.html` is missing?

pelmers:
derp

erikrose:
When you do a search, the caption "2 results from the ____ tree:" is missing the "____", which is supposed to be the tree name and a link back to the directory listing.

erikrose:
If there are only path results (and no normal ones), there is no way to navigate directly back to the directory listing.

erikrose:
Sometimes there are funny spaces between path results: screen shot 2016-05-16 at 18 00 05. That one is durable against window resizes and text-only zooms. Let me know if you have trouble reproducing it.

erikrose:
I think the layout concept is sound but the CSS and/or markup needs work. Here's another weird way the path box goes goofy, expanding in an attempt to accomodate a long name and then failing and wrapping anyway: 
screen shot 2016-05-16 at 18 03 07

Also, I think it would be better to float the path box over the results rather than forcing even results well below it to avoid the space below it.

kleintom:
In a coincidence 6 days in the making, I was writing as Erik was commenting, but I'll leave it for a slightly different perspective:

It feels like an error to me when it says there are, say, `20 path results:` and then lists five.  Could we add a `...View all results` link or vertical dots or something just to give a visual impression that there are more than are being shown?  [Might it also be helpful at some point to return the path results in the same weighted order when you click on the `20 path results` link?  Right now you kind of start all over when you go to the full results.]

It seems like the link to the tree in the `n results from...` went away?  I find that I use that link, so I would be in favor of bringing it back.  Also could we maybe say something along the lines of `n text results from the blah tree` (or something) to distinguish further from path results (maybe not necessary)?  Then I think it would be nice to bring back the notification when there are 0 file results, both when there are path results but no file results, and particularly when there are no results of either kind (now you just get a blank page).

I'd prefer to see the path results below the tree selector box...

I'm getting two tree selector boxes when I browse to an empty folder, which doesn't seem to be happening on master.

pelmers:
@erikrose 

> When you do a search, the caption "2 results from the ____ tree:" is missing the "____", which is supposed to be the tree name and a link back to the directory listing.

Good catch! I think I lost that in the merge...

> Also, I think it would be better to float the path box over the results rather than forcing even results well below it to avoid the space below it.

The paths container is currently set to float: right, but each path result gets its own table. Since all the rows in a table have the same width, they all must stop at the path box. However any table that starts below the paths box can run all the way across the page.
screen shot 2016-05-16 at 4 58 15 pm

> Sometimes there are funny spaces between path results:

Actually I haven't seen this one before, can you tell me how to make it happen?

pelmers:
@kleintom

> It feels like an error to me when it says there are, say, 20 path results: and then lists five.

Good point, how do you feel about "View all [20 path results]()" rather than just "20 path results"?

> It seems like the link to the tree in the n results from... went away?

Thanks for catching this! I lost it in the merge.

> I'd prefer to see the path results below the tree selector box...

I'll test how absolute'ing the menu to the right will look (the drop down was already absolute right: 0).

> Then I think it would be nice to bring back the notification when there are 0 file results, both when there are path results but no file results, and particularly when there are no results of either kind (now you just get a blank page).

Agreed, it does feel strange to see the blank page -- I'll return the 0 results messages.

> I'm getting two tree selector boxes when I browse to an empty folder, which doesn't seem to be happening on master.

I'll figure this one out too!

kleintom:
> how do you feel about "View all 20 path results" rather than just "20 path results"?

I think it would feel out of place if it came before the first five results instead of after.  Maybe add a "Top results:" (or so) line above the results when there are more than 5 results?

As long as I'm here: I can reproduce Erik's funny-spaces-between-results by searching for 'fold' in test_path_file_filters.

pelmers:
Thanks for all the comments! I made a commit that responds to them, summarized in [commit message](https://github.com/mozilla/dxr/pull/541/commits/d5873da2ef84bfae39c53c3339ce8f8e5edc9244).

I wasn't able to reproduce the funny spaces result... (searched for 'fold')
screen shot 2016-05-17 at 9 32 12 am

kleintom:
Thanks for bringing back the tree link! :-)

With the new layout I need a longer name to create the extra space issue:
`touch fishfishfishfishfishfishfishfishfishfishfishfishfish`
fish_results
One issue with having the tree selector inside the results is that if we're only aware of one tree (at least I think that's when it happens) then there's no tree selector box and the path results box looks like it's wide for no reason.

I'm personally not a fan of the `View all n path results:` change; it's trying to serve double duty as a heading for the five results being shown (maybe just getting rid of the colon would help with that) and also telling how many total results there are and providing a link to the full list of results, and for me at least there's just a little dissonance there.  But (of course) don't let me stop you from landing this if you and Erik agree on something!  :-)

pelmers:
@kleintom Do you still see the spacing bug? I could not make it anymore once I changed from display:inline to display:inline-block (tbh I don't even know exactly the difference 🤔)

kleintom:
Looks like that fixed it :-)

I like the "Showing first few of" change, thanks!, but could we only show that when there are more results than are being displayed?

I guess having the tree selector inside the path results box works, but it occurred to me that maybe we just had a miscommunication (sorry if that's the case): I had been thinking of "below" in the sense of the way the nav pane is below the tree selector (lower on the page) when the page loads, not in the sense of underneath.  Just throwing that out there in case you prefer it.

Nice touch including FILE operators in the path results :D  It looks like there's one small issue though: if you include a `-file:` negation, for example, it's referenced in the `n path results` link without the `-`.  For a concrete example, search for `-file:fish folder` in the fish tests index, which returns 4 path results, and then click on the `4 path results` link in the path results box, which takes you to the 3 results for `file:fish path:folder`.

pelmers:
@kleintom thanks for all the comments! I made some layout changes that should look more like what you're thinking, and added another case for the header wording.
Also I changed the file operator __str__ to reflect negation or qualification (and will have to do case sensitivity now too, once I merge back to master).

dxr/app.py
```diff
@@ -168,26 +175,42 @@ def _search_json(query, tree, query_text, is_case_sensitive, offset, limit, conf
...
-                    'lines': [{'line_number': nb, 'line': l} for nb, l in lines]}
-                   for icon, path, lines in count_and_results['results']]
+        # Pull up all the non-negated text terms.
+        single_term = query.single_text_term()
+        count, results = query.results(offset, limit)
+        promoted_count, promoted, promoted_query = 0, [], ""
+        if offset == 0 and single_term:
+            # Now we pull out the single line term and pass into promoted paths.
+            try:
+                has_promoted_results = True
```
kleintom: Move to the end of the block.

dxr/plugins/core.py
```diff
@@ -51,13 +66,13 @@
         },
         'properties': {
             # FILE filters query this. It supports globbing via JS regex script.
```
kleintom: Delete this (or move it).

dxr/static_unhashed/css/icons.css
```diff
@@ -100,6 +100,12 @@ a.type {
 /* End of Panel only icons */
 
 /* MimeType Icons */
+.icon-container.folder {
+    background: transparent url('/static/icons/mimetypes/folder.png');
+}
+.promoted-icon.folder {
+    background: transparent url('/static/icons/mimetypes/folder.png');
+}
```
kleintom: Combine these two, and move down to keep alphabetized.

dxr/static_unhashed/css/tree-selector.css
```diff
@@ -9,14 +9,13 @@
 
 .tree-selector {
     display: inline-block;
-    position: relative;
     float: right;
     width: auto;
+    text-align: right;
```
kleintom: I'm wondering if d.m.o's tree list is going to be harder to read with this - is it for a particular purpose?

dxr/templates/nunjucks/breadcrumbs.html
```diff
@@ -0,0 +1,11 @@
+{% macro breadcrumbs(www_root, paths_and_names) %}
+  <div class="breadcrumbs">
+    {%- for path, name in paths_and_names -%}
+      {%- if not loop.first or loop.last and loop.length != 1 -%}
```
kleintom: I can never remember operator precedence order, so this is hard for me to read - can we add parentheses?

dxr/testing.py
```diff
@@ -206,7 +206,8 @@ def search_results(self, query, is_case_sensitive=True):
         """
         response = self.search_response(query,
                                         is_case_sensitive=is_case_sensitive)
-        return json.loads(response.data)['results']
+        data = json.loads(response.data)
+        return data['promoted'] + data['results']
```
kleintom: I'm surprised this didn't cause more trouble.  I could see us wanting to add some functions to just return one or the other at some point, but I guess we can cross that bridge when we come to it.

tests/test_path_promotion/code/file
```diff
@@ -0,0 +1 @@
+Top
```
kleintom: Do you use this file or the `top` file?

dxr/templates/nunjucks/promoted.html
```diff
@@ -0,0 +1,43 @@
+{% from "result_lines.html" import result_lines -%}
+
+{% macro promoted_list(query, promoted, promoted_count, www_root, tree) -%}
+  {% for result in promoted %}
+    <div class="promoted-row {% if result.is_binary %} binary_row {% endif %}">
+      <div class="promoted-icon {{ result.iconClass }}"></div>
+      <div class="promoted-result">
+        <span class="promoted-path">{{ result.pathLine | safe }}</span>
+        {% for entry in result.lines %}
```
kleintom: Do we ever have lines here?

dxr/app.py
```diff
@@ -28,9 +24,11 @@
 from dxr.plugins import plugins_named
 from dxr.query import Query, filter_menu_items
 from dxr.utils import (non_negative_int, decode_es_datetime, DXR_BLUEPRINT,
-                       format_number, append_update, append_by_line, cumulative_sum)
+                       format_number, append_by_line, cumulative_sum)
 from dxr.vcs import file_contents_at_rev
 
+
+
```
kleintom: This makes three blank lines.

kleintom:
Thanks for the updates - yup, that's what I was thinking :-)  For what it's worth, it looks good to me, but I won't be offended if Erik wants to take another peek as well - hopefully he likes the changes :-)

tests/test_path_promotion/test_path_promotion.py
```diff
@@ -0,0 +1,32 @@
+from nose.tools import eq_
+
+from dxr.testing import DxrInstanceTestCase
+
+
+class PathPromotionTests(DxrInstanceTestCase):
+    """Tests for functionality that isn't specific to particular filters"""
```
kleintom: I almost forgot this one :-)

dxr/templates/nunjucks/breadcrumbs.html
```diff
@@ -0,0 +1,11 @@
+{% macro breadcrumbs(www_root, paths_and_names) %}
+  <div class="breadcrumbs">
+    {%- for path, name in paths_and_names -%}
+      {%- if (not loop.first) or (loop.last and loop.length != 1) -%}
```
kleintom: Sorry for holding you up on something you inherited, but...
1) This is equivalent to `if not loop.first` (since you only get to the `loop.last and loop.length != 1` check if it's `loop.first`, but in that case it's never the case that you're also `loop.last` if `loop.length != 1`).
2) This not/and/or version is the one that came from folder.html - the one that came from file.html was
 `if not loop.first or loop.last`,
 i.e. `if (not loop.first) or loop.last`,
 which *would* add a leading slash if there was only one path term (because the one and only term would be `loop.last`), but we always lead file paths with the tree name (right?), so in our case `loop.last` always implies `not loop.first` for a file (I'm guessing maybe that wasn't always the case?).  So the original file.html version is actually equivalent to `if not loop.first` as well (for us).

So I'm suggesting just making it `if not loop.first`.  :-)

dxr/static_unhashed/css/icons.css
```diff
@@ -137,6 +137,9 @@ a.type {
 .html {
     background-image: url('/static/icons/mimetypes/html.png');
 }
+.icon-container.folder, .promoted-icon.folder {
+    background: transparent url('/static/icons/mimetypes/folder.png');
```
kleintom: Nit: I think the rest of these are alphabetized by the image name/class name, `folder` in this case.

dxr/static_unhashed/css/dxr.css
```diff
@@ -191,12 +188,48 @@ mark {
...
+    width: 16px;
+    height: 16px;
+    display: block;
+    left: 0px;
+    position: absolute;
+}
+.promoted-result {
+    display: inline;
+}
+.promoted-line.left-column {
```
kleintom: I don't think this or .promoted-text are used now.

dxr/templates/nunjucks/breadcrumbs.html
pelmers: Hah, good explanation! I guess I didn't really think about simplifying the expression, I just looked up the precedence rules and unpacked them...

dxr/static_unhashed/css/dxr.css
pelmers: Good point, thanks!

dxr/templates/nunjucks/breadcrumbs.html
kleintom: > I guess I didn't really think about simplifying the expression, I just looked up the precedence rules and unpacked them...

Yeah, I did the same thing the first time around :-)

Thanks for all the updates!  Not sure if you're still planning on adding `@` processing to `Filter.__str__`?

pelmers:
Thanks for pointing out the @ in str(Filter) too, I added that and put in some test cases to protect it.

dxr/filters.py
```diff
@@ -71,6 +72,26 @@ def __init__(self, term, enabled_plugins):
...
+        infix = ""
+        if 'not' in self._term and self._term['not']:
+            prefix += "-"
+        if 'qualified' in self._term and self._term['qualified']:
+            prefix += "+"
+        if 'case_sensitive' in self._term and self._term['case_sensitive']:
+            infix = "@"
+        # If the term has spaces, then surround with double quotes and escape internal quotes.
+        if ' ' in self._term['arg']:
+            return '%s%s:"%s"' % (prefix, self.name, self._term['arg'].replace('"', r'\"'))
```
kleintom: Insert `infix` here as well?

pelmers: And here I was thinking quotation forced case sensitive. fixed...

kleintom:
Looks good to me!

pelmers:
@erikrose Do you spot any other issues?

erikrose:
Waiting for an updated merge in from master to make the review easier.

erikrose:
Here's some nicer padding for the promoted results window:
```css
.promoted-container {
    border: 1px solid #a0a0a0;
    border-radius: 4px;
    margin-bottom: 1em;
    padding: 0.05em 1.0em 0.8em;
}
```

dxr/templates/nunjucks/promoted.html
```diff
@@ -0,0 +1,27 @@
...
+        {{ result.pathLine | safe }}
+      </div>
+    </div>
+  {% endfor %}
+{%- endmacro %}
+
+{%- macro promoted_header(promoted_query, promoted_count, promoted_count_formatted, top_of_tree, tree) -%}
+  <p class="promoted-see-more">
+    {% if promoted_count == 1 %}
+      Found 1 path result:
```
erikrose: Let's get the relationship straight between this caption and the main "2 results from the code tree:" one:

1. Right now, there's this weird interleaving where it says "2 results:", then there's the promoted box, then finally the results show up. 
screen shot 2016-07-22 at 16 34 42
 This used to make sense when the promoted stuff was on the right.
2. We might as well homogenize the wording, changing this to "1 path result". There's no need to say "Found."

dxr/templates/nunjucks/promoted.html
```diff
@@ -0,0 +1,27 @@
...
+    {% elif promoted_count <= 5 %}
+      Showing all of <a href="{{ promoted_query }}">{{ promoted_count_formatted }} path results</a>:
```
erikrose: I think we can just say "2 path results" here, and I don't think we need the link. I can't think of a common case in which someone would need the query link when there are very few results, can you?

dxr/templates/nunjucks/promoted.html
```diff
@@ -0,0 +1,27 @@
...
+    {% else %}
+      Showing first few of <a href="{{ promoted_query }}">{{ promoted_count_formatted }} path results</a>:
```
erikrose: I can't remember what we were thinking before, but we might as well do the conventional thing here and say...

```
Path results:
* blah
* blah
* blah

___37 more...___
```
Can you think of a reason not to?

dxr/app.py
```diff
@@ -172,38 +182,51 @@ def _search_json(query, tree, query_text, offset, limit, config):
...
-        # Convert to dicts for ease of manipulation in JS:
-        results = [{'icon': icon,
-                    'path': file_path,
-                    'lines': [{'line_number': nb, 'line': l} for nb, l in lines]}
-                   for icon, file_path, lines in count_and_results['results']]
+        count, results = query.results(offset, limit)
     except BadTerm as exc:
         return jsonify({'error_html': exc.reason, 'error_level': 'warning'}), 400
+    # If we're asked to redirect and there's a single result, redirect to the result.
+    if (request.values.get('redirect') == 'true' and count == 1):
```
erikrose: excess parens

dxr/app.py
```diff
@@ -172,38 +182,51 @@ def _search_json(query, tree, query_text, offset, limit, config):
...
+            'tree': tree,
+            'path': path,
+            'q': query_text,
+            'redirect_type': 'single'
+        }
+        return jsonify({'redirect': url_for('.browse', _anchor=line, **params)})
+
+    # Pull up all the non-negated text terms.
+    single_term = query.single_text_term()
+    promoted_count, promoted, promoted_query = 0, [], ""
```
erikrose: single quotes

dxr/app.py
```diff
@@ -172,38 +182,51 @@ def _search_json(query, tree, query_text, offset, limit, config):
...
+    if offset == 0 and single_term:
+        # Now we pull out the single line term and pass into promoted paths.
```
erikrose: "single text term", you mean?

dxr/query.py
```diff
@@ -38,19 +69,24 @@ def __init__(self, es_search, querystr, enabled_plugins):
...
-        If there is more than one term in the query or if the single term is a
-        non-textual one, return None.
+        If there is not exactly one such term, return None.
 
         """
-        if len(self.terms) == 1:
-            term = self.terms[0]
-            if term['name'] == 'text' and not term['not']:
-                return term
+        text_terms = [term for term in self._terms if term['name'] == 'text' and not term['not']]
```
erikrose: Can you remind me of some use cases for returning path results when there is more than one term (but only one text term, of course)?

dxr/query.py
```diff
@@ -29,6 +30,36 @@ def direct_searchers(plugins):
     return [searcher for searcher, _ in sortables]
 
 
+def _make_es_query(filters, is_line_query):
+    """Construct an ES query from nested list of filters by applying OR to each
```
erikrose: "from a"

dxr/query.py
```diff
@@ -126,52 +161,39 @@ def group_filters_by_name(predicate):
...
+    def promoted_paths(self, term, promote_limit=5):
+        """Return a tuple (total number of path results, promoted paths, querystring),
+        where the number of promoted paths is no more than promote_limit."""
+
+        # Filter out the FILE domain filters, and add our own path filter at the end.
+        file_filters = [[f for f in named_filters if f.domain == FILE] for named_filters in
+                        self._filters] + [[PathFilter(term, self.enabled_plugins)]]
+        filtered_query = _make_es_query(file_filters, False)
+        path_query = {'constant_score': {'query': filtered_query, 'boost': 0.5}}
+        # We add a second query that boosts exact path segment matches.
```
erikrose: Whizzy! :-D

dxr/query.py
```diff
@@ -126,52 +161,39 @@ def group_filters_by_name(predicate):
...
+
+    def promoted_paths(self, term, promote_limit=5):
+        """Return a tuple (total number of path results, promoted paths, querystring),
+        where the number of promoted paths is no more than promote_limit."""
+
+        # Filter out the FILE domain filters, and add our own path filter at the end.
+        file_filters = [[f for f in named_filters if f.domain == FILE] for named_filters in
+                        self._filters] + [[PathFilter(term, self.enabled_plugins)]]
+        filtered_query = _make_es_query(file_filters, False)
+        path_query = {'constant_score': {'query': filtered_query, 'boost': 0.5}}
```
erikrose: It would have helped me if the names of `path_query` and `exact_query` contrasted with each other, like maybe `segment_query` and `substring_query` or something.

dxr/query.py
```diff
@@ -209,6 +231,43 @@ def direct_result(self):
                 elif len(results) > 1:
                     return None
 
+    def _do_search_and_highlight(self, query, filters, is_line_query, offset, limit,
```
erikrose: What additional meaning does "do" convey?

dxr/query.py
```diff
@@ -209,6 +231,43 @@ def direct_result(self):
...
+                                 alphabetize=True):
+        """Perform a query using self.es_search and highlight the results,
+        returning a tuple (total count of results, list of results)
+
+        :arg query: valid ES query
+        :arg filters: nested list of filters
+        :arg is_line_query: whether to query lines (otherwise query files)
+        :arg offset: return results starting from this number
+        :arg limit: maximum number of results to return
+        :arg alphabetize: whether to alphabetize the results alphabetically by path and
```
erikrose: Everywhere else in the project, I've been naming boolean kwargs with a leading "should" or "is", like is_line_query above. Granted, it's a bit long here, but still.

dxr/query.py
```diff
@@ -126,52 +161,39 @@ def group_filters_by_name(predicate):
...
-        return {'result_count': result_count,
-                'results': self._line_query_results(filters, results, path_highlighters)
-                           if is_line_query
-                           else self._file_query_results(results, path_highlighters)}
-
-        # Test: If var-ref (or any structural query) returns 2 refs on one line, they should both get highlit.
+        }
+        # We ask not to sort by path and line, to use score ranking instead.
+        result_count, results = self._do_search_and_highlight(query, file_filters, False, 0, promote_limit, False)
+        return result_count, results, ' '.join((str(f) for f in chain.from_iterable(file_filters)))
```
erikrose: str(), not unicode()? What if somebody searches for something unicodey? If I type "readmé" in the search field, it crashes. (Add a test, too.)

dxr/filters.py
```diff
@@ -244,3 +266,14 @@ def _should_be_highlit(self, entity):
         return ((not self._term['qualified'] and
                  super(QualifiedNameFilterBase, self)._should_be_highlit(entity))
                 or is_in(self._term['arg'], entity['qualname']))
+
+
+def some_filters(plugins, condition):
```
erikrose: Good move.

dxr/static_unhashed/css/dxr.css
```diff
@@ -206,6 +218,16 @@ code a {
     text-align: right;
     color: #555;
     width: 6.5ch;
+    vertical-align: top;
+}
+.results .right-column {
```
erikrose: These right-column styles are dead now.

dxr/static_unhashed/css/dxr.css
```diff
@@ -215,7 +237,11 @@ code a {
     cursor: context-menu;
 }
 .results code {
-    white-space: pre;
+    white-space: pre-wrap;
```
erikrose: Make sure the rest of this stuff is still needed as well.

dxr/static_unhashed/js/dxr.js
```diff
@@ -380,29 +380,29 @@ $(function() {
...
             if (!append) {
                 contentContainer
                     .empty()
                     .append(nunjucks.render('results_container.html', data));
             }
         } else {
-            var results = data.results;
-            resultsLineCount = countLines(results);
-
-            for (var result in results) {
```
erikrose: Thank you for finding and fixing this bug. :-)

dxr/templates/file.html
```diff
@@ -13,18 +14,11 @@
...
-  <div class="breadcrumbs">
-    {%- for path, name in paths_and_names -%}
-      {%- if not loop.first or loop.last -%}
-        <span class="path-separator">/</span><a href="{{ www_root }}{{ path }}">{{ name }}</a>
-      {%- else -%}
-        <a href="{{ www_root }}{{ path }}">{{ name }}</a>
-      {%- endif -%}
-    {%- endfor -%}
+  <div class="breadcrumbs-container">
+    {{ breadcrumbs(www_root, paths_and_names) }}
```
erikrose: Soooo good to have this factored out!

dxr/templates/nunjucks/promoted.html
```diff
@@ -0,0 +1,27 @@
+{% macro promoted_list(query, promoted, promoted_count, www_root, tree) -%}
+  {% for result in promoted %}
+    <div class="result-head">
+      <div class="left-column">
+        <div class="icon-container {{ result.iconClass }}"></div>
+      </div>
+      <div>
+        {{ result.pathLine | safe }}
```
erikrose: |safe scares the heck out of me; it's been the source of a lot of security holes and is basically against Moz's code guidelines now. Can we instead wrap this in jinja.Markup() at some point in the Python code?

dxr/templates/nunjucks/results_container.html
```diff
@@ -1,10 +1,26 @@
...
 
-{{ results_header(result_count, result_count_formatted, top_of_tree, tree_tuples, tree) }}
+<div class="results-row">
 
-{% if results|length > 0 %}
-  <div class="results">
-    {{ results_list(results, www_root, tree) }}
+  <div class="result_header">
+  {{ results_header(result_count, result_count_formatted, top_of_tree, tree) }}
+  {{ tree_menu(tree_tuples, tree) }}
```
erikrose: Just give this a moment of thought to decide whether we still need to move this, since the promoted results are no longer on the right.

tests/test_core_plugin.py
```diff
@@ -23,6 +23,27 @@ class PathFilterTests(TestCase):
     More is covered in integration tests.
 
     """
+    def filter_str(self):
```
erikrose: I suppose this will turn into filter_unicode.

erikrose:
Otherwise, lgtm!

dxr/query.py
pelmers: For example it's common to exclude the object directory in a search, and it can be helpful to still have the promoted results.

dxr/templates/nunjucks/promoted.html
pelmers: pathLine is the result of concatenating some nunjucks renders, [source](https://github.com/mozilla/dxr/blob/8e56b088d3bdd074495140cf9ee9711ddf98d3e8/dxr/static_unhashed/js/dxr.js#L110-L116), and path_line.html does not have any |safe usage.

pelmers: But I can port buildResultHead into the Python.

dxr/query.py
erikrose: Of course. Thanks. Any FILE-domain filter could be useful.

dxr/templates/nunjucks/promoted.html
erikrose: No, that's okay. Your reasoning shows that it should really be safe. (A comment here laying out that same reasoning would not go amiss.)

pelmers:
I pushed a commit to clear out some comments, but looks like I missed a few. Not ready for review yet.

pelmers:
Alright, now it should be up to date.

dxr/query.py
```diff
@@ -126,52 +161,39 @@ def group_filters_by_name(predicate):
...
-                             if hasattr(f, 'highlight_path')]
-        return {'result_count': result_count,
-                'results': self._line_query_results(filters, results, path_highlighters)
-                           if is_line_query
-                           else self._file_query_results(results, path_highlighters)}
-
-        # Test: If var-ref (or any structural query) returns 2 refs on one line, they should both get highlit.
+        }
+        # We ask not to sort by path and line, to use score ranking instead.
+        result_count, results = self._search_and_highlight(query, file_filters, False, 0, promote_limit, False)
```
erikrose: should_alphabetize should be called using a kwarg at the callsite for clarity.

dxr/query.py
```diff
@@ -209,6 +231,43 @@ def direct_result(self):
...
+        :arg limit: maximum number of results to return
+        :arg should_alphabetize: whether to should_alphabetize the results alphabetically by path and
+        numerically by line
+
+        """
+        # Decide on the sort mode: either alphanumerically or by score.
+        if should_alphabetize:
+            sort_mode = ['path', 'number'] if is_line_query else ['path']
+        else:
+            sort_mode = ['_score']
```
erikrose: I think we need a secondary sort field to avoid arbitrary ordering like this: 
screen shot 2016-07-26 at 16 45 30

dxr/static_unhashed/css/dxr.css
```diff
@@ -190,13 +188,19 @@ code a {
     background-color: transparent;
 }
 
-/* results */
+/* promoted stuff, on the top */
+.promoted-container {
+    padding: 0.05em 1.0em 0;
```
erikrose: The border went poof. Did you mean for the border to disappear?

pelmers: I dropped the border because it overlapped the switch tree menu. And now I realize I could avoid it by using a margin instead.

erikrose: Why don't you see if you can bring it back and get it to work with the margin. The 2 sets of results blend together right now—but not quite enough to be consistent, with their differing indents for the icons—so it seems accidental.

pelmers:
Just realized the format bump was lost in the merge (since in the mean time master also bumped), so I'll need to increment it again.

erikrose:
In that case, we'll hold off on merging this until all the trees are green so we don't hold up future deploys.

dxr/static_unhashed/css/dxr.css
erikrose: Hmm. 
screen shot 2016-07-28 at 13 16 16

PR #544, opened 2016-05-15
Fix up destructor name tokenization in clang.

kleintom:
* Make sure we're including the full extents of a destructor name and not just the
leading '~'.

* Allow '~' in the 'Search for the substring' context menu option for C++ files.

### DIFF
--- a/dxr/plugins/clang/tests/test_menus/test_menus.py
+++ b/dxr/plugins/clang/tests/test_menus/test_menus.py
@@ -283,3 +283,28 @@ class MenuTests(DxrInstanceTestCase):
                 'pVirtualFunc',
                 {'html': 'Find overridden',
                  'href': '/code/search?q=%2Boverridden%3ADerivedClass%3A%3ApVirtualFunc%28%29'})
+
+    def test_destructor_declaration(self):
+        """Make sure we link the full destructor name and not just the '~' on a
+        destructor declaration, which can be easily confused for one's
+        complement on the result of a function call."""
+        menu_on(self.source_page('BaseTypes.h'),
+                '~Z',
+                {'html': 'Find declarations',
+                 'href': '/code/search?q=%2Bfunction-decl%3AZ%3A%3A%7EZ%28%29'})
+
+    def test_destructor_def(self):
+        """Make sure we link the full destructor name and not just the '~' on a
+        destructor def."""
+        menu_on(self.source_page('BaseTypes.h'),
+                '~BaseClass',
+                {'html': 'Find callers',
+                 'href': '/code/search?q=%2Bcallers%3ABaseClass%3A%3A%7EBaseClass%28%29'})
+
+    def test_destructor_call(self):
+        """Make sure we link the full destructor name and not just the '~' on a
+        destructor call."""
+        menu_on(self.source_page('main.cpp'),
+                '~Z',
+                {'html': 'Find references',
+                 'href': '/code/search?q=%2Bfunction-ref%3AZ%3A%3A%7EZ%28%29'})
--- a/dxr/plugins/clang/tests/test_menus/test_menus.py
+++ b/dxr/plugins/clang/tests/test_menus/test_menus.py
@@ -296,10 +296,10 @@ class MenuTests(DxrInstanceTestCase):
     def test_destructor_def(self):
         """Make sure we link the full destructor name and not just the '~' on a
         destructor def."""
-        menu_on(self.source_page('BaseTypes.h'),
-                '~BaseClass',
+        menu_on(self.source_page('Z.cpp'),
+                '~Z',
                 {'html': 'Find callers',
-                 'href': '/code/search?q=%2Bcallers%3ABaseClass%3A%3A%7EBaseClass%28%29'})
+                 'href': '/code/search?q=%2Bcallers%3AZ%3A%3A%7EZ%28%29'})
 
     def test_destructor_call(self):
         """Make sure we link the full destructor name and not just the '~' on a
--- a/dxr/static_unhashed/js/context_menu.js
+++ b/dxr/static_unhashed/js/context_menu.js
@@ -6,7 +6,16 @@ $(function() {
     // Get the file content container
     var fileContainer = $('#file'),
         queryField = $('#query'),
-        contentContainer = $('#content');
+        contentContainer = $('#content'),
+        nonWordCharRE = /[^A-Z0-9_]/i;
+
+    if (['cpp', 'cc', 'cxx', 'h', 'hxx', 'hpp'].indexOf(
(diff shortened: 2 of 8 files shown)


kleintom:
Sorry for all the talking, but there are two clang issues to deal with here:

1)  The clang lexer can't always differentiate between whether `~X();` should mean destructor decl or one's complement on the return of a function call.  In fact in that case it always chooses the latter (I think that's this: http://clang.llvm.org/doxygen/ParseExprCXX_8cpp_source.html#l02570 see also http://efcs.ca/draft/expr.unary.op.html 5.3.10), which is never what we want in our plugin uses of the lexer.

2) I'm seeing two different behaviors from clang related to destructors on d.m.o and locally.  On dxr.mozilla.org it seems every type of destructor reference gets a context menu only on the '~', with nothing on the accompanying class name part of the destructor string.  So, for example, here's a
destructor decldef: https://dxr.mozilla.org/mozilla-central/source/accessible/base/EventTree.h#30
destructor def: https://dxr.mozilla.org/mozilla-central/source/accessible/base/AccEvent.h#125
destructor ref: https://dxr.mozilla.org/mozilla-central/source/js/src/vm/UbiNodeCensus.cpp#77
(the fact that the context menu doesn't load on the '~'s is the JS part of this PR).  In terms of the plugin, the fact that we get destructor extents on *only* the '~' means that begin loc and end loc are the same for the def in `VisitFunctionDecl` and the ref in `printReference` [because if they were different then `afterToken(end loc)` would jump to the end of the class name and not the end of the '~'].

*Locally*, on my machine (with a slightly older m-c), I get different results:
at that decldef I get a context menu on the '~' only, but I also get a context menu from a type ref on the `TreeMutation();` part of `~TreeMutation();` (!):
```
decldef,name,"~TreeMutation",qualname,"mozilla::a11y::TreeMutation::~TreeMutation()",loc,"accessible/base/EventTree.h:30:2",locend,"accessible/base/EventTree.h:30:3",defloc,"accessible/base/EventTree.cpp:91:14",kind,"function"
ref,defloc,"accessible/base/EventTree.h:23:6",loc,"accessible/base/EventTree.h:30:3",locend,"accessible/base/EventTree.h:30:15",kind,"type",name,"TreeMutation",qualname,"mozilla::a11y::TreeMutation"
```
At the def and ref I get a context menu on the full destructor name (not just the '~'), but I also get refs on the non-tilde part of the destructor name as in the decldef.  Here's the plugin output for the ref:
```
call,callloc,"js/src/vm/UbiNodeCensus.cpp:75:8",calllocend,"js/src/vm/UbiNodeCensus.cpp:75:21",calleeloc,"js/src/vm/UbiNodeCensus.cpp:43:11",name,"~Count",qualname,"JS::ubi::SimpleCount::Count::~Count()",calltype,"static"
ref,defloc,"js/src/vm/UbiNodeCensus.cpp:43:11",loc,"js/src/vm/UbiNodeCensus.cpp:75:14",locend,"js/src/vm/UbiNodeCensus.cpp:75:20",kind,"function",name,"~Count",qualname,"JS::ubi::SimpleCount::Count::~Count()"
ref,defloc,"js/src/vm/UbiNodeCensus.cpp:43:11",loc,"js/src/vm/UbiNodeCensus.cpp:75:15",locend,"js/src/vm/UbiNodeCensus.cpp:75:20",kind,"type",name,"Count",qualname,"JS::ubi::SimpleCount::Count"
```
In the local case I checked that destructor names get different begin loc and end locs in the "ref" and "function" visits.

The combination of the two issues is why I wound up needing to both change `afterToken` (for the decldef issue 1) and add a new `recordLocEndForName` (for the difference in beginLoc and endLoc behavior for destructor refs and function defs issue 2).

kleintom:
Note that the spurious ref in the local case of issue 2) above in the "ref" and "function" cases leads to bad overlapping extents and the dreaded `htmlifier plugins requested overlapping <a> tags. Fix the plugins.` warning during build.  I think it would be possible to filter those out in the plugin, but it would be a hack, and we already happen to handle these overlapping extents they way we want (we drop the second ref), so personally I would lean toward just letting these go.  If we'd rather not though, I'd suggest a different bug :-)

dxr/static_unhashed/js/context_menu.js
```diff
@@ -6,7 +6,16 @@ $(function() {
     // Get the file content container
     var fileContainer = $('#file'),
         queryField = $('#query'),
-        contentContainer = $('#content');
+        contentContainer = $('#content'),
+        nonWordCharRE = /[^A-Z0-9_]/i;
+
+    if (['cpp', 'cc', 'cxx', 'h', 'hxx', 'hpp'].indexOf(
+          extension(window.location.pathname).toLowerCase()) > -1) {
```
kleintom: I'm not wild about doing this, so if people have better ideas feel free to let me know :-)
The only issue I can think of with just always including '~' as a word character is that in JS you see `~~` sometimes, and since we don't have analysis to separate the `~~` from what it's acting on with markup, the JS will consider `~~thing` to be a word it should search on, which isn't really what we want.

kleintom:
@pelmers r?

dxr/plugins/clang/tests/test_menus/test_menus.py
```diff
@@ -283,3 +283,28 @@ def test_override_pure_overridden(self):
                 'pVirtualFunc',
                 {'html': 'Find overridden',
                  'href': '/code/search?q=%2Boverridden%3ADerivedClass%3A%3ApVirtualFunc%28%29'})
+
+    def test_destructor_declaration(self):
+        """Make sure we link the full destructor name and not just the '~' on a
+        destructor declaration, which can be easily confused for one's
+        complement on the result of a function call."""
+        menu_on(self.source_page('BaseTypes.h'),
+                '~Z',
```
pelmers: I notice there are two '~Z' on the page with menus to check, can we test both? (or does this already?)

dxr/static_unhashed/js/context_menu.js
pelmers: It does look patchy, but I don't see a cleaner way either.

pelmers:
I'm not sure why there's inconsistency between the local and production m-c copies. Could it be from a different version of clang?
Other than the test comment I made, this looks good to me.

kleintom:
I moved the `~Z` def into its own file and then changed the destructor def test to refer to `~Z` instead of `~BaseClass` (I don't think we gain anything by doing both), but kept the `~BaseClass` def to avoid a compiler warning about calling delete on an abstract class with non-virtual destructor.

Thanks for the review Peter!

erikrose:
> I'm not sure why there's inconsistency between the local and production m-c copies. Could it be from a different version of clang?

That's a good guess. Prod uses, I believe, clang 3.3, and we're past that on the Docker container.

PR #546, opened 2016-05-25
Support Javascript!

pelmers:
I port over https://github.com/bill-mccloskey/mozsearch/blob/master/js-analyze.js of searchfox.org into DXR.

This plugin splits javascript identifiers into "vars" and "props," where "var" refers to local variables and "prop" refers to either global variables (i.e. properties of the global object) or properties of objects/classes.

At post_build, the plugin calls a node program that walks the source tree, looking for unignored .js files (nb. if we merge https://github.com/mozilla/dxr/pull/535 then I would not have to reimplement this; the Python could give the node these paths), and dumps JSON lines for each file in a mirrored directory tree in the temp folder.

At index the Python reads these file dumps to drop needles and refs to ES.

I intend to run this plugin on mozilla-central in the near future to see how the output looks and how long it takes (perhaps the organization of temp files could borrow from clang's version and use hashes).


tooling/docker/dev/set_up_ubuntu.sh
```diff
@@ -12,6 +12,11 @@ apt-get -q update \
         llvm-3.5 libclang-3.5-dev clang-3.5 \
         curl
 
+# Install newer node.
+apt-get remove -y nodejs
+curl -sL https://deb.nodesource.com/setup_6.x | bash -
```
erikrose: Not a blocker for this PR, but we should see if we can get the various curl-to-root-shell commands out of here. Since we've moved from a proper VM to Docker, our containment is worse.

dxr/plugins/js/makefile
```diff
@@ -1,5 +1,5 @@
 build:
-	npm install
+	cd analyze_js; npm install
```
erikrose: We'll need to use npm-lockdown on this before we can deploy it, like we do at DXR's top level. Otherwise, we could get fed any old code from the npm repo.

erikrose:
FWIW, using hashes probably won't get us anything and might even get us worse performance by sticking a zillion files in one dir. The reason we use hashes on C++ is that a given file can end up with multiple effective contents due to preprocessor directives, so we throw both the path and the preprocessed contents into the hash.

pelmers:
With js as the only plugin and no build run, mozilla-central completes in 44 minutes. I'll be rolling in another update to this which reflects what it took to get that working...

dxr/plugins/js/analyze_js/analyze_tree.js
```diff
@@ -0,0 +1,81 @@
+const path = require("path");
+const fs = require('fs');
```
erikrose: Eventually we should standardize on a kind of quotes. And use eslint. I have a good default config in Fathom.

dxr/plugins/js/analyze_js/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "js_esprima",
+  "description": "esprima-based javascript plugin for DXR",
+  "author": "Mozilla",
+  "version": "1.0.0",
+  "engine": "node >= 0.8.x",
```
erikrose: I betcha it needs at least node 6.0.0 (when ES6 support was introduced), since it uses [ES6 method properties](http://es6-features.org/#MethodProperties),

dxr/plugins/js/analyze_js/analyze_file.js
```diff
@@ -0,0 +1,727 @@
...
+function analyzeJS(filepath, relpath, tempFilepath)
+{
+  fileIndex = relpath;
+  nextSymId = 0;
+  outLines = [];
+  let text = preprocess(String(fs.readFileSync(filepath)), line => "//" + line);
+  try {
+    let ast = esprima.parse(text, {loc: true, source: path.basename(filepath), line: 1, sourceType: "script"});
+    if (ast) {
+        Analyzer.program(ast);
```
erikrose: I notice we don't make a `new Analyzer`. Are some of the hashes Analyzer keeps around important for maintaining cross-file relationships?

dxr/plugins/js/filters.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class _QualifiedNameFilter(QualifiedNameFilterBase):
+    lang = PLUGIN_NAME
+
+
+class PropFilter(_QualifiedNameFilter):
+    name = 'prop'
+    is_identifier = True
+    description = Markup('Property definition filter: <code>prop:foo</code>')
```
erikrose: Let's stick "JavaScript" in the description somewhere for now. Eventually, I want to badge the filters in the menu according to which languages they support and then also badge the trees in the Switch Tree menu according to which language plugins are enabled.

dxr/plugins/js/indexers.py
```diff
@@ -0,0 +1,105 @@
+from os.path import basename, dirname, relpath, join, exists
+import json
+import subprocess
+
+from funcy import imap
```
erikrose: Would itertools.imap work instead? I rather don't trust funcy; I've found some real inefficiencies in there, and I would like to get rid of it eventually.

dxr/plugins/js/indexers.py
```diff
@@ -0,0 +1,105 @@
...
+class ReadAnalysis(object):
+    def __init__(self, tree, lines, contents):
+        self.needles = []
+        self.refs = []
+        # Build map of line number -> byte offset to use for emitting refs.
+        self.offsets = list(cumulative_sum(imap(len, contents.splitlines(True))))
+        for line in lines:
+            row, (start, end) = line['loc']
+            qref = QualifiedRef(tree, (line['sym'], line['name'], line['type']), qualname=line['sym'])
+            if line['kind'] == "use":
```
erikrose: double quotes

dxr/plugins/js/indexers.py
```diff
@@ -0,0 +1,105 @@
...
+                self.yield_needle(line['type'] + "_ref", row, start, end, line['name'], line['sym'])
+                self.yield_ref(row, start, end, qref)
+            elif line['kind'] == "def":
+                self.yield_needle(line['type'], row, start, end, line['name'], line['sym'])
```
erikrose: Can we factor out the repetition between this and the `if` clause?

dxr/plugins/js/indexers.py
```diff
@@ -0,0 +1,105 @@
...
+    def yield_needle(self, filter_name, line, start, end, name, qualname=None):
+        """Add needle for qualified filter_name from line:start
+        to line:end with given name and qualname.
+        """
+        mapping = {'name': name}
+        if qualname:
+            mapping['qualname'] = qualname
+        self.needles.append((PLUGIN_NAME + '_' + filter_name,
+                             mapping,
+                             Extent(Position(row=line, col=start), Position(row=line, col=end))))
```
erikrose: I wonder if JS has any line-continuation stuff that might screw this up in weird cases.

dxr/plugins/js/__init__.py
```diff
@@ -0,0 +1,30 @@
+"""JS plugin: analyze Javascript files by building a symbol map on parse by
+esprima.
+"""
+
+from dxr.plugins import Plugin, filters_from_namespace, refs_from_namespace
+from dxr.filters import LINE
+from dxr.indexers import QUALIFIED_LINE_NEEDLE
+from dxr.plugins.js.indexers import _TreeToIndex
```
erikrose: Nope, can't import a _private_name from another module, even our own. :-) Generally I like to `import dxr.indexers` and explicitly subclass `dxr.indexers.TreeToIndex` in the other file.

dxr/plugins/js/indexers.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class _FileToIndex(FileToIndex):
+    def __init__(self, path, contents, plugin_name, tree):
+        super(_FileToIndex, self).__init__(path, contents, plugin_name, tree)
+        self.analysis_path = join(join(join(tree.temp_folder, "plugins/js"),
+                                       relpath(dirname(self.absolute_path()), tree.source_folder)),
+                                  basename(path) + '.data')
+        lines = []
+        if exists(self.analysis_path):
```
erikrose: Call `self.is_interesting()` here iff you think the semantic is durable.

dxr/plugins/js/indexers.py
```diff
@@ -0,0 +1,105 @@
...
+        return exists(self.analysis_path)
+
+    def parse_analysis(self, lines):
+        def parse_loc(line):
+            if 'loc' in line:
+                row, col = line['loc'].split(':', 1)
+                if '-' in col:
+                    col = tuple(map(int, col.split('-', 1)))
+                else:
+                    col = (int(col), int(col))
```
erikrose: parens not needed

dxr/plugins/js/analyze_js/analyze_file.js
```diff
@@ -392,6 +392,12 @@ let Analyzer = {
     case "Super":
       break;
 
+    case "TaggedTemplate":
```
erikrose: Good idea.

dxr/plugins/js/tests/test_refs_and_needles/test_refs_and_needles.py
```diff
@@ -16,6 +16,10 @@ def test_needles(self):
                            "return <b>y</b>;", 9)
         self.found_line_eq("ref:foo",
                            "stuff.doStuff(<b>foo</b>);", 16)
+        self.found_line_eq("+id:Guy#fun",
+                           "this.<b>fun</b> = function thathasaname() {", 21)
```
erikrose: This is particularly impressive to me: that it realized `this` refers to the `new Guy`.

erikrose:
This looks really good. I'm looking forward to it landing! Give me a poke once you've had a chance to look at my comments.

erikrose:
r+!

PR #547, opened 2016-05-26
JS testing

kleintom:
I didn't realize at first that slimerjs (and phantomjs) is not a node module:
https://www.npmjs.com/package/slimerjs#a-note-on-slimerjs
So to get integration with mocha/chai I tried node-phantom-simple, which, in spite of the name, also supports slimerjs (since phantomjs and slimerjs basically have the same api).

I got that working outside docker on my machine, including running inside xvfb-run, but then discovered when I moved it to docker that there's an issue with node-phantom-simple/slimerjs on docker: cf. https://github.com/baudehlo/node-phantom-simple/issues/127 for a vagrant issue that produces what I'm seeing (comment 0 and the node-phantom-simple self-test results in particular), even under xvfb-run and the travis setup he suggests trying.  So right now it's using phantomjs instead of slimerjs (fortunately that's just a path change).


kleintom:
There are other options:

For one, to make things less nested, I can probably (maybe?) get `node-phantom-promise`:
https://github.com/foray1010/node-phantom-promise
 working with `co`.  Right now node-phantom-promise uses `for(blah of bleh)` which isn't supported for older nodes like our docker, so I guess we'd either ask the maintainer to change  that or just pull in a patched version since it's just a small wrapper to promisify node-phantom-simple calls.

Alternatively, since it turned out node-phantom-simple actually *doesn't* work with slimerjs in our case, maybe we'd be more interested in something like mocha-casperjs https://github.com/nathanboktae/mocha-casperjs instead, which also doesn't currently support slimerjs?  There's also casperjs on its own, which supports slimerjs (in docker?) and has its own testing facilities.  Or we could always just try calling straight slimerjs from mocha, but I feel like then you lose any sort of fine-grainedness reporting/debugging-wise (or you create a lot of small slimerjs tests).

Feedback?

pelmers:
Remark: we have node 6 now in the dev setup, and soon we should have it in prod too to support the new js plugin.

tests/js/dxr.js
```diff
@@ -0,0 +1,87 @@
+"use strict";
+
+const exec = require('child_process').exec;
+const node_modules = '../../tooling/node/node_modules/'
```
pelmers: While not necessary, I'd end this line with a ; to match the others

pelmers:
Do we want to standardize on a particular quoting convention? static/dxr.js mainly uses single quotes.

pelmers:
Beyond that, this PR looks excellent to me; it'll be a big event to merge this and start writing some more tests.

kleintom:
Thanks Peter!  Any thoughts @erikrose?

kleintom:
> I got that working outside docker on my machine, including running inside xvfb-run, but then discovered when I moved it to docker that there's an issue with node-phantom-simple/slimerjs on docker: cf. baudehlo/node-phantom-simple#127 for a vagrant issue that produces what I'm seeing (comment 0 and the node-phantom-simple self-test results in particular), even under xvfb-run and the travis setup he suggests trying. So right now it's using phantomjs instead of slimerjs (fortunately that's just a path change).

Update: looks like there's headless firefox work going on:

Headless slimerjs support (resolved fixed):
https://bugzilla.mozilla.org/show_bug.cgi?id=1353935

and the meta Headless Browsing Mode:
https://bugzilla.mozilla.org/show_bug.cgi?id=1338004

PR #550, opened 2016-05-27
Reload on popState event every time the page being loaded is a search.

kleintom:
This fixes the case where we start a new page with a search url (which doesn't
trigger a history push), and then search some more, and then go back.


kleintom:
@pelmers r?

pelmers:
Looks like that fixes it! r+
You can reference Bug 1276270 to resolve the bug.

PR #551, opened 2016-05-31
Add support for clang 3.7.

kleintom:
My system just upgraded to 3.7 - tests pass and m-c compiles; also tests pass on docker at 3.5.  (Cur stable ubuntu is at 3.8, debian is at 3.5, for reference.)

Let me know if you don't like the indenting :-)


erikrose:
lgtm! Good job!

PR #553, opened 2016-06-01
Reduce memory usage in js plugin

pelmers:
Mozilla-central has not been able to index since JS landed because of out of memory.
A js-only run's memory profile looks as follows.
jsonly
The largest spike corresponds to mandreel.js, a 277k-line file which was the most likely cause of out of memory.
I add a limit of 100k lines to the JS analyzer because even with the improvements this file will still require about 1.4G of memory, which might still be too much given 4 workers at a time.

### DIFF
--- a/dxr/plugins/js/indexers.py
+++ b/dxr/plugins/js/indexers.py
@@ -1,3 +1,4 @@
+from collections import namedtuple
 from itertools import imap
 import json
 import subprocess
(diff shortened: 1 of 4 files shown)


dxr/plugins/js/indexers.py
```diff
@@ -72,31 +53,52 @@ def __init__(self, path, contents, plugin_name, tree):
...
+        # All lines from the analysis output file.
+        self.lines = []
+        # Map of line number -> byte offset to use for emitting refs.
+        self.offsets = []
         if self.is_interesting():
             with open(self.analysis_path) as analysis:
-                lines = self.parse_analysis(analysis.readlines())
-            lines = sorted(lines, key=lambda x: x['loc'])
-        self.analyzer = ReadAnalysis(tree, lines, contents)
+                self.lines = sorted((self.parse_analysis(line) for line in analysis), key=lambda x: x.loc)
```
erikrose: So the entries in the JSON file don't get laid down in order?

pelmers: They're laid down in AST traversal order, which isn't necessarily the order of appearance in the source file.

pelmers: For example, the right side of an assignment is visited before the left side.

PR #554, opened 2016-06-02
Reduce peak memory

pelmers:
Clear dicts after use, add 'clean' to skip_stages option.
Because a list held references to all the dicts, memory usage increased
as the indexer adds fields to the dict, even though it never used the
dict after yielding it, especially applicable to files with many lines.
Reduce peak memory usage by 10-20%.

I add "clean" to skip_stages since I had it for testing and it seems intuitive to have, but I can back that part out.

Also the sorted iterable_per_line only reduces memory usage marginally, because the peak still occurs when all the needles are concretized and then sorted in finished_tags, so I could bring that out too for simplification without much impact (tens of megabytes).

The second change is that files are initially only read up to the first 4kb, and the full contents are not read until either 1) we know it might not be binary, or 2) it's an image whose contents we store.

### DIFF
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -197,7 +197,7 @@ def index_tree(tree, es, verbose=False):
 
     skip_indexing = 'index' in config.skip_stages
     skip_build = 'build' in config.skip_stages
-    skip_cleanup  = skip_indexing or skip_build
+    skip_cleanup = skip_indexing or skip_build or 'clean' in config.skip_stages
 
     # Create and/or clear out folders:
     ensure_folder(tree.object_folder, tree.source_folder != tree.object_folder)
@@ -418,7 +418,6 @@ def unignored(folder, ignore_paths, ignore_filenames, want_folders=False):
             for f in folders:
                 yield join(root, f)
 
-
 def index_file(tree, tree_indexers, path, es, index):
     """Index a single file into ES, and build a static HTML representation of it.
 
@@ -474,7 +473,11 @@ def index_file(tree, tree_indexers, path, es, index):
                                file_to_index.annotations_by_line())
 
     def docs():
-        """Yield documents for bulk indexing."""
+        """Yield documents for bulk indexing.
+
+        Big Warning: docs also clears the contents of all elements of
+        needles_by_line because they will no longer be used.
+        """
         # Index a doc of type 'file' so we can build folder listings.
         # At the moment, we send to ES in the same worker that does the
         # indexing. We could interpose an external queueing system, but I'm
@@ -522,6 +525,11 @@ def index_file(tree, tree_indexers, path, es, index):
                 if annotations_for_this_line:
                     total['annotations'] = annotations_for_this_line
                 yield es.index_op(total)
+
+                # Because needles_by_line holds a reference, total is not
+                # garbage collected. Since we won't use it again, we can clear
+                # the contents, saving substantial memory on long files.
+                total.clear()
 
     # Indexing a 277K-line file all in one request makes ES time out (>60s),
     # so we chunk it up. 300 docs is optimal according to the benchmarks in
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -12,6 +12,7 @@ from sys import exc_info
 from traceback import format_exc
 from uuid import uuid1
 
+from binaryornot.helpers import is_binary_string
 from concurrent.futures import as_completed, ProcessPoolExecutor
 from click import progressbar
 from flask import current_app
@@ -26,7 +27,7 @@ from dxr.es import UNINDEXED_STRING, UNANALYZED_STRING, TREE, create_index_and_w
 from dxr.exceptions import BuildError
 from dxr.filters import LINE, FILE
 from dxr.lines import es_lines, finished_tags
-from dxr.mime import decode_file, is_binary_image
+from dxr.mime import decode_data, is_binary_image
 from dxr.utils import (open_log, deep_update, append_update,
                        append_update_by_line, append_by_line, bucket)
 from dxr.vcs import VcsCache
@@ -353,11 +354,9 @@ def _unignored_folders(folders, source_path, ignore_filenames, ignore_paths):
                 yield folder
 
 
-def file_contents(path, encoding_guess, force_read=False):  # TODO: Make accessible to TreeToIndex.post_build.
-    """Return the unicode contents of a file if we can figure out a decoding.
-    Otherwise, we could not decode it and the result is a string with some of
-    the contents. Unless force_read is true, in which case we return a string
-    of the file's contents even if they could not decode.
+def file_contents(path, encoding_guess):  # TODO: Make accessible to TreeToIndex.post_build.
+    """Return the unicode contents of a file if we can figure out a decoding,
+    or else None.
 
     :arg path: A sufficient path to the file
     :arg encoding_guess: A guess at the encoding of the file, to be applied if
@@ -366,12 +365,15 @@ def file_contents(path, encoding_guess, force_read=False):  # TODO: Make accessi
     """
     # Read the binary contents of the file.
     with open(path, 'rb') as source_file:
-        decoded, contents = decode_file(source_file, encoding_guess)
-        if not decoded and force_read:
+        initial_portion = source_file.read(4096)
+        if not is_binary_string(initial_portion):
+            # Move the cursor back to the start of the file.
             source_file.seek(0)
-            contents = source_file.read()
-        # unicode if we were able to decode, str if not
-        return contents
+            decoded, contents = decode_data(source_file.read(),
+                                            encoding_guess,
+                                            can_be_binary=False)
+            if decoded:
+                return contents
 
 
 def unignored(folder, ignore_paths, ignore_filenames, want_folders=False):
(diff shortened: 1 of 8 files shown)


pelmers:
The next peak on the graph is from reading in the entire contents of big binaries and then tossing them out once we find out they're binary. More to come...

dxr/build.py
```diff
@@ -523,6 +522,11 @@ def docs():
                     total['annotations'] = annotations_for_this_line
                 yield es.index_op(total)
 
+                # Because needles_by_line holds a reference, total is not
```
erikrose: Can you point out where needles_by_line takes on a reference to `total`? I'm not seeing it.

dxr/build.py
```diff
@@ -197,7 +197,7 @@ def delete_index_quietly(es, index):
 
     skip_indexing = 'index' in config.skip_stages
     skip_build = 'build' in config.skip_stages
-    skip_cleanup  = skip_indexing or skip_build
+    skip_cleanup = skip_indexing or skip_build or 'clean' in config.skip_stages
```
erikrose: We'll need to add this to configuration.rst.

dxr/lines.py
```diff
@@ -464,15 +464,11 @@ def finished_tags(lines, refs, regions):
...
-
-    tags.extend(line_boundaries(lines))
-
-    # Sorting is actually not a significant use of time in an actual indexing
-    # run.
-    tags.sort(key=nesting_order)  # balanced_tags undoes this, but we tolerate
-                                  # that in html_lines().
+    # balanced_tags undoes the sorting, but we tolerate that in html_lines().
+    # Remark: this sort is the memory peak.
+    tags = sorted(chain(tag_boundaries(chain(refs, regions)),
```
erikrose: The reason I used sort() rather than sorted() was that I believe sorted() makes a copy of the collection being sorted, doubling RAM use. (This would have to be true, since sorted() works on any (possibly lazy) iterable, not just lists.)

dxr/build.py
pelmers: total came from needles_by_line since it's the first part of the izip'd tuple.

dxr/lines.py
pelmers: sorted does copy the collection, but the collection here is a generator iterable that will not become a list until within sorted.

erikrose: Good point. I do mourn the loss of the "sorting not a significant use of time" comment, though. Can we keep that?

dxr/build.py
erikrose: So total points to needles_by_line but nothing points to total. So how is total not getting GC'd?

Ah, I see: total is a pointer into a member of needles_by_line, which we mutably inflate by updating it with the file-wide needles. It makes me a little uncomfortable to go around destructively clear()ing pieces of a closed-over variable in a function that might not obviously be expected to. So I think we should either make a loud note about that in docs()'s docstring or else take a different tack: make a shallow copy of the member of needles_by_line and put that into `total` as our first step in the loop. (If the latter, I'd be curious how much it hurts speed.)

pelmers: shallowcpy
.copy()
mandreel_better_w_sorted
.clear()
Looks like a small difference, I can switch over to the copying.

dxr/build.py
pelmers: [done](https://github.com/mozilla/dxr/pull/554/commits/2eb232c7252696cf1d27658f71e55337267c8713)

dxr/build.py
erikrose: I like the clearing better because it's noticeably faster (5 seconds across an 80-second index—correct me if I'm wrong). Let's just document the hell out of it.

dxr/build.py
```diff
@@ -370,9 +367,19 @@ def file_contents(path, encoding_guess):  # TODO: Make accessible to TreeToIndex
...
+            # Create a read-only mmap of the entire file.
+            mmap_block = mmap.mmap(source_file.fileno(), 0, prot=mmap.PROT_READ, flags=mmap.MAP_PRIVATE)
+            decoded, contents = decode_block(mmap_block, encoding_guess)
+            if not decoded and force_read:
+                mmap_block.seek(0)
+                contents = str(mmap_block)
+            mmap_block.close()
+            # unicode if we were able to decode, str if not
+            return contents
+        except ValueError:
```
erikrose: Does the exception occur on the initial mmap() call? Can we tighten up this exception block at all? I worry because ValueError is pretty generic.

dxr/build.py
```diff
@@ -433,7 +439,7 @@ def index_file(tree, tree_indexers, path, es, index):
 
     """
     try:
-        contents = file_contents(path, tree.source_encoding)
+        contents = file_contents(path, tree.source_encoding, is_binary_image(path))
```
erikrose: Call kwarg by name

dxr/mime.py
```diff
@@ -14,11 +14,26 @@ def icon(path, is_binary=False):
...
+    """Given mmap'd block, return an (is_text, data) tuple, where data is
+    returned as unicode if we think the block is text and were able to
+    determine an encoding for it, otherwise a string."""
+    initial_portion = mmap_block[:4096]
+    if not is_binary_string(initial_portion):
+        # Move the cursor back to the start of the block.
+        mmap_block.seek(0)
+        return decode_data(mmap_block.read(mmap_block.size()),
+                           encoding_guess, False)
+    return False, initial_portion
```
erikrose: What's the use of returning just the first 4K?

dxr/build.py
```diff
@@ -2,10 +2,10 @@
 from errno import ENOENT
 from fnmatch import fnmatchcase
 from itertools import chain, izip, repeat
-from operator import attrgetter
+import mmap
```
erikrose: We should note how much RAM mmap saves in the mmap commit message.

erikrose:
Otherwise, looks good to me.

dxr/mime.py
pelmers: The previous behavior returns a str on decode failure, so I keep it consistent here by returning str. The 4kb is enough for anyone later to figure out it came from a binary file.

dxr/build.py
```diff
@@ -370,9 +367,19 @@ def file_contents(path, encoding_guess):  # TODO: Make accessible to TreeToIndex
     """
     # Read the binary contents of the file.
     with open(path, 'rb') as source_file:
-        contents = source_file.read()  # always str
-    _, contents = decode_data(contents, encoding_guess)
-    return contents  # unicode if we were able to decode, str if not
+        try:
+            # Create a read-only mmap of the entire file.
+            mmap_block = mmap.mmap(source_file.fileno(), 0, prot=mmap.PROT_READ, flags=mmap.MAP_PRIVATE)
```
erikrose: Can we import these deeply?

dxr/mime.py
erikrose: It makes me uncomfortable because there's no way for the caller to tell if the result has been truncated.

pelmers: That's true, I wasn't sure what else I could reasonably return here though. I can write about this in the FileToIndex docs. Do you see a better way?

dxr/build.py
pelmers: added!

dxr/build.py
```diff
@@ -359,9 +353,11 @@ def _unignored_folders(folders, source_path, ignore_filenames, ignore_paths):
                 yield folder
 
 
-def file_contents(path, encoding_guess):  # TODO: Make accessible to TreeToIndex.post_build.
+def file_contents(path, encoding_guess, force_read=False):  # TODO: Make accessible to TreeToIndex.post_build.
     """Return the unicode contents of a file if we can figure out a decoding.
-    Otherwise, return the contents as a string.
+    Otherwise, we could not decode it and the result is a string with some of
```
erikrose: These 2 sentences seem to say the same thing happens for 2 different scenarios.

dxr/indexers.py
```diff
@@ -131,7 +131,8 @@ def file_to_index(self, path, contents):
         :arg path: A path to the file to index, relative to the tree's source
             folder
         :arg contents: What's in the file: unicode if we managed to guess an
-            encoding and decode it, str otherwise
+            encoding and decode it, str otherwise (only the first 4 kb unless
```
erikrose: The first-4K thing still doesn't strike me as useful and is potentially surprising. Either an indexer won't want to do anything with the first 4K, or they'll probably want the whole file. What do you think of simply returning None for binary files? That saves RAM in the case of huge ones, and interested plugins can still go get the file themselves based on the path and tree location and read it in segments. This also cleans up file_contents(), which no longer would need a force_read kwarg: it would just return None on binary.

pelmers: That sounds great, actually

dxr/build.py
```diff
@@ -360,8 +355,8 @@ def _unignored_folders(folders, source_path, ignore_filenames, ignore_paths):
 
 
 def file_contents(path, encoding_guess):  # TODO: Make accessible to TreeToIndex.post_build.
-    """Return the unicode contents of a file if we can figure out a decoding.
-    Otherwise, return the contents as a string.
+    """Return the unicode contents of a file if we can figure out a decoding,
```
erikrose: One last thing: since we changed the semantics, we should rename this `unicode_contents()` to reflect them. That also provides contrast with `file_contents_at_rev()`, which might still be on the binary-returning bandwagon.

dxr/plugins/core.py
```diff
@@ -464,6 +464,10 @@ def needles(self):
         # We store both the contents of textual images twice so that they can
         # both show up in searches and be previewed in the browser.
         if is_binary_image(self.path) or is_textual_image(self.path):
+            # If the file was binary, then contents are None, so read it here.
+            if self.contents is None:
+                with open(self.absolute_path(), "rb") as imagefile:
```
erikrose: single quotes

erikrose: image_file

PR #557, opened 2016-06-06
Fix fallout from unicode_contents introduction.

pelmers:
Have the Python and JS extensions calls parent's is_interesting as well.

See https://jenkins-dxr.mozilla.org/job/hgcustom_version-control-tools/lastBuild/console for an example of the error.


dxr/plugins/python/analysis.py
```diff
@@ -50,17 +50,19 @@ def _analyze_file(self, path, encoding):
 
         """
         try:
-            syntax_tree = ast_parse(unicode_contents(path, encoding))
+            contents = unicode_contents(path, encoding)
+            syntax_tree = ast_parse(contents) if contents else None
```
erikrose: Consider whether this and the one in js.indexers should be `if contents is None`.

dxr/indexers.py
```diff
@@ -326,12 +329,8 @@ def _line_offsets(self):
...
                 raise ValueError("Can't get line offsets for a file that isn't"
                                  " text.")
-            lines = self.contents.splitlines(True) if self.contents else []
-            self._line_offset_list = []
-            chars = 0
-            for i in xrange(0, len(lines)):
-                self._line_offset_list.append(chars)
-                chars += len(lines[i])
+            lines = self.contents.splitlines(True) if self.contents is not None else []
+            self._line_offset_list = list(cumulative_sum(imap(len, lines)))
```
erikrose: nice!

erikrose: Might even be able to de-dupe further with _browse_file(). Not sure if it makes sense.

pelmers: I noticed the offset-building there too, but thought it was already pretty concise, shall I move both up to a build_offsets(lines) function?

dxr/plugins/xpidl/indexers.py
```diff
@@ -34,7 +34,7 @@ def idl(self):
 
     def is_interesting(self):
         # TODO: consider adding a link from generated headers back to the idl
-        return self.path.endswith('.idl')
+        return self.path.endswith('.idl') and super(FileToIndex, self).is_interesting()
```
erikrose: The order here is different from the other ones. :-)

erikrose:
r+ assuming tests pass

dxr/plugins/xpidl/indexers.py
pelmers: I picked this one since the idl filter is more restrictive and doesn't require a disk access so it makes sense to do first.

erikrose: Good excuse for a comment, then, so I don't screw it up later :-)

dxr/indexers.py
pelmers: And I notice the offset-building happens two more times in test_lines, so definitely should make the function.

PR #558, opened 2016-06-06
Reduce memory usage by using __slots__ on frozendict.

pelmers:
On the largest files (sqlite3.c) this change reduces memory usage around
10%.

### DIFF
--- a/dxr/utils.py
+++ b/dxr/utils.py
@@ -186,17 +186,54 @@ def cached(f):
     return inner
 
 
-class frozendict(dict):
-    """A dict that can be hashed if all its values are hashable
+# Cache FrozenDictLike objects to avoid reconstruction on every usage.
+_dict_cache = {}
 
-    You shouldn't modify one of these once constructed; it will change the
-    hash.
 
+def _frozendict_factory(dict_keys):
+    """Factory that produces FrozenDictLike objects whose slots match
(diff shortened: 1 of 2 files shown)


dxr/utils.py
```diff
@@ -186,17 +186,54 @@ def inner(plugins):
...
+
+    return FrozenDictLike
+
+
+def frozendict(current_dict):
+    """Freeze current_dict whose keys are all strings.
+    """
+    keys = tuple(sorted(current_dict.keys()))
+    if keys not in _dict_cache:
+        _dict_cache[keys] = _frozendict_factory(keys)
```
erikrose: So this cache just grows forever? I think we're getting away with this only because we use this function only from workers, which have a limited lifetime. That might not always be true.

pelmers: Yes, currently it grows forever... for now there's only as many as POSSIBLE_KINDS because only the clang plugin uses frozendict.

pelmers: Do any of the currently included packages have a sort of limited cache decorator?

erikrose: No. Python 3 introduces an LRU cache, but that doesn't help us.

pelmers: What do you think of a very simple policy: replace the oldest-created item?

PR #559, opened 2016-06-08
Avoid reflow on click by unwrapping and removing padding

pelmers:
Fixes bug 1279001 by avoiding DOM mutation and box model recalculating (not changing padding).

Unfortunately it looks a little less pretty because we don't have the padding around the background.
screen shot 2016-06-08 at 1 28 38 pm

### DIFF
--- a/dxr/lines.py
+++ b/dxr/lines.py
@@ -160,7 +160,7 @@ class Ref(object):
         else:
             title = ''
         if self.qualname_hash is not None:
-            cls = ' class="tok%i"' % self.qualname_hash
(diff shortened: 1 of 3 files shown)


pelmers:
I closed temporarily because I hadn't run any tests and feared that I'd break menu_on. (but I think it's alright now)

mstange:
You could add a box-shadow if you want the yellow rectangle to be larger.

erikrose:
I think querySelector will still iterate across the whole tree (I recently read the source, and the only fast path is for IDs), but this should at least be an improvement.

PR #561, opened 2016-06-09
Add a descriptor plugin that emulates MXR's descriptor column.

pelmers:
Also add some framework for plugins to supply needles on folders and
adding headers to the browsing view.

Fixes bug 1157968.
Example
screen shot 2016-06-09 at 8 55 06 am

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -203,6 +203,7 @@ collection of subcomponents which do the actual work:
 
 .. digraph:: plugin
 
+   "Plugin" -> "FolderToIndex";
    "Plugin" -> "TreeToIndex" -> "FileToIndex";
    "Plugin" -> "FileToSkim";
    "Plugin" -> "filters";
@@ -248,8 +249,14 @@ manually:
     .. autoclass:: dxr.plugins.Plugin
        :members:
 
-Actual plugin functionality is implemented within tree indexers, file
-indexers, filters, and skimmers.
+Actual plugin functionality is implemented within file indexers, tree indexers,
+file indexers, filters, and skimmers.
+
+Folder Indexers
+=============
+
+.. autoclass:: dxr.indexers.TreeToIndex
+   :members:
 
 Tree Indexers
 =============
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -253,9 +253,9 @@ Actual plugin functionality is implemented within file indexers, tree indexers,
 folder indexers, filters, and skimmers.
 
 Folder Indexers
-=============
+===============
 
-.. autoclass:: dxr.indexers.TreeToIndex
+.. autoclass:: dxr.indexers.FolderToIndex
    :members:
 
 Tree Indexers
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -250,7 +250,7 @@ manually:
        :members:
 
 Actual plugin functionality is implemented within file indexers, tree indexers,
-file indexers, filters, and skimmers.
+folder indexers, filters, and skimmers.
 
 Folder Indexers
 =============
--- a/dxr/app.py
+++ b/dxr/app.py
@@ -358,6 +358,14 @@ def browse(tree, path=''):
                             frozen['generated_date'])
 
 
+def concat_plugin_headers(plugin_list):
+    """Return a list of the concatenation of all browse_headers in the
+    FolderToIndexes of given plugin list.
+    """
+    return list(chain.from_iterable(p.folder_to_index.browse_headers
+                                    for p in plugin_list if p.folder_to_index))
+
+
 def _browse_folder(tree, path, config):
     """Return a rendered folder listing for folder ``path``.
 
@@ -378,6 +386,7 @@ def _browse_folder(tree, path, config):
 
     frozen = frozen_config(tree)
 
+    plugin_headers = concat_plugin_headers(plugins_named(frozen['enabled_plugins']))
     files_and_folders = filtered_query(
         frozen['es_alias'],
         FILE,
@@ -385,7 +394,7 @@ def _browse_folder(tree, path, config):
         sort=[{'is_folder': 'desc'}, 'name'],
         size=1000000,
         include=['name', 'modified', 'size', 'link', 'path', 'is_binary',
-                 'is_folder'])
+                 'is_folder'] + plugin_headers)
 
     if not files_and_folders:
         raise NotFound
@@ -403,6 +412,7 @@ def _browse_folder(tree, path, config):
         generated_date=frozen['generated_date'],
         google_analytics_key=config.google_analytics_key,
         paths_and_names=_linked_pathname(path, tree),
+        plugin_headers=plugin_headers,
         filters=filter_menu_items(
             plugins_named(frozen['enabled_plugins'])),
         # Autofocus only at the root of each tree:
@@ -416,6 +426,7 @@ def _browse_folder(tree, path, config):
              f['name'],
              decode_es_datetime(item_or_list(f['modified'])) if 'modified' in f else None,
              f.get('size'),
+             [f.get(h, [''])[0] for h in plugin_headers],
              url_for('.browse', tree=tree, path=f.get('link', f['path'])[0]))
             for f in files_and_folders])
 
--- a/dxr/app.py
+++ b/dxr/app.py
@@ -361,6 +361,7 @@ def browse(tree, path=''):
 def concat_plugin_headers(plugin_list):
     """Return a list of the concatenation of all browse_headers in the
     FolderToIndexes of given plugin list.
+
     """
     return list(chain.from_iterable(p.folder_to_index.browse_headers
                                     for p in plugin_list if p.folder_to_index))
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -602,7 +602,7 @@ def index_folders(tree, index, es):
                 continue
             needles = {'is_folder': True}
             for name, folder_to_index in folder_indexers:
-                needles.update(dict((folder_to_index(name, tree, folder)).needles()))
+                needles.update(dict(folder_to_index(name, tree, folder)).needles())
             es.index(index, FILE, needles)
 >>>>>>> 639bf4c... Add a descriptor plugin that emulates MXR's descriptor column.
 
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -604,7 +604,6 @@ def index_folders(tree, index, es):
             for name, folder_to_index in folder_indexers:
                 needles.update(dict(folder_to_index(name, tree, folder)).needles())
             es.index(index, FILE, needles)
->>>>>>> 639bf4c... Add a descriptor plugin that emulates MXR's descriptor column.
 
 
 def index_files(tree, tree_indexers, index, pool, es):
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -587,6 +587,8 @@ def index_chunk(tree,
 
 def index_folders(tree, index, es):
     """Index the folder hierarchy into ES."""
+    folder_indexers = [(p.name, p.folder_to_index)
+                       for p in tree.enabled_plugins if p.folder_to_index]
     with aligned_progressbar(unignored(tree.source_folder,
                                        tree.ignore_paths,
                                        tree.ignore_filenames,
@@ -594,19 +596,15 @@ def index_folders(tree, index, es):
                      show_eta=False,  # never even close
                      label='Indexing folders') as folders:
         for folder in folders:
-            rel_path = relpath(folder, tree.source_folder)
-            # If the path is not in UTF-8, then elasticsearch will not
-            # accept the path, so just move on.
             try:
-                rel_path = rel_path.decode('utf-8')
+                folder = folder.decode('utf-8')
             except UnicodeDecodeError:
                 continue
-            superfolder_path, folder_name = split(rel_path)
-            es.index(index, FILE, {
-                'path': [rel_path],  # array for consistency with non-folder file docs
-                'folder': superfolder_path,
-                'name': folder_name,
-                'is_folder': True})
+            needles = {'is_folder': True}
+            for name, folder_to_index in folder_indexers:
+                needles.update(dict((folder_to_index(name, tree, folder)).needles()))
+            es.index(index, FILE, needles)
+>>>>>>> 639bf4c... Add a descriptor plugin that emulates MXR's descriptor column.
 
 
 def index_files(tree, tree_indexers, index, pool, es):
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -378,7 +378,7 @@ def unicode_contents(path, encoding_guess):  # TODO: Make accessible to TreeToIn
 
 
 def unignored(folder, ignore_paths, ignore_filenames, want_folders=False):
-    """Return an iterable of UTF-8 encoded absolute paths to unignored source
+    """Return an iterable of Unicode absolute paths to unignored source
     tree files or the folders that contain them. Skip any non-utf8-decodeable
     paths.
 
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -602,7 +602,7 @@ def index_folders(tree, index, es):
                 continue
             needles = {'is_folder': True}
             for name, folder_to_index in folder_indexers:
-                needles.update(dict(folder_to_index(name, tree, folder)).needles())
+                needles.update(dict(folder_to_index(name, tree, folder).needles()))
             es.index(index, FILE, needles)
 
 
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -414,7 +414,7 @@ def unignored(folder, ignore_paths, ignore_filenames, want_folders=False):
                 try:
                     yield join(root, f).decode('utf-8')
                 except UnicodeDecodeError:
-                    continue
+                    pass
 
         # Exclude folders that match an ignore pattern.
         # os.walk listens to any changes we make in `folders`.
@@ -425,7 +425,7 @@ def unignored(folder, ignore_paths, ignore_filenames, want_folders=False):
                 try:
                     yield join(root, f).decode('utf-8')
                 except UnicodeDecodeError:
-                    continue
+                    pass
 
 def index_file(tree, tree_indexers, path, es, index):
     """Index a single file into ES, and build a static HTML representation of it.
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -378,8 +378,9 @@ def unicode_contents(path, encoding_guess):  # TODO: Make accessible to TreeToIn
 
 
 def unignored(folder, ignore_paths, ignore_filenames, want_folders=False):
-    """Return an iterable of absolute paths to unignored source tree files or
-    the folders that contain them.
+    """Return an iterable of UTF-8 encoded absolute paths to unignored source
+    tree files or the folders that contain them. Skip any non-utf8-decodeable
+    paths.
 
     Returned files include both binary and text ones.
 
@@ -410,7 +411,10 @@ def unignored(folder, ignore_paths, ignore_filenames, want_folders=False):
                 if any(fnmatchcase("/" + path.replace(os.sep, "/"), e) for e in ignore_paths):
                     continue  # Ignore the file.
 
-                yield join(root, f)
+                try:
+                    yield join(root, f).decode('utf-8')
+                except UnicodeDecodeError:
+                    continue
 
         # Exclude folders that match an ignore pattern.
         # os.walk listens to any changes we make in `folders`.
@@ -418,7 +422,10 @@ def unignored(folder, ignore_paths, ignore_filenames, want_folders=False):
             folders, rel_path, ignore_filenames, ignore_paths)
         if want_folders:
             for f in folders:
-                yield join(root, f)
+                try:
+                    yield join(root, f).decode('utf-8')
+                except UnicodeDecodeError:
+                    continue
 
 def index_file(tree, tree_indexers, path, es, index):
     """Index a single file into ES, and build a static HTML representation of it.
@@ -446,10 +453,6 @@ def index_file(tree, tree_indexers, path, es, index):
     # Just like index_folders, if the path is not in UTF-8, then elasticsearch
     # will not accept the path, so just move on.
     rel_path = relpath(path, tree.source_folder)
-    try:
-        rel_path = rel_path.decode('utf-8')
-    except UnicodeDecodeError:
-        return
     is_text = isinstance(contents, unicode)
     is_link = islink(path)
     # Index by line if the contents are text and the path is not a symlink.
@@ -572,7 +575,7 @@ def index_chunk(tree,
                        open_log(tree.log_folder,
                                 'index-chunk-%s.log' % worker_number))
                 for path in paths:
-                    log and log.write('Starting %s.\n' % path)
+                    log and log.write('Starting %s.\n' % path.encode('utf-8'))
                     index_file(tree, tree_indexers, path, es, index)
                 log and log.write('Finished chunk.\n')
             finally:
@@ -596,10 +599,6 @@ def index_folders(tree, index, es):
                      show_eta=False,  # never even close
                      label='Indexing folders') as folders:
         for folder in folders:
-            try:
-                folder = folder.decode('utf-8')
-            except UnicodeDecodeError:
-                continue
             needles = {'is_folder': True}
             for name, folder_to_index in folder_indexers:
                 needles.update(dict(folder_to_index(name, tree, folder).needles()))
@@ -637,7 +636,7 @@ def index_files(tree, tree_indexers, index, pool, es):
             result = future.result()
             if result:
                 formatted_tb, type, value, path = result
-                print 'A worker failed while indexing %s:' % path
+                print 'A worker failed while indexing %s:' % path.encode('utf-8')
                 print formatted_tb
                 # Abort everything if anything fails:
                 raise type, value  # exits with non-zero
--- a/dxr/indexers.py
+++ b/dxr/indexers.py
@@ -65,6 +65,21 @@ class PluginConfig(object):
         return getattr(self.tree, self.plugin_name)
 
 
+class FolderToIndex(PluginConfig):
+    """The FolderToIndex generates needles for folders, and provides an
+    optional list of headers to display in browse view as `browse_headers`.
+    """
+    browse_headers = []
+
+    def __init__(self, plugin_name, tree, path):
+        self.plugin_name = plugin_name
+        self.tree = tree
+        self.path = path
+
+    def needles(self):
+        return []
+
+
 class TreeToIndex(PluginConfig):
     """A TreeToIndex performs build environment setup and teardown and serves
     as a repository for scratch data that should persist across an entire
--- a/dxr/indexers.py
+++ b/dxr/indexers.py
@@ -66,7 +66,7 @@ class PluginConfig(object):
 
 
 class FolderToIndex(PluginConfig):
-    """The FolderToIndex generates needles for folders, and provides an
+    """The FolderToIndex generates needles for folders and provides an
     optional list of headers to display in browse view as `browse_headers`.
     """
     browse_headers = []
--- a/dxr/plugins/descriptor/__init__.py
+++ b/dxr/plugins/descriptor/__init__.py
@@ -10,10 +10,15 @@ from warnings import warn
 import dxr.indexers
 
 
+def is_readme(filename):
+    """Return whether filename is probably a readme."""
+    return filename.lower() in {'readme', 'readme.md', 'readme.txt'}
+
+
 def describe_readme(lines):
-    """Return a string that represents a description for the given lines of a
-    presumed readme file or None if it can extract no suitable description.
-    """
+    """Return a string that represents a description for the given lines
+    of a presumed readme file or None if it can extract no suitable
+    description."""
     # For now the heuristic is just the first non-empty line.
     return next(ifilter(None, (line.strip() for line in lines)), None)
 
@@ -35,9 +40,13 @@ class FolderToIndex(dxr.indexers.FolderToIndex):
             path = join(self.path, entry)
             # If we find a readme, then open it and return the first line if
             # it's non-empty.
-            if "readme" in entry.lower() and isfile(path):
+            if is_readme(entry) and isfile(path):
                 with open(path) as readme:
-                    description = describe_readme([readme.readline(100)])
+                    try:
+                        first_line = readme.readline(100).decode(self.tree.source_encoding)
+                    except UnicodeDecodeError:
+                        continue
+                    description = describe_readme([first_line])
                     if description:
                         # Pack into a list for consistency with the file needle.
                         return [("Description", [description])]
@@ -108,8 +117,7 @@ class FileToIndex(dxr.indexers.FileToIndex):
         readme, then return the first line."""
 
         filename = basename(self.path)
-        # Is it a readme? Just return the first non-empty line.
-        if "readme" in filename.lower():
+        if is_readme(filename):
             possible_description = describe_readme(self.sixty_lines)
             if possible_description:
                 return possible_description
--- a/dxr/plugins/descriptor/__init__.py
+++ b/dxr/plugins/descriptor/__init__.py
@@ -58,8 +58,8 @@ class FileToIndex(dxr.indexers.FileToIndex):
     """Do lots of work to yield a description needle."""
 
     comment_re = re.compile(r'(?:.*?/\*+)(?:\s*\*?\s*)(?P<description>.*?)(?:(?:\*+/)|(?:$))', flags=re.S)
-    docstring_res = [re.compile(r'"""\s*(?P<description>[^"]*)', flags=re.M),
-                     re.compile(r"'''\s*(?P<description>[^']*)", flags=re.M)]
+    docstring_res = [re.compile(r'"""\s*(?P<description>[^"]*)"""', flags=re.M),
+                     re.compile(r"'''\s*(?P<description>[^']*)'''", flags=re.M)]
     title_re = re.compile(r'<title>([^<]*)</title>')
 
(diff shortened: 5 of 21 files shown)


dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+        """Return the contents of the first line of the first docstring if
+        there is one in the first 60 lines."""
+
+        sixty_lines = self.contents.splitlines()[:60]
+        match = re.search(self.docstring_re, '\n'.join(sixty_lines))
+        if match:
+            return match.group('description')
+
+    def generic_describe(self):
+        """Look at the first 60 lines for a match for
```
erikrose: Crazy-narrow wrapping

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+        {{self.path|description}} [delimiter] text,
+        and return the first text we find.
+        Unless it's a readme, then return the first line."""
+
+        sixty_lines = self.contents.splitlines()[:60]
```
erikrose: Might as well splitlines(True) for consistency with elsewhere in the system. It probably doesn't matter, but maybe there's some crazy way in which it does.

dxr/plugins/descriptor/tests/test_descriptions/test_descriptions.py
```diff
@@ -0,0 +1,37 @@
...
+        """Test that the expected descriptions appear on the root page.
+        """
+
+        response = self.client().get('/code/source/').data
+        ok_("Highly important subfolder. with a description over 100 bbytesbytesbytesbytesbytesbytesbytesbytesbyt" in response)
+        # Make sure we didn't exceed the limit.
+        ok_("Highly important subfolder. with a description over 100 bbytesbytesbytesbytesbytesbytesbytesbytesbyte" not in response)
+        ok_("A line of the readme, with prepended spaces." in response)
+        # Check that I stripped prepended spaces.
+        ok_("  A line of the readme, with prepended spaces." in response)
```
erikrose: Do you mean `not in`?

dxr/templates/folder.html
```diff
@@ -27,14 +27,20 @@
     <thead>
       <tr>
         <th scope="col">Name</th>
+        {% for header in plugin_headers %}
+          <th scope="col">{{ header }}</th>
+        {% endfor %}
         <th scope="col">Modified</th>
```
erikrose: Just so you know, "Modified" is kind of useless atm, since it reflects age on disk and not age in the VCS.

erikrose:
I'm very excited about this. I'll give it a proper review later.

dxr/plugins/descriptor/tests/test_descriptions/test_descriptions.py
pelmers: Yes I did, and now I should figure out why it passed anyway...

pelmers: Aha, forgot to call strip() in the plugin

dxr/templates/folder.html
pelmers: right, and just as note to self, https://bugzilla.mozilla.org/show_bug.cgi?id=1189576

dxr/app.py
```diff
@@ -307,6 +307,14 @@ def browse(tree, path=''):
                             frozen['generated_date'])
 
 
+def concat_plugin_headers(plugin_list):
+    """Return a list of the concatenation of all browse_headers in the
+    FolderToIndexes of given plugin list.
+    """
```
erikrose: For consistency with the rest of the code, there should be a skipped blank line before the closing """.

dxr/app.py
```diff
@@ -307,6 +307,14 @@ def browse(tree, path=''):
                             frozen['generated_date'])
 
 
+def concat_plugin_headers(plugin_list):
```
erikrose: I was going to say something about ["nouning" this procedure name](https://grinchcentral.com/function-names-to-verb-or-not-to-verb), but it's already a mess in app.py, so there's no clear convention. :-(

dxr/build.py
```diff
@@ -581,20 +581,19 @@ def index_chunk(tree,
...
-            rel_path = relpath(folder, tree.source_folder)
-            superfolder_path, folder_name = split(rel_path)
-            es.index(index, FILE, {
-                'path': [rel_path],  # array for consistency with non-folder file docs
-                'folder': superfolder_path,
-                'name': folder_name,
-                'is_folder': True})
+            needles = {'is_folder': True}
+            for name, folder_to_index in folder_indexers:
+                needles.update(dict((folder_to_index(name, tree, folder)).needles()))
```
erikrose: unneeded parens

dxr/indexers.py
```diff
@@ -67,6 +67,21 @@ def plugin_config(self):
         return getattr(self.tree, self.plugin_name)
 
 
+class FolderToIndex(PluginConfig):
+    """The FolderToIndex generates needles for folders, and provides an
```
erikrose: No comma before "and", since this isn't a compound sentence

dxr/build.py
```diff
@@ -581,20 +581,19 @@ def index_chunk(tree,
...
                                        tree.ignore_paths,
                                        tree.ignore_filenames,
                                        want_folders=True),
                      show_eta=False,  # never even close
                      label='Indexing folders') as folders:
         for folder in folders:
-            rel_path = relpath(folder, tree.source_folder)
-            superfolder_path, folder_name = split(rel_path)
-            es.index(index, FILE, {
-                'path': [rel_path],  # array for consistency with non-folder file docs
```
erikrose: You're a good person for pulling this all out into a core plugin. :-)

dxr/indexers.py
```diff
@@ -67,6 +67,21 @@ def plugin_config(self):
...
+    optional list of headers to display in browse view as `browse_headers`.
+    """
+    browse_headers = []
```
erikrose: good name

dxr/plugins/descriptor/tests/test_descriptions/test_descriptions.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+from nose.tools import ok_
+
+from dxr.testing import DxrInstanceTestCase
+
+
+class DescriptionTests(DxrInstanceTestCase):
+    def test_browse_page(self):
+        """Test that the expected descriptions appear on the root page.
+        """
```
erikrose: If you're going to skip a line, it goes inside the quotes (for some odd historical emacs reason in PEP0008 that has since been deleted, but it's consistent with the rest of the codebase).

dxr/plugins/descriptor/tests/test_descriptions/test_descriptions.py
```diff
@@ -0,0 +1,37 @@
...
+
+        response = self.client().get('/code/source/').data
+        ok_("Highly important subfolder. with a description over 100 bbytesbytesbytesbytesbytesbytesbytesbytesbyt" in response)
+        # Make sure we didn't exceed the limit.
+        ok_("Highly important subfolder. with a description over 100 bbytesbytesbytesbytesbytesbytesbytesbytesbyte" not in response)
+        ok_("A line of the readme, with prepended spaces." in response)
+        # Check that I stripped prepended spaces.
+        ok_("  A line of the readme, with prepended spaces." not in response)
+        # Javascript description with /** */ comment
+        ok_("Define foo, a dynamic higher order weakly typed late binded function." in response)
```
erikrose: I love that we haven't had to start parsing HTML in our tests yet. ;-)

dxr/plugins/descriptor/tests/test_descriptions/code/foo.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Define foo, a dynamic higher order weakly typed late binded function.
```
erikrose: Ha. :-)

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
+"""Descriptor - Provide useful descriptions for common file types.
+Refer to https://mxr.mozilla.org/webtools-central/source/mxr/Local.pm#27
+"""
+import re
```
erikrose: TLDR LGTM JK

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+import dxr.indexers
+
+
+class FolderToIndex(dxr.indexers.FolderToIndex):
+    browse_headers = ['Description']
+
+    def needles(self):
+        """If the folder contains a readme, then yield the first line of the
+        readme as the description.
+        Similar to https://mxr.mozilla.org/webtools-central/source/mxr/Local.pm#251
```
erikrose: Skip a line between paragraphs.

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        for entry in listdir(self.path):
```
erikrose: Fine for now. Maybe someday we can get the FS to do globbing for us. Python's `glob` lib is infamously slow.

erikrose: We should sort this, as the docs say it comes back in arbitrary order, and it'd be a shame to be nondeterministic.

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+        match = re.search(self.title_re, self.contents)
+        if match:
+            return match.group(1)
+
+    def describe_py(self):
+        """Return the contents of the first line of the first docstring if
+        there is one in the first 60 lines."""
+
+        sixty_lines = self.contents.splitlines(True)[:60]
+        match = re.search(self.docstring_re, ''.join(sixty_lines))
```
erikrose: '', not '\n'?

pelmers: splitlines(True) kept the linebreaks, so joining the lines brings them back.

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+                        return [("Description", [first_line])]
+        # Didn't find anything to use as a description
+        return []
+
+
+class FileToIndex(dxr.indexers.FileToIndex):
+    """Do lots of work to yield a description needle."""
+
+    comment_re = re.compile(r'(?:.*?/\*+)(?:\s*\*?\s*)(?P<description>.*?)(?:(?:\*+/.*)|(?:$))', flags=re.M)
+    docstring_re = re.compile(r'(\'\'\'|""")(?:\s*)(?P<description>.*?)(?:(?:(\'\'\'|"""))|(?:$))', flags=re.M)
```
erikrose: This will potentially run a lot faster if you avoid the non-greedy quantifiers and somehow say "scan until triple-quotes or EOF". Possible?

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+
+        match = re.search(self.title_re, self.contents)
+        if match:
+            return match.group(1)
+
+    def describe_py(self):
+        """Return the contents of the first line of the first docstring if
+        there is one in the first 60 lines."""
+
+        sixty_lines = self.contents.splitlines(True)[:60]
```
erikrose: Hrrm, so we have to read in and split the whole file just to get at the first 60 lines? Can we pull in the first 10K or something instead? If their lines are 2000 chars wide or something, there's probably not a good description after line 1 anyway.

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+        match = re.search(self.docstring_re, ''.join(sixty_lines))
+        if match:
+            return match.group('description')
+
+    def generic_describe(self):
+        """Look at the first 60 lines for a match for {{self.path|description}}
+        [delimiter] text, and return the first text we find. Unless it's a
+        readme, then return the first line."""
+
+        sixty_lines = self.contents.splitlines(True)[:60]
```
erikrose: Again, let's use the cached split.

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+        filename = basename(self.path)
+        # Is it a readme? Just return the first non-empty line.
+        if "readme" in filename.lower():
```
erikrose: Would it makes sense to share this with the FolderToIndex's extractor?

dxr/plugins/descriptor/__init__.py
erikrose: The first group doesn't need to be capturing, does it?

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+            for line in sixty_lines:
+                if line:
+                    return line
+        # Not a readme file, try to match the filename: description pattern.
+        root, ext = splitext(filename)
+        delimiters = ':,-'
+        try:
+            description_re = re.compile(r'({}|{}|description)\
```
erikrose: Most of these groups probably don't need to be capturing.

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+                                         ({})?\s*([{}]\n?)\s*\
+                                         (?P<description>[\w\s-]+)'.format(self.path,
+                                                                           root,
+                                                                           ext,
+                                                                           delimiters),
+                                        re.IGNORECASE)
+            for line in sixty_lines:
+                match = re.search(description_re, line)
```
erikrose: I mean, you might as well call the search method on description_re.

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+                if match:
+                    return match.group('description')
+        except re.error:
+            warn("Error on compiling or search regexp for {}".format(self.path))
+
+        # Haven't returned so we can fall back to the first non-empty line of
+        # the first doc-comment.
+        # TODO: skip the license
+        match = re.search(self.comment_re, ''.join(sixty_lines))
```
erikrose: likewise

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+        delimiters = ':,-'
+        try:
+            description_re = re.compile(r'({}|{}|description)\
+                                         ({})?\s*([{}]\n?)\s*\
+                                         (?P<description>[\w\s-]+)'.format(self.path,
+                                                                           root,
+                                                                           ext,
+                                                                           delimiters),
+                                        re.IGNORECASE)
+            for line in sixty_lines:
```
erikrose: So what's the obsession with 60 lines (presumably lifted from MXR)? Do we have some leading license block that's 59 lines long?

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+                match = re.search(description_re, line)
+                if match:
+                    return match.group('description')
+        except re.error:
+            warn("Error on compiling or search regexp for {}".format(self.path))
```
erikrose: There's a (kinda) nice (but overkill) escaper for regexes in re.escape. If you escape the things you're subbing in, the compiler should never crash.

dxr/plugins/descriptor/__init__.py
erikrose: '', not '\n'?

dxr/templates/folder.html
```diff
@@ -27,14 +27,20 @@
...
         <th scope="col">Size</th>
       </tr>
     </thead>
     <tbody>
-      {% for icon, name, modified, size, relative_url in files_and_folders %}
+      {% for icon, name, modified, size, plugin_columns, relative_url in files_and_folders %}
         <tr>
           <td><a href="{{ relative_url }}" class="icon {{ icon }}">{{ name }}</a></td>
+          {% for col in plugin_columns %}
+            <td scope="col">{{ col }}</th>
```
erikrose: So this one gets a `scope` but no others?

erikrose:
That's all I've got!

dxr/plugins/descriptor/__init__.py
erikrose: good point

dxr/plugins/descriptor/__init__.py
pelmers: mxr/Local.pm:
"# Read in the beginning of the file into a string. I chose 60 because the
# Berkeley copyright notice is around 40 lines long so we need a bit more
# than this."

pelmers:
Thanks for the reviewing!

dxr/templates/folder.html
pelmers: Nope, probably a copey paste, thanks for catching!

dxr/plugins/descriptor/tests/test_descriptions/test_descriptions.py
pelmers: I think you made an identical comment last year :D

pelmers:
self-notes on todo that the commit collpased:
1) optimize docstring regex
2) factor readme description pull?

dxr/plugins/descriptor/__init__.py
erikrose: Actually, we perhaps should sort this everywhere we call listdir, if we aren't sorting it on the way out.

dxr/plugins/descriptor/tests/test_descriptions/test_descriptions.py
erikrose: Still true! And consistent!

dxr/plugins/descriptor/__init__.py
```diff
@@ -9,6 +9,16 @@
 import dxr.indexers
 
 
+def describe_readme(lines):
+    """Return a string that represents a description for the given lines of a
+    presumed readme file or None if it can extract no suitable description.
+    """
+    # For now the heuristic is just the first non-empty line.
+    for candidate in (line.strip() for line in lines):
```
erikrose: Or even `return first(ifilter(None, (line.strip() for line in lines)))`, using more_itertools.first.

pelmers: Oh, perfect

dxr/plugins/descriptor/__init__.py
pelmers: I puzzled my head against the avoiding non-greedy quantifiers, but I'm not sure how else to say "scan until 3 quotes in a row".

erikrose: You can just scan until anything but a quote. That'll greatly reduce the max length of the rewind in most cases.

erikrose:
Ah yes: don't forget to document the new hooks in development.rst.

pelmers:
Updated, that should be everything so far, I'll make a note in the title.

dxr/plugins/descriptor/__init__.py
```diff
@@ -9,6 +10,17 @@
...
 
 
+def describe_readme(lines):
+    """Return a string that represents a description for the given lines of a
+    presumed readme file or None if it can extract no suitable description.
+    """
+    # For now the heuristic is just the first non-empty line.
+    try:
+        return next(ifilter(None, (line.strip() for line in lines)))
+    except StopIteration:
```
erikrose: Instead, you can just pass None as the 2nd arg of `next()`.

docs/source/development.rst
```diff
@@ -248,8 +249,14 @@ manually:
     .. autoclass:: dxr.plugins.Plugin
        :members:
 
-Actual plugin functionality is implemented within tree indexers, file
-indexers, filters, and skimmers.
+Actual plugin functionality is implemented within file indexers, tree indexers,
```
erikrose: File indexers and file indexers, eh? :-)

dxr/plugins/descriptor/__init__.py
```diff
@@ -48,9 +48,24 @@ class FileToIndex(dxr.indexers.FileToIndex):
     """Do lots of work to yield a description needle."""
 
     comment_re = re.compile(r'(?:.*?/\*+)(?:\s*\*?\s*)(?P<description>.*?)(?:(?:\*+/.*)|(?:$))', flags=re.M)
-    docstring_re = re.compile(r'(?:\'\'\'|""")(?:\s*)(?P<description>.*?)(?:(?:(\'\'\'|"""))|(?:$))', flags=re.M)
+    docstring_re1 = re.compile(r'"""\s*(?P<description>[^"]*)', flags=re.M)
```
erikrose: Why not fold these into 1 regex with a | between them, simplifying the code below?

dxr/plugins/descriptor/__init__.py
pelmers: Ah, didn't realize it took a second arg!

erikrose: Me neither at first. I wrote more_itertools.first() to fill that niche before somebody filled me in.

dxr/plugins/descriptor/__init__.py
pelmers: The main reason is that if I put the or around the single/double quote parts then I would not necessarily be matching them (i.e. ''' docstrings would terminate on seeing a ") and if I put the or around the entire thing I would need two capture names and still have to check for both cases later. Also I don't have to escape many quotes this way.

erikrose: I tried it, and you're right: you can't repeat a named group on opposite sides of an alternation. Surprising! We could use backreferences…

```python
>>> re.search(r'(foo|bar)A\1', 'barAbar').groups()
('bar',)
>>> re.search(r'(foo|bar)A\1', 'fooAfoo').groups()
('foo',)
```
…to choose the right quotes, but it would get hard to read. Let's go with your way but use a `for` loop so we don't have to worry about having/testing both branches of code.

erikrose:
r+ after adding that for loop (or whatever DRY you like) around the docstring regex testing.

erikrose:
People are going to be excited about this. :-)

dxr/plugins/descriptor/__init__.py
```diff
@@ -107,7 +107,8 @@ def generic_describe(self):
         [delimiter] text, and return the first text we find. Unless it's a
         readme, then return the first line."""
 
-        filename = basename(self.path)
+        utf8_path = self.path.encode('utf-8')
```
erikrose: I do want to review this again before you merge, since this is substantial stuff. Just let me know when.

pelmers: This undoes the decode in build.py, we shouldn't be reassigning there (note to self)

dxr/build.py
```diff
@@ -447,7 +447,7 @@ def index_file(tree, tree_indexers, path, es, index):
     # will not accept the path, so just move on.
     rel_path = relpath(path, tree.source_folder)
     try:
-        rel_path = rel_path.decode('utf-8')
+        rel_path = unicode(rel_path, 'utf-8')
```
erikrose: We know rel_path to be a str, so wouldn't it be simpler to call rel_path.decode('utf-8')? `unicode()` brings in the complexity of other exception types: the docs say it will throw ValueError if decoding fails (and it's in strict error-handling mode).

But actually, I think we should move Unicode decoding into the config layer so all paths are unicode everywhere in the system. Then we wouldn't have to deal with it here. What do you think? If we do go that way, we need to decide whether it needs (or is worth) a format version bump. It may not need one, since everything comes back from ES as unicode IIRC.

pelmers: What do you mean by moving to the config layer (the Tree/FileToIndexers?)

pelmers: Also I agree, this shouldn't require any format bump.

pelmers: Hmm, looking at the common points, it seems that all the paths come from unignored at some point, so we could toss the decode there, and ignore anything that won't decode.

erikrose: By "config layer", I meant config.py. In general, it minimizes the opportunity for differing representations if we convert everything to unicode at the earliest possible point after it enters the system.

pelmers: The paths arrive from the os.walk in unignored, so I thought that's the earliest point I can convert them.

erikrose: If you pass unicode into os.walk, does it spit unicode back out, or not?

pelmers: It does! At least locally.
```
>>> list(walk('哈'))
[('\xe5\x93\x88', [], ['\xe4\xbb\x96'])]
>>> list(walk(u'哈'))
[(u'\u54c8', [], [u'\u4ed6'])]
```

erikrose: Then perhaps we can do it all at once, in the config layer. :-)

pelmers:
I reverted back to the state of this morning/last night (one decode call for files, one for folders), and put on my todo the item of unifying them under unignored(), rather than try to potentially rewrite os.walk and ride that along with the description plugin.

erikrose:
All you'd really have to do is write an iterator that wraps walk() and utf-8 decodes each item as it's yielded.

pelmers:
Alright, I think the whack-a-mole is done. Since unignored() is already a wrapper for os.walk, I have it yield only unicode paths, so the input to indexers is now unicode as well.

dxr/plugins/descriptor/__init__.py
```diff
@@ -45,8 +49,8 @@ class FileToIndex(dxr.indexers.FileToIndex):
     """Do lots of work to yield a description needle."""
 
     comment_re = re.compile(r'(?:.*?/\*+)(?:\s*\*?\s*)(?P<description>.*?)(?:(?:\*+/.*)|(?:$))', flags=re.M)
-    docstring_re1 = re.compile(r'"""\s*(?P<description>[^"]*)', flags=re.M)
-    docstring_re2 = re.compile(r"'''\s*(?P<description>[^']*)", flags=re.M)
+    docstring_res = [re.compile(r'"""\s*(?P<description>[^"]*)', flags=re.M),
+                     re.compile(r"'''\s*(?P<description>[^']*)", flags=re.M)]
     title_re = re.compile(r'<title>([^<]*)</title>')
```
erikrose: I guess the title tag could have attrs, but we can come and fix that if it ever matters.

dxr/build.py
```diff
@@ -378,8 +378,9 @@ def unicode_contents(path, encoding_guess):  # TODO: Make accessible to TreeToIn
 
 
 def unignored(folder, ignore_paths, ignore_filenames, want_folders=False):
-    """Return an iterable of absolute paths to unignored source tree files or
-    the folders that contain them.
+    """Return an iterable of UTF-8 encoded absolute paths to unignored source
```
erikrose: "Return an iterable of Unicode absolute paths to unignored source files or folders. ..."

dxr/build.py
```diff
@@ -410,15 +411,21 @@ def raise_(exc):
                 if any(fnmatchcase("/" + path.replace(os.sep, "/"), e) for e in ignore_paths):
                     continue  # Ignore the file.
 
-                yield join(root, f)
+                try:
+                    yield join(root, f).decode('utf-8')
+                except UnicodeDecodeError:
+                    continue
```
erikrose: `pass` would suffice here. I misread the `continue` as jumping back to line 394, which had me concerned about skipping the folders[:] thing just below. Oops.

dxr/build.py
```diff
@@ -410,15 +411,21 @@ def raise_(exc):
...
         # os.walk listens to any changes we make in `folders`.
         folders[:] = _unignored_folders(
             folders, rel_path, ignore_filenames, ignore_paths)
         if want_folders:
             for f in folders:
-                yield join(root, f)
+                try:
+                    yield join(root, f).decode('utf-8')
+                except UnicodeDecodeError:
+                    continue
```
erikrose: here too

dxr/build.py
```diff
@@ -573,7 +576,7 @@ def index_chunk(tree,
                        open_log(tree.log_folder,
                                 'index-chunk-%s.log' % worker_number))
                 for path in paths:
-                    log and log.write('Starting %s.\n' % path)
+                    log and log.write('Starting %s.\n' % path.encode('utf-8'))
```
erikrose: Did you actually run into a problem just passing unicode?

dxr/build.py
```diff
@@ -638,7 +637,7 @@ def path_chunks(tree):
             result = future.result()
             if result:
                 formatted_tb, type, value, path = result
-                print 'A worker failed while indexing %s:' % path
+                print 'A worker failed while indexing %s:' % path.encode('utf-8')
```
erikrose: and here?

tests/test_non_ascii_paths/makefile
```diff
@@ -1,7 +0,0 @@
-all:
```
erikrose: Can you help me understand why this went away?

dxr/plugins/descriptor/__init__.py
```diff
@@ -119,9 +119,9 @@ def generic_describe(self):
         try:
             description_re = re.compile(r'(?:{}|{}|description)\
```
erikrose: I don't think you can just throw line breaks and other whitespace into the middle of regexes and have it ignored, unless you set the re.X flag.

dxr/plugins/descriptor/__init__.py
```diff
@@ -119,9 +119,9 @@ def generic_describe(self):
...
                                           (?:{})?\s*(?:[{}]\n?)\s*\
-                                          (?P<description>[\w\s-]+)'.format(re.escape(self.path),
-                                                                            re.escape(root),
-                                                                            re.escape(ext),
+                                          (?P<description>[\w\s-]+)'.format(re.escape(self.path.encode('utf-8')),
```
erikrose: Why go through special trouble to encode everything back to strs? Why not let the regex itself be unicode? Then it will actually match any applicable unicode pathnames from the unicode `self.contents`.

Witness:
```python
>>> re.compile(u'füd'.encode('utf-8')).match(u'füd') is None
True
>>> re.compile(u'füd').match(u'füd')
<_sre.SRE_Match object at 0x107bee510>
```

dxr/build.py
pelmers: Yes, using %s will convert its argument via str(), which defaults to ascii.

erikrose: Not if the string it's subbing into is also unicode

tests/test_non_ascii_paths/makefile
pelmers: I'd accidentally duplicated it from https://github.com/mozilla/dxr/blob/master/tests/test_non_ascii_paths/code/makefile, and this file is not run.

dxr/build.py
pelmers: Writing `log and log.write(u'Starting %s.\n' % path)` fails with 'ascii failed to decode ordinal,' see https://travis-ci.org/mozilla/dxr/builds/143472385.

erikrose: It looks like it's actually the `print` statement that failed. I was going to suggest that the decoding error happens when the Python internals try to jam the Unicode output into an ASCII tty, but I see from the Travis logs that LANG=C.UTF-8. Can you try doing the encode just on the print statement and see what happens?

pelmers: Discussed elsewhere, but for the record, yes it still errors.

dxr/plugins/descriptor/__init__.py
```diff
@@ -141,7 +141,10 @@ def generic_describe(self):
...
-        # TODO: skip the license
-        match = self.comment_re.search(''.join(self.sixty_lines))
-        if match:
-            return match.group('description')
+        for match in self.comment_re.finditer(''.join(self.sixty_lines)):
+            desc = match.group('description')
+            desc_lower = desc.lower()
+            # Skip any comment that contains the license or a tab-width
+            # emacs/vim setting.
+            if not any((pattern in desc_lower for pattern in {'tab-width', 'license', 'vim:'})):
```
erikrose: There's an extra set of parens here. (A genexpr doesn't need its own set if it's the only arg to a func.) And there's no sense using a set (and the extra hasing and RAM) because we're just iterating over it.

docs/source/development.rst
```diff
@@ -248,8 +249,14 @@ manually:
...
+folder indexers, filters, and skimmers.
+
+Folder Indexers
+=============
+
+.. autoclass:: dxr.indexers.TreeToIndex
```
erikrose: FolderToIndex

docs/source/development.rst
```diff
@@ -248,8 +249,14 @@ manually:
     .. autoclass:: dxr.plugins.Plugin
        :members:
 
-Actual plugin functionality is implemented within tree indexers, file
-indexers, filters, and skimmers.
+Actual plugin functionality is implemented within file indexers, tree indexers,
+folder indexers, filters, and skimmers.
+
+Folder Indexers
+=============
```
erikrose: too short

dxr/plugins/descriptor/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+                        return [("Description", [description])]
+        # Didn't find anything to use as a description
+        return []
+
+
+class FileToIndex(dxr.indexers.FileToIndex):
+    """Do lots of work to yield a description needle."""
+
+    comment_re = re.compile(r'(?:.*?/\*+)(?:\s*\*?\s*)(?P<description>.*?)(?:(?:\*+/)|(?:$))', flags=re.S)
+    docstring_res = [re.compile(r'"""\s*(?P<description>[^"]*)', flags=re.M),
```
erikrose: When I said…

> You can just scan until anything but a quote. That'll greatly reduce the max length of the rewind in most cases.

…I didn't mean "and then stop". I meant "and then see if it was the first quote of a """.

erikrose:
I just want to see the revised regex before merging.

pelmers:
revised regex: https://github.com/mozilla/dxr/pull/561/commits/35fbef5bc12b7db5f958929ff96ae2abf0e8b8c9#diff-1983f06f8b72881fa49fad67de088e8bR61

erikrose:
r+! Hooray for user-visible changes!

PR #562, opened 2016-06-09
Find overrides/overriddens from different translation units.  Fixes b…

kleintom:
…ug 1261534.


dxr/plugins/clang/dxr-index.cpp
```diff
@@ -600,21 +600,15 @@ class IndexConsumer : public ASTConsumer,
                methodDecl->begin_overridden_methods(),
                end = methodDecl->end_overridden_methods();
              iter != end; ++iter) {
-          const FunctionDecl *overriddenDecl = nullptr;
-          // Get the overridden definition if it exists...
-          (*iter)->isDefined(overriddenDecl);
```
erikrose: Because we got rid of isDefined (which scribbles on its arg, IIRC), are we going to start finding declarations rather than definitions?

kleintom: No - we only lay down overridden/overrides needles on 'function' lines: https://github.com/mozilla/dxr/blob/master/dxr/plugins/clang/needles.py#L211 and those only come from defs (well, strictly speaking that includes pure virtuals as well).

erikrose: Sold! Thanks!

PR #563, opened 2016-06-09
Update blame links with selected line, fix bug 1007293

pelmers:
take 2! This time simplified from #448 (no format bump, mostly a javascript update).

Plugins can put {{line}} in the href of any links, and these will be substituted with the most recently selected line number.


pelmers:
It's all javascript!
r? @mythmon

dxr/static_unhashed/js/code-highlighter.js
```diff
@@ -132,6 +138,26 @@ $(function () {
...
+        navlinks.each(function() {
+            var $this = $(this);
+            // Copy the href to a template attr, since we will be updating it
+            // and don't want to lose the template. This behavior will happen
+            // only the first time this function is called.
+            if (!$this.data('template')) {
+                $this.data('template', $this.attr('href'));
+            }
+            if (lastNumber) {
+                $this.attr('href', $this.data('template').replace(/{{line}}/g, lastNumber));
```
mythmon: Does this work? I know that `{}` are special characters in some regex dialects. Not sure if it is a problem here.

dxr/vcs.py
```diff
@@ -147,7 +147,7 @@ def generate_diff(self, path):
         return self.upstream + 'diff/' + self.previous_revisions[path] + '/' + path
 
     def generate_blame(self, path):
-        return self.upstream + 'annotate/' + self.revision + '/' + path
+        return self.upstream + 'annotate/' + self.revision + '/' + path + '#l{{line}}'
```
mythmon: Probably makes sense to use string formatting here, instead of repeated concatenation.

mythmon:
This seems like a fragile approach, since if the JS screws up you would get bad urls in your links. Maybe have the thing that generates the HTML fill in `data-template` directly with the template, and `href` with a safe fallback? That also simplifies the JS a bit.

pelmers:
Yes it would! I'll do that and ping you again

dxr/static_unhashed/js/code-highlighter.js
pelmers: It did work when I tested it locally. I think curlies are only used in {n} to match n occurrences of the preceding character.

dxr/static_unhashed/js/code-highlighter.js
```diff
@@ -143,17 +143,13 @@ $(function () {
...
-                $this.attr('href', $this.data('template').replace(/{{line}}/g, lastNumber));
-            } else {
-                $this.attr('href', $this.data('template').replace(/{{line}}/g, ''));
+            const $this = $(this);
+            if ($this.data('template')) {
+                if (lastNumber) {
+                    $this.attr('href', $this.data('template').replace(/{{line}}/g, lastNumber));
+                } else {
+                    $this.attr('href', $this.data('template').replace(/{{line}}/g, ''));
+                }
```
mythmon: You can be a bit more concise here like this:

```javascript
$this.attr('href', $this.data('template').replace(/{{line}}/g, lastNumber || ''));
```

pelmers:
@mythmon thanks for the reviews!

dxr/static_unhashed/js/code-highlighter.js
```diff
@@ -327,7 +345,9 @@ $(function () {
     }
 
     // Highlight any lines specified by hash in either a direct page load or a history pop.
-    $(document).ready(processHash);
+    $(document).ready(function() {
```
erikrose: Why wrap this in a function? The terser original form was equivalent, no?

dxr/templates/file.html
```diff
@@ -37,7 +37,11 @@
           <ul>
             {% for item in section['items'] %}
               <li>
-                <a href="{{ item.href }}" title="{{ item.title }}" class="{{ item.icon }} icon">{{ item.title }}</a>
+                {% if 'template' in item %}
```
erikrose: I have a slight preference for DRYing this and putting the `if` only around the template attr.

dxr/indexers.py
```diff
@@ -224,6 +224,9 @@ def links(self):
 
             (sort order, heading, [(icon, title, href), ...])
 
+            File views will replace any {{line}} within the href with the
+            last-selected line number.
```
erikrose: So we scroll to the first-selected line but blame to the last. Any reason for the inconsistency?

pelmers: No, actually when I was writing this I hadn't realized that we scrolled to the first selected line.

erikrose:
My only reservation is that there's a hanging `#L` on any blame URLs that don't have a line number. It shouldn't hurt anything; it just looks shoddy. But it doesn't need to block merging this. r+. (And file a bug about the trailing #L.)

pelmers:
Thanks for review!

PR #564, opened 2016-06-09
Display the last-committed time as modified column when possible.

pelmers:
Fixes bug 1189576

I'm not sure if the tests depend on the location where it's running, it may be better to generate the tests from timestamps instead in that case, i.e. I record the expected timestamp in the test, then the test case constructs the locale-specific format for it at run time.

### DIFF
--- a/dxr/templates/folder.html
+++ b/dxr/templates/folder.html
@@ -35,7 +35,11 @@
       {% for icon, name, modified, size, relative_url in files_and_folders %}
         <tr>
           <td><a href="{{ relative_url }}" class="icon {{ icon }}">{{ name }}</a></td>
-          <td><a href="{{ relative_url }}"><time>{{ '' if modified is none else modified.strftime("%Y %b %d %H:%m") }}</time></a></td>
+          <td><a href="{{ relative_url }}">
+            {% if modified is not none %}
+              <time>{{ '' if modified is none else modified.strftime("%Y %b %d %H:%m") }}</time>
+            {% endif %}
+          </a></td>
           <td><a href="{{ relative_url }}">{{ '' if size is none else size|filesizeformat }}</a></td>
         </tr>
       {% endfor %}
--- a/dxr/templates/folder.html
+++ b/dxr/templates/folder.html
@@ -35,7 +35,7 @@
       {% for icon, name, modified, size, relative_url in files_and_folders %}
         <tr>
           <td><a href="{{ relative_url }}" class="icon {{ icon }}">{{ name }}</a></td>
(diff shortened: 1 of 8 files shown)


pelmers:
Consider using <time> html tag.

dxr/vcs.py
```diff
@@ -75,6 +76,12 @@ def generate_raw(self, path):
         """Construct URL to upstream view to raw file at path."""
         return NotImplemented
 
+    def last_modified_date(self, path):
+        """Return a datetime object that represents the last time a commit was
+        made to the given path.
+        """
+        return NotImplemented
```
pelmers: `raise notimplementederror`

erikrose:
r+ with notimplementederror as exception and fix of testcase to work regardless of timezone

pelmers:
Just realized that this change touches the shape of the file index, since 'modified' is a one-element list rather than a string.
This isn't worth a format bump, right? I can let it accept either the string or the list.

erikrose:
Yes, just let it accept either. But comment that spot, and say that we can remove support for the non-list case when we hit format version X.

dxr/templates/folder.html
```diff
@@ -35,7 +35,7 @@
       {% for icon, name, modified, size, relative_url in files_and_folders %}
         <tr>
           <td><a href="{{ relative_url }}" class="icon {{ icon }}">{{ name }}</a></td>
-          <td><a href="{{ relative_url }}">{{ '' if modified is none else modified.strftime("%Y %b %d %H:%m") }}</a></td>
+          <td><a href="{{ relative_url }}"><time>{{ '' if modified is none else modified.strftime("%Y %b %d %H:%m") }}</time></a></td>
```
erikrose: Probably we shouldn't have an empty time tag here in any case.

erikrose:
r+ with empty time tags gone

dxr/templates/folder.html
```diff
@@ -35,7 +35,11 @@
       {% for icon, name, modified, size, relative_url in files_and_folders %}
         <tr>
           <td><a href="{{ relative_url }}" class="icon {{ icon }}">{{ name }}</a></td>
-          <td><a href="{{ relative_url }}"><time>{{ '' if modified is none else modified.strftime("%Y %b %d %H:%m") }}</time></a></td>
+          <td><a href="{{ relative_url }}">
+            {% if modified is not none %}
+              <time>{{ '' if modified is none else modified.strftime("%Y %b %d %H:%m") }}</time>
```
erikrose: `modified` is not none, so it can't be none. :-)

erikrose:
r+ if tests pass!

PR #565, opened 2016-06-13
Fix scrolling to selected line on page load in Chrome.

abbeyj:
Trying to scroll in the document ready handler in Chrome doesn't work
because Chrome will reset the scroll position later.  Delaying the scroll
with setTimeout works around this problem.

### DIFF
--- a/dxr/static_unhashed/js/code-highlighter.js
+++ b/dxr/static_unhashed/js/code-highlighter.js
@@ -319,6 +319,10 @@ $(function () {
             if (jumpPosition < 0) {
                 jumpPosition = 0;
             }
+
+            // Trying to scroll in the document ready handler doesn't work because some
+            // browsers (e.g. Chrome) will reset the scroll position later.
+            // Delaying the scroll with setTimeout works around this problem.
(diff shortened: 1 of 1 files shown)


dxr/static_unhashed/js/code-highlighter.js
```diff
@@ -316,11 +316,12 @@ $(function () {
...
             //but only scrollTo if the offset is more than 150px in distance from the top of the page
             jumpPosition = parseInt(jumpPosition.top, 10) - 150;
-            if (jumpPosition >= 0) {
-                window.scrollTo(0, jumpPosition);
-            } else {
-                window.scrollTo(0, 0);
+            if (jumpPosition < 0) {
+                jumpPosition = 0;
             }
+            window.setTimeout(function() {
```
pelmers: Let's drop a comment here about why we're putting scrollTo in the setTimeout.

pelmers:
Other than adding that comment I mention, it looks good to me.

erikrose:
Merge it when you guys are ready (in case you weren't going to already)!

PR #569, opened 2016-06-22
Add an XBL plugin for method identifier searches.

pelmers:
This plugin adds qualified needles for the method, property, and field
definitions within bindings. It also adds needles and refs for the
implements attr on the implementation tag.

Fixes bug 771004.

Next in the nice-to-have list would be JS analysis of the CDATA sections
of these files.

### DIFF
--- a/dxr/plugins/xbl/__init__.py
+++ b/dxr/plugins/xbl/__init__.py
@@ -34,8 +34,8 @@ class FileToIndex(dxr.indexers.FileToIndex):
         return self._analyzer
 
     def is_interesting(self):
-        return (self.path.endswith('.xml') and '<bindings' in self.contents
-                and super(FileToIndex, self).is_interesting())
+        return (super(FileToIndex, self).is_interesting and
+                self.path.endswith('.xml') and '<bindings' in self.contents)
 
     def refs(self):
         return self.analyzer.refs
--- a/dxr/plugins/xbl/__init__.py
+++ b/dxr/plugins/xbl/__init__.py
@@ -36,11 +36,11 @@ class FileToIndex(dxr.indexers.FileToIndex):
         return "<bindings" in self.contents and super(FileToIndex, self).is_interesting()
 
     def refs(self):
-        return self.analyzer.refs if self.analyzer else []
+        return self.analyzer.refs
 
     def needles_by_line(self):
         return iterable_per_line(
-            with_start_and_end(split_into_lines(self.analyzer.needles if self.analyzer else [])))
+            with_start_and_end(split_into_lines(self.analyzer.needles)))
 
 
 plugin = Plugin(
--- a/dxr/plugins/xbl/__init__.py
+++ b/dxr/plugins/xbl/__init__.py
@@ -1,0 +1,51 @@
+from functools import partial
+
+from dxr.filters import LINE
+import dxr.indexers
+from dxr.indexers import iterable_per_line, with_start_and_end, split_into_lines, QUALIFIED_LINE_NEEDLE
+from dxr.plugins import Plugin, filters_from_namespace, refs_from_namespace, AdHocTreeToIndex
+from dxr.plugins.xbl import refs, filters
+from dxr.plugins.xbl.analyzer import XBLAnalyzer
+from dxr.plugins.xbl.refs import PLUGIN_NAME
+
+
+mappings = {
+    LINE: {
+        'properties': {
+            # Method/prop/field definitions.
+            PLUGIN_NAME + '_prop': QUALIFIED_LINE_NEEDLE,
+            # Implementations of interfaces.
+            PLUGIN_NAME + '_type': QUALIFIED_LINE_NEEDLE,
+        }
+    }
+}
+
+
+class FileToIndex(dxr.indexers.FileToIndex):
+    def __init__(self, path, contents, plugin_name, tree):
+        super(FileToIndex, self).__init__(path, contents, plugin_name, tree)
+        self._analyzer = None
+
+    @property
+    def analyzer(self):
+        if not self._analyzer:
+            self._analyzer = XBLAnalyzer(self.tree, self.contents, self.tree.source_encoding)
+        return self._analyzer
+
+    def is_interesting(self):
+        return "<bindings" in self.contents and super(FileToIndex, self).is_interesting()
+
+    def refs(self):
+        return self.analyzer.refs if self.analyzer else []
+
+    def needles_by_line(self):
+        return iterable_per_line(
+            with_start_and_end(split_into_lines(self.analyzer.needles if self.analyzer else [])))
+
+
+plugin = Plugin(
+    tree_to_index=partial(AdHocTreeToIndex,
+                          file_to_index_class=FileToIndex),
+    refs=refs_from_namespace(refs.__dict__),
+    filters=filters_from_namespace(filters.__dict__),
+    mappings=mappings)
--- a/dxr/plugins/xbl/__init__.py
+++ b/dxr/plugins/xbl/__init__.py
@@ -29,11 +29,13 @@ class FileToIndex(dxr.indexers.FileToIndex):
     @property
     def analyzer(self):
         if not self._analyzer:
-            self._analyzer = XBLAnalyzer(self.tree, self.contents, self.tree.source_encoding)
+            self._analyzer = XBLAnalyzer(self.path, self.tree,
+                                         self.contents, self.tree.source_encoding)
         return self._analyzer
 
     def is_interesting(self):
-        return "<bindings" in self.contents and super(FileToIndex, self).is_interesting()
+        return (self.path.endswith('.xml') and '<bindings' in self.contents
+                and super(FileToIndex, self).is_interesting())
 
     def refs(self):
         return self.analyzer.refs
--- a/dxr/plugins/xbl/analyzer.py
+++ b/dxr/plugins/xbl/analyzer.py
@@ -20,9 +20,9 @@ class XBLAnalyzer(object):
         self.parser.StartElementHandler = self.StartElementHandler
         self.parser.EndElementHandler = self.EndElementHandler
         try:
-            self.parser.Parse(self.contents)
+            self.parser.Parse(self.contents.encode(encoding))
         except expat.ExpatError as err:
-            warn('Exception occurred on parsing XBL file {}'.format(self.path))
+            warn('Exception occurred on parsing XBL file {}'.format(self.path.encode('utf-8')))
             warn(err.message)
 
     def extent_for_name(self, name):
@@ -30,7 +30,7 @@ class XBLAnalyzer(object):
         # Because the parser does not provide location info on attr positions,
         # we can only search for them from the current position.
         # First check the current line.
-        offset = self.lines[self.parser.CurrentLineNumber - 1].find(name, self.parser.currentColumnNumber - 1)
+        offset = self.lines[self.parser.CurrentLineNumber - 1].find(name, self.parser.CurrentColumnNumber - 1)
         if offset != -1:
             row, col = self.parser.CurrentLineNumber, offset
             found = True
@@ -68,22 +68,24 @@ class XBLAnalyzer(object):
     # Begin definitions of parser event handlers.
 
     def StartElementHandler(self, name, attrs):
-        if 'implementation' == name:
+        name = name.lower()
+        if name == 'implementation' and 'implements' in attrs:
             for impl in attrs['implements'].split(','):
                 impl_name = impl.strip()
                 self.yield_needle('type', impl_name)
                 self.yield_ref(TypeRef(self.tree, impl_name), impl_name)
 
-        elif 'binding' in name:
+        elif name == 'binding':
             # Enter the scope of a binding.
             self.namespace.append(attrs['id'])
 
-        elif 'property' in name or 'method' in name or 'field' in name:
+        elif name in {'property', 'method', 'field'} and 'name' in attrs:
             def_name = attrs['name']
             # Build the qualified name from the most recent binding id.
-            self.yield_needle('prop', def_name, '{}#{}'.format(self.namespace[-1], def_name))
+            self.yield_needle('prop', def_name, u'{}#{}'.format(self.namespace[-1], def_name))
 
     def EndElementHandler(self, name):
-        if 'binding' in name:
+        name = name.lower()
+        if name == 'binding':
             # Left scope of a binding tag, pop the stack.
             self.namespace.pop()
--- a/dxr/plugins/xbl/analyzer.py
+++ b/dxr/plugins/xbl/analyzer.py
@@ -1,3 +1,4 @@
+from itertools import islice
 from warnings import warn
 import xml.parsers.expat as expat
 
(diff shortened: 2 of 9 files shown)


dxr/plugins/xbl/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+        self._analyzer = None
+
+    @property
+    def analyzer(self):
+        if not self._analyzer:
+            self._analyzer = XBLAnalyzer(self.tree, self.contents, self.tree.source_encoding)
+        return self._analyzer
+
+    def is_interesting(self):
+        return "<bindings" in self.contents and super(FileToIndex, self).is_interesting()
```
erikrose: Wow. There's no file extension or anything cheaper we can test?

dxr/plugins/xbl/filters.py
```diff
@@ -0,0 +1,23 @@
...
+class _NameFilter(NameFilterBase):
+    lang = PLUGIN_NAME
+
+
+class _QualifiedNameFilter(QualifiedNameFilterBase):
+    lang = PLUGIN_NAME
+
+
+class PropFilter(_QualifiedNameFilter):
+    name = 'prop'
```
erikrose: We're going to need that language metafilter one day soon. :-)

dxr/plugins/xbl/filters.py
```diff
@@ -0,0 +1,23 @@
...
+    is_identifier = True
+    description = Markup('XBL property definition filter: <code>prop:foo</code>')
```
erikrose: Now we hit the cut corner were only the description from the first plugin for a given filter name is shown. So, in a repo that has JS and XBL enabled, people won't be able to discover that XBL props are supported. For this and to see how the various new MXR trees are indexed, it's probably high time to add plugin (language) badging to the Operators and the Switch Tree menu.

dxr/plugins/xbl/refs.py
```diff
@@ -0,0 +1,25 @@
...
+PLUGIN_NAME = 'xbl'
+
+
+class _XblRef(Ref):
+    plugin = PLUGIN_NAME
+
+
+class TypeRef(_XblRef):
+    """Ref that yields a menu option for search on type and type-decl.
+    """
```
erikrose: More weird docstring quotes, eh? :-)

dxr/plugins/xbl/__init__.py
pelmers: Yes, I can test .xml ending first

dxr/plugins/xbl/filters.py
pelmers: https://bugzilla.mozilla.org/show_bug.cgi?id=983472 !

pelmers: Oh, I dropped that reply on the wrong comment

pelmers: https://bugzilla.mozilla.org/show_bug.cgi?id=1282969

dxr/plugins/xbl/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+    def refs(self):
+        return self.analyzer.refs
+
+    def needles_by_line(self):
+        return iterable_per_line(
+            with_start_and_end(split_into_lines(self.analyzer.needles)))
+
+
+plugin = Plugin(
+    tree_to_index=partial(AdHocTreeToIndex,
```
erikrose: It seems like time to make `mappings` detected automatically, as with FileToIndex and TreeToIndex. We have several plugins that have to resort to this partial(AdHocTreeToIndex) chicanery, which is much more boilerplatey and complex than I intended.

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+            self.parser.Parse(self.contents)
+        except expat.ExpatError:
+            warn("Exception occurred on parsing XBL file {}".format(self.path))
+
+    def extent_for_name(self, name):
+        """Return Extent for the next occurrence of name."""
+        # Because the parser does not provide location info on attr positions,
+        # we can only search for them from the current position.
+        found = False
+        for (number, line) in enumerate(self.lines[self.parser.CurrentLineNumber - 1:]):
```
erikrose: parens not needed

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+
+from dxr.indexers import Extent, Position
+from dxr.plugins.xbl.refs import PLUGIN_NAME, TypeRef
+
+
+class XBLAnalyzer(object):
+    def __init__(self, tree, contents, encoding):
+        self.tree = tree
+        self.contents = contents
+        self.lines = contents.splitlines(True)
```
erikrose: I'm starting to wonder whether we should implement a caching method on FileToIndex or FileToSkim that returns the split lines. A lot of plugins do this. I wonder if we'd save any time. Oh well, this is for another time.

dxr/plugins/xbl/analyzer.py
erikrose: I'm not sure if it would help, but enumerate() does take a second arg that means "start from". Would passing 1 here save any trouble?

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+        except expat.ExpatError:
+            warn("Exception occurred on parsing XBL file {}".format(self.path))
+
+    def extent_for_name(self, name):
+        """Return Extent for the next occurrence of name."""
+        # Because the parser does not provide location info on attr positions,
+        # we can only search for them from the current position.
+        found = False
+        for (number, line) in enumerate(self.lines[self.parser.CurrentLineNumber - 1:]):
+            if name in line:
```
erikrose: We search for name in line twice: once here and again with index() below. Let's not.

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+                row, col = self.parser.CurrentLineNumber + number, line.index(name)
+                found = True
+                break
+        if found:
+            return Extent(Position(row, col), Position(row, col + len(name)))
+        else:
+            warn("Could not find {} from line {}".format(name, self.parser.CurrentLineNumber))
```
erikrose: single quotes

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+        # If qualname is not provided, then use name.
+        mapping = {'name': ident, 'qualname': qualname or ident}
+        extent = self.extent_for_name(ident)
+        if extent:
+            self.needles.append((PLUGIN_NAME + '_' + name, mapping, extent))
+
+    # Begin definitions of parser event handlers.
+
+    def StartElementHandler(self, name, attrs):
+        if "implementation" in name:
```
erikrose: single quotes. We should add that to the linter.

erikrose: Why are we doing `in` here when ISTM like `==` would do?

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+            for impl in attrs['implements'].split(','):
+                name = impl.strip()
```
erikrose: Let's not shadow the arg; it can lead to confusion down the road.

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,78 @@
+from warnings import warn
+import xml.parsers.expat as expat
```
erikrose: Any particular reason you chose expat? I've never seen it used before. OTOH, it doesn't make it hard for us to move to PyPy, unlike lxml. I'm a little nervous that it hasn't been updated since 2012 and it written in C and so probably has security holes, but presumably Python takes some measure of responsibility for it since it's in the stdlib.

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,78 @@
...
+            warn("Exception occurred on parsing XBL file {}".format(self.path))
+
+    def extent_for_name(self, name):
+        """Return Extent for the next occurrence of name."""
+        # Because the parser does not provide location info on attr positions,
+        # we can only search for them from the current position.
+        found = False
+        for (number, line) in enumerate(self.lines[self.parser.CurrentLineNumber - 1:]):
+            if name in line:
+                row, col = self.parser.CurrentLineNumber + number, line.index(name)
```
erikrose: Seems kind of shoddy. Wouldn't a property called "property" find the tag name instead of the attr? To do this right, we'd have to reparse the start tag. I wonder if a different XML parser would return offsets for us.

dxr/plugins/xbl/analyzer.py
pelmers: Expat was the first xml parser in the stdlib I found that lets me query location.

dxr/plugins/xbl/analyzer.py
erikrose: What if the start tag is split across many lines? Could this not look back far enough to catch the attr?

erikrose:
Everything else looks fine to me.

dxr/plugins/xbl/analyzer.py
pelmers: I'm not sure what you mean by splitting the start tag, something like
```
<foo

bar="baz"
>
```
?

dxr/plugins/xbl/analyzer.py
pelmers: Tried adding it to the linter, with the result...
```
$ make lint | grep Q000 | wc -l
534
```

dxr/plugins/xbl/analyzer.py
erikrose: Exactly. Or, even further, having multiple attrs, each on its own line

dxr/plugins/xbl/analyzer.py
pelmers: Yes, but I didn't find any xml parser in the stdlib which gives offsets for tag attributes, and in practice I don't think we'll encounter that situation much.

dxr/plugins/xbl/analyzer.py
pelmers: I'll add this to the test case, and it should work correctly since the given line number is for the start of the tag.

dxr/plugins/xbl/analyzer.py
erikrose: Huh, that's surprising.

dxr/plugins/xbl/analyzer.py
erikrose: I think we need to do a little more—maybe even just a regex. As it stands, something like `<property name="rope">` would find the wrong offset.

erikrose: Also, aren't we assuming that the tag in question is the first one on the line? That's not guaranteed. Can we get ahold of the tag start offset at least and start searching in that column?

dxr/plugins/xbl/__init__.py
```diff
@@ -29,11 +29,13 @@ def __init__(self, path, contents, plugin_name, tree):
...
     def analyzer(self):
         if not self._analyzer:
-            self._analyzer = XBLAnalyzer(self.tree, self.contents, self.tree.source_encoding)
+            self._analyzer = XBLAnalyzer(self.path, self.tree,
+                                         self.contents, self.tree.source_encoding)
         return self._analyzer
 
     def is_interesting(self):
-        return "<bindings" in self.contents and super(FileToIndex, self).is_interesting()
+        return (self.path.endswith('.xml') and '<bindings' in self.contents
```
erikrose: It's academic the way things currently stand (but things change): we should probably put the super() call first, since it's cheapest. It's the thing that will skip binary files if we started reading them in again for some reason and if one ends up having an .xml extension due to changes in our definition of what a binary file is, etc.

dxr/plugins/xbl/analyzer.py
pelmers: Yes, the parser should have both CurrentLine and CurrentColumn, and I can start the search from there.

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,91 @@
...
+        except expat.ExpatError as err:
+            warn('Exception occurred on parsing XBL file {}'.format(self.path.encode('utf-8')))
+            warn(err.message)
+
+    def extent_for_name(self, name):
+        """Return Extent for the next occurrence of name."""
+        # Because the parser does not provide location info on attr positions,
+        # we can only search for them from the current position.
+        # First check the current line.
+        offset = self.lines[self.parser.CurrentLineNumber - 1].find(name, self.parser.CurrentColumnNumber - 1)
```
erikrose: Is CurrentLineNumber the start of the tag or the start of the line (for tags split across lines)?

erikrose: If the start of the line, would it be very hard to scan backward (rfind) until we find the first ">"? (I'm not sure if this would work with CDATA sections, but it's closer.)

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,91 @@
...
+        if offset != -1:
+            row, col = self.parser.CurrentLineNumber, offset
+            found = True
+        else:
+            found = False
+        # Not on the current line, keep going until we find it.
```
erikrose: I gather from this that CurrentLineNumber is the tag-start line. Please correct me if I'm wrong.

pelmers: Yep, CurrentLineNumber is the tag's starting line.

dxr/plugins/xbl/analyzer.py
```diff
@@ -0,0 +1,91 @@
...
+        if not found:
+            for number, line in enumerate(self.lines[self.parser.CurrentLineNumber:]):
```
erikrose: Python 3 might do something clever with a view, but I think CPython 2 physically copies all the line pointers from the current line onward to make a new list.

dxr/plugins/xbl/analyzer.py
erikrose: I'm still not very happy that we'll find names in attr names as well as values, but I'm willing to merge it if we put a big comment here to that effect. Then whoever comes in here chasing a markup bug won't have to figure it out.

pelmers: Agreed, it's possible that the buggy situation could happen; I'll leave a comment and skipped test. Similarly the xpidl plugin [did the same](https://github.com/mozilla/dxr/blob/91cef85924e5ee456350e814e59fd6cfa50013f0/dxr/plugins/xpidl/visitor.py#L22). However in practice the circumstance should be rare, it would require a binding called "binding" or an interface named "implements."

erikrose: r+ with those things. Thanks! :-)

pelmers:
Yes, it should not make much difference at all. Also good tip on the islice notation, it confused me that the second arg of islice changes meaning depending on the total number of args provided.

erikrose:
r+ assuming tests pass

PR #570, opened 2016-06-23
Try to invoke vcs before claiming source.

pelmers:
This should prevent errors that would occur with bad .git or
.hg folders present in the source trees.


erikrose:
r+!

PR #572, opened 2016-06-23
Check that paths can decode in utf8 before sending to ES

pelmers:
There should be a way to index non-utf8 paths, but for now it seems prudent to block them and move on with getting the addon tree up.

Note: The test case does not work on OS X because HFS+ filesystem utf-8 encodes all nodes.

### DIFF
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -444,7 +444,13 @@ def index_file(tree, tree_indexers, path, es, index):
         else:
             raise
 
+    # Just like index_folders, if the path is not in UTF-8, then elasticsearch
+    # will not accept the path, so just move on.
     rel_path = relpath(path, tree.source_folder)
+    try:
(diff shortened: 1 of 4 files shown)


erikrose:
Looks good. But let's have the test raise SkipTest on the Mac so it doesn't appear to succeed.

erikrose:
r+ with that change

pelmers:
I don't think I can detect whether it's on a mac host, since in our dev environment it's an HFS filesystem hoisted into a linux VM.

erikrose:
Good point. We could do something really exotic, but it's not worth the complexity. Ship it.

pelmers:
Ah, it turned out the empty test class was not even executing (nosetest optimization?). I committed a fix for that, will merge if travis accepts.

PR #574, opened 2016-06-24
Make wildcard matches anchored to the ends of the string by default.

abbeyj:
For query terms like path: and file: that allow wildcards, ensure that
they match the entire string.  For example, now file:bar.htm matches
only "bar.htm" and not "bar.html" or "foobar.htm".
If a user wants the old behavior of matching anywhere within the name
they can add an asterisk to the beginning and/or end of their search
term.  For example, file:\*bar.htm\* still matches all of "bar.htm",
"bar.html" and "foobar.htm".


erikrose:
r+ assuming tests pass. The linter just landed, so you might want to `make lint` before you merge this to make sure it gets along with it.

PR #575, opened 2016-06-24
Move VIRTUAL_ENV up in the file, scope lint to dxr/

pelmers:
Fix keeps flake8 from recursing into all subfolders from the root, such as virtualenvs kept with the project.


erikrose:
r+

PR #576, opened 2016-06-25
instruct docker to label the mount with the correct selinux label

db48x:
because otherwise it won't work on on hosts that use selinux


erikrose:
Thanks for the fix!

PR #577, opened 2016-06-27
Increase contrast of text in source code, directory listings, and menus.

erikrose:
Philip Chee was having some trouble seeing on his monitor, and there's no reason for the main content not to have the highest possible contrast.


pelmers:
r+

PR #578, opened 2016-06-27
Bug 1192290 - Factor up splitlines

jay-z007:
refactored the method 
`string.splitlines(True)` with `splitlines_keepends(string)`

the method `splitlines_keepends(string)` is defined in the file **dxr/utils.py**

Appropriate changes were made to 6 files : 

1.  dxr/plugins/python/utils.py
2.  dxr/plugins/xpidl/indexers.py
3.  dxr/plugins/core.py
4.  dxr/indexers.py
5.  dxr/app.py
6.  dxr/build.py


pelmers:
Hi Jay, could you make sure to not commit the build/ directory?

dxr/plugins/core.py
```diff
@@ -20,7 +20,7 @@
 from dxr.plugins import direct_search
 from dxr.trigrammer import (regex_grammar, NGRAM_LENGTH, es_regex_filter,
                             NoTrigrams, PythonRegexVisitor)
-from dxr.utils import glob_to_regex
+from dxr.utils import (glob_to_regex, splitlines_keepends)
```
erikrose: No parentheses are needed here.

dxr/build/code-style.css
```diff
@@ -0,0 +1,33 @@
+/** Syntax Regions
```
erikrose: It looks like you've accidentally checked in your `build` directory.

dxr/utils.py
```diff
@@ -285,3 +285,7 @@ def is_in(needle, haystack):
 def without_ending(ending, string):
     """If ``string`` ends with ``ending``, strip it off."""
     return string[:-len(ending)] if string.endswith(ending) else string
+
+def splitlines_keepends(string):
```
erikrose: Rather than naming this for what it mechanically does, let's name it for its intended use. That way, if we decide to change the mechanics later, it doesn't have to get renamed. Something like `split_content_lines` might be appropriate.

jay-z007: I'll surely make the recommended changes @erikrose 
and will not commit the build/ directory @pelmers

PR #579, opened 2016-06-28
Rust: update the command line flags and env vars

nrc:


pelmers:
Any idea why all the tests failed? Does the rust nightly DXR downloads need to be updated as well?
https://github.com/mozilla/dxr/blob/master/tooling/docker/dev/set_up_common.sh

nrc:
@pelmers oh yeah, it will, I forgot it had been pinned to some revision

nrc:
Hmm, still busted. This looks like we're getting CSV data from rustc, but it is not correct in some way and the Rust DXR plugin is choking.

@pelmers (or anyone else with an interest in Rust and/or DXR) do you fancy having a go at debugging this? I'm afraid I'm not going to have the time to do so and DXR is pretty paged out for me. I expect that one of two things is happening (possibly in multiple places): DXR is expecting something that isn't there any more or has changed (format of ids is probably the first thing I'd check), in which case something will need tweaking in the plugin. Or, something has rotted on the rustc side without us noticing, in which case you just need to disable the test in DXR and either file an issue against Rust or let me know and I'll do that. If you need any help, please ping me on irc.

erikrose:
After the MXR trees are all happy and green in DXR, we can circle back and hit this. For the moment, I've had to disable the rust and servo trees since they were running the build node out of memory.

Issue #571, opened 2016-06-23
SVN support


caugner:
I would like to use DXR in a subversion project.

Any ideas on how to start and add this?
I would be willing to implement this myself.

erikrose:
Sure! You pretty much just need to add an Svn class to vcs.py and then add `Svn` to `every_vcs` later on in that file.

erikrose:
Feel free to join us in #static on irc.mozilla.org if you have questions.

PR #580, opened 2016-06-28
Add SVN support (#571).

caugner:
SVN support: Provide permalink and raw VCS link for subversion repositories.


dxr/vcs.py
```diff
@@ -327,8 +330,125 @@ def get_contents(cls, path, revision, stderr=None):
                                         filename + '@' + revision],
                                        cwd=directory, env=env, stderr=stderr)
 
+class Subversion(Vcs):
+    command = 'svn'
+
+    def __init__(self, root):
+        super(Subversion, self).__init__(root)
+        # All tracked files as a dictionary (file => revision).
+        self.files = __get_files_with_revision()
```
pelmers: needs self in front of the method call

dxr/plugins/omniglot.py
```diff
@@ -27,12 +27,22 @@ def is_interesting(self):
 
     def links(self):
         def items():
-            yield 'log', "Log", self.vcs.generate_log(vcs_relative_path)
-            yield 'blame', "Blame", self.vcs.generate_blame(vcs_relative_path)
-            yield 'diff',  "Diff", self.vcs.generate_diff(vcs_relative_path)
-            yield 'raw', "Raw", self.vcs.generate_raw(vcs_relative_path)
+            log = self.vcs.generate_log(vcs_relative_path)
+            if log:
+                yield 'log', "Log", log
```
pelmers: Not your fault here, but could you make the quotes all single?

dxr/vcs.py
```diff
@@ -327,8 +329,125 @@ def get_contents(cls, path, revision, stderr=None):
                                         filename + '@' + revision],
                                        cwd=directory, env=env, stderr=stderr)
 
+class Subversion(Vcs):
+    command = 'svn'
+
+    def __init__(self, root):
+        super(Subversion, self).__init__(root)
+        # All tracked files as a dictionary (file => revision).
+        self.files = self.__get_files_with_revision()
```
pelmers: get_files_with_revision requires a path parameter

dxr/vcs.py
```diff
@@ -327,8 +329,125 @@ def get_contents(cls, path, revision, stderr=None):
...
+                if not revision:
+                    # Ignore untracked files.
+                    print "%s has no revision!" % abs_path
+                    continue
+
+                files[rel_path] = revision
+
+        return files
+
+    def __get_revision(self, path):
```
pelmers: unused path param

dxr/vcs.py
```diff
@@ -327,8 +329,125 @@ def get_contents(cls, path, revision, stderr=None):
...
+        return Subversion.svn_info(self.root, "Revision")
+
+    def __get_upstream_url(self, path):
```
pelmers: unused path

pelmers:
Could you add an svn test case? That probably involves adding subversion to the set_up_ubuntu script. You can look at test_vcs_hg and test_vcs_git for how to do it (specifically they package their source control folder within a tarball so it doesn't pollute the diff).

caugner:
@pelmers Thanks a lot for your feedback!

dxr/vcs.py
```diff
@@ -327,8 +329,123 @@ def get_contents(cls, path, revision, stderr=None):
...
+    def svn_info(cls, path, prop=None, args=[]):
+        try:
+            info = cls.invoke_vcs(['info'] + args, path)
+
+            if not prop:
+                # Return full info.
+                return info
+
+            # Find and return specific property.
+            regex = re.compile("(?<=%s: )(.+)" % prop)
```
pelmers: Just to be safe, it might make sense to re.escape(prop)

dxr/vcs.py
```diff
@@ -327,8 +329,123 @@ def get_contents(cls, path, revision, stderr=None):
...
+        args = ['cat', '-r', str(revision), os.path.abspath(path)]
+        return cls.invoke_vcs(args, directory)
+
+    # Interface instance methods.
+
+    def is_tracked(self, path):
+        return path in self.files
+
+    def has_upstream(self):
+        return self.upstream != ""
```
pelmers: It looks like on error the upstream constructor could return None, rather than "".

dxr/vcs.py
caugner: Actually, my implementation of `_get_files_with_revision` does not require a path parameter, see here:
https://github.com/mozilla/dxr/pull/580/files#diff-fdc95e419de8a362ed3a4518be94408bR437

dxr/vcs.py
```diff
@@ -17,11 +17,12 @@
 - Check if the mercurial paths are specific to Mozilla's customization or not.
 
 """
-from datetime import datetime
```
caugner: I probably shouldn't have removed this line, because it is the cause for the failing CI.

dxr/vcs.py
caugner: Sorry for the delay, this is now fixed in 12eb447f.

caugner:
I'm closing this PR now, as I won't be able to finish it. Anybody who would like to add SVN support to DXR in the future may use my work if they like.

PR #582, opened 2016-06-29
Test that plugins can handle non-ascii characters in paths.

pelmers:
Add a test with all plugins enabled that contains CJK non-ascii (but still unicode) characters in paths, to extend the previous non-utf8 path handling test.


tests/test_non_ascii_paths/code/makefile
```diff
@@ -0,0 +1,8 @@
+all: c
```
erikrose: (Commenting here because it won't let me comment on a binary file.) Did you mean to check "c" in?

tests/test_non_ascii_paths/code/makefile
```diff
@@ -0,0 +1,8 @@
...
+	mkdir code
+	cd code && python -c "import os; os.mkdir('\x8d\xae\xa2\xa0\xef')"
+	cd code && python -c "open('\x8d\xae\xa2\xa0\xef/\xaf\xa0\xaf\xaa\xa0', 'w').write('Hello!\n')"
+c: 第一.c
```
erikrose: I bet we're also going to test various versions of `make` pretty effectively. ;-)

tests/test_non_ascii_paths/code/makefile
pelmers: Nope, thanks for the catch; I forgot to add it on the clean command.

erikrose: And stick it in gitignore if git doesn't already ignore it.

erikrose:
So you've turned the non-UTF8 tests into merely non-ASCII ones. While this is nice for HFS+ people like myself, do we lose anything important?

pelmers:
We keep the non-UTF8 tests (non-utf8 is a subset of non-ascii)

erikrose:
sgtm! r+ when the tests pass.

PR #585, opened 2016-07-04
Add /dxr/build to .gitignore.

caugner:
When running `make test` after `make shell`, the directory */dxr/build/* is created which is not currently covered by *.gitignore* and therefore shows up as an untracked file on `git status`.


erikrose:
Good catch! I wasn't seeing this because I had "build" in my global ignore (which was a weird choice).

PR #589, opened 2016-07-05
Remove ES6-isms. Fixes bug 1283645.

erikrose:
The latest Safari (9.1.1) and pre-44 versions of FF can't handle them, so search was broken.


pelmers:
r+ !

PR #592, opened 2016-07-08
Update docker build trigger to point to new mozilla-org repo.

erikrose:
I can't tell which of the old encrypted values was the QUAY_AUTH_TOKEN, nor can I think of a way to test them without breaking master (I've lost track of the original GH_TOKEN, and encrypting the original QUAY_AUTH_TOKEN doesn't result in a matching string), so let's just append this new one and hope the last one wins.


erikrose:
@klibby r?

klibby:
r+!

PR #593, opened 2016-07-11
Raise max number of entries DXR will show when browsing a folder.

erikrose:
This fixes a problem when viewing the ridiculously large addons tree. ES makes us specify a ceiling, so we just raise it.


pelmers:
r+ when test pass!

PR #594, opened 2016-07-14
Properly index Python method calls, chained calls, and nested calls

### LINKED ISSUE #490: The python plugin is broken
Hello guys,

It seems the 348026a6bdea921e6f5e4e6ba0bf7c4103ea65a6 has broken the python plugin because it can no longer handle function calls, e.g. an indexing of the following example:
```python
import collections
Person = collections.namedtuple('Person', 'name age gender')
```
produces next output:
```
Traceback (most recent call last):
  File "/home/vagrant/venv/bin/dxr", line 9, in <module>
    load_entry_point('dxr==2.0', 'console_scripts', 'dxr')()
  File "/home/vagrant/dxr/dxr/cli/__init__.py", line 29, in main
    return dxr.invoke(ctx)
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/click/core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/click/core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "/home/vagrant/dxr/dxr/cli/index.py", line 26, in index
    index_and_deploy_tree(tree, verbose=verbose)
  File "/home/vagrant/dxr/dxr/build.py", line 64, in index_and_deploy_tree
    index_name = index_tree(tree, es, verbose=verbose)
  File "/home/vagrant/dxr/dxr/build.py", line 275, in index_tree
    index_files(tree, tree_indexers, index, pool, es)
  File "/home/vagrant/dxr/dxr/build.py", line 628, in index_files
    swallow_exc=False)
  File "/home/vagrant/dxr/dxr/build.py", line 581, in index_chunk
    index_file(tree, tree_indexers, path, es, index)
  File "/home/vagrant/dxr/dxr/build.py", line 482, in index_file
    refses.append(file_to_index.refs())
  File "/home/vagrant/dxr/dxr/plugins/python/indexers.py", line 216, in refs
    return self.visitor.refs
  File "/home/vagrant/dxr/dxr/plugins/python/indexers.py", line 195, in visitor
    self._visitor.visit(syntax_tree)
  File "/usr/lib/python2.7/ast.py", line 241, in visit
    return visitor(node)
  File "/usr/lib/python2.7/ast.py", line 249, in generic_visit
    self.visit(item)
  File "/usr/lib/python2.7/ast.py", line 241, in visit
    return visitor(node)
  File "/usr/lib/python2.7/ast.py", line 251, in generic_visit
    self.visit(value)
  File "/usr/lib/python2.7/ast.py", line 241, in visit
    return visitor(node)
  File "/home/vagrant/dxr/dxr/plugins/python/indexers.py", line 100, in visit_Call
    start, end = self.file_to_index.get_node_start_end(node)
  File "/home/vagrant/dxr/dxr/plugins/python/indexers.py", line 295, in get_node_start_end
    start, end = self.call_start_table[loc]
KeyError: (10, 9)
```

jbradberry:
Previous work (#487) broke the behavior of indexing calls that are
anything other than a plain function call.

fixes #490

### DIFF
--- a/dxr/plugins/python/indexers.py
+++ b/dxr/plugins/python/indexers.py
@@ -317,7 +317,7 @@ class FileToIndex(FileToIndexBase):
         loc = node.lineno, node.col_offset
         try:
             self.node_start_table[loc].pop()
-        except Exception:
+        except (KeyError, IndexError):
             pass
 
 
--- a/dxr/plugins/python/indexers.py
+++ b/dxr/plugins/python/indexers.py
@@ -260,10 +260,11 @@ class FileToIndex(FileToIndexBase):
                 # names, we start the queue for the current parenthesis level
                 # at the byte offset for the keyword token, but only start
                 # pushing character offsets once we're past the keyword.
-                paren_stack.setdefault(paren_level, (utf8_start, []))
+                queue_start, node_queue = paren_stack.setdefault(paren_level, (utf8_start, []))
+                node_start_table.setdefault(queue_start, node_queue)
 
                 if tok_name not in ('def', 'class'):
-                    paren_stack[paren_level][1].append((start, end))
+                    node_queue.append((start, end))
 
                 continue
 
@@ -276,22 +277,16 @@ class FileToIndex(FileToIndexBase):
                 if tok_name in '([{':
                     paren_level += 1
                 elif tok_name in '}])':
-                    node_start, node_queue = paren_stack.pop(paren_level, (None, None))
-                    if node_start is not None:
-                        node_start_table[node_start] = node_queue
+                    paren_stack.pop(paren_level, (None, None))
                     paren_level -= 1
                 elif tok_name == '.':
                     # Attribute access.  Don't reset, stay at the same level.
                     pass
                 else:
-                    node_start, node_queue = paren_stack.pop(paren_level, (None, None))
-                    if node_start is not None:
-                        node_start_table[node_start] = node_queue
+                    paren_stack.pop(paren_level, (None, None))
 
             elif tok_type == token.NEWLINE:
-                node_start, node_queue = paren_stack.pop(paren_level, (None, None))
-                if node_start is not None:
-                    node_start_table[node_start] = node_queue
+                paren_stack.pop(paren_level, (None, None))
 
         return node_start_table
 
--- a/dxr/plugins/python/indexers.py
+++ b/dxr/plugins/python/indexers.py
@@ -254,6 +254,32 @@ class FileToIndex(FileToIndexBase):
         # This table will contain, e.g. {(42, 4): [((42, 8), (42, 14)), ...], ...}
         node_start_table = {}
 
+        # However, if we have a sequence of attribute name tokens inside, say,
+        # method call parens or indexing brackets, the ast nodes for each of
+        # these will contain the location of the first token of the sequence
+        # inside, not the first of the outside sequence, e.g. for the line
+        #
+        #     a.b(d.e.f).c
+        #
+        # the nodes for 'd', 'e', and 'f' will each have the location of 'd',
+        # not 'a'.  Therefore, we need to push and pop the actual locations
+        # gleaned from the tokenizer onto a series of stacks that are aware of
+        # how many parens (or whichever) deep the token is.  So for the above
+        # example, the paren_stack would look like this by the time we process 'f'
+        #
+        #     {0: ((42, 4), [((42, 4), (42, 5)),       # node 'a'
+        #                    ((42, 6), (42, 7))]),     # node 'b'
+        #      1: ((42, 8), [((42, 8), (42, 9)),       # node 'd'
+        #                    ((42, 10), (42, 11)),     # node 'e'
+        #                    ((42, 12), (42, 13))])}   # node 'f'
+        #
+        # These lists that are being built up as the second item of each pair
+        # in the dict will then be popped out of the dict at the closing paren,
+        # bracket, or brace, and will have already been added to the
+        # longer-term node_start_table described above.  In this example, node
+        # 'c' would then be appended to the existing list at level 0, which
+        # would then terminate and be popped off in turn.
+
         paren_level, paren_stack = 0, {}
 
         for unicode_token, utf8_token in izip(token_gen, utf8_token_gen):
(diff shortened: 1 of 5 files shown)


dxr/plugins/python/indexers.py
```diff
@@ -295,14 +302,23 @@ def get_node_start_end(self, node):
...
-        return start, end
+    def advance_node(self, node):
+        """Destructively change which actual token offset we'll get on a call
+        to get_node_start_end.
+
+        """
+        loc = node.lineno, node.col_offset
+        try:
+            self.node_start_table[loc].pop()
+        except Exception:
```
pelmers: We should avoid except Exception, presumably it's one or both of KeyError and IndexError.

jbradberry: A fair point.

dxr/plugins/python/indexers.py
```diff
@@ -249,44 +257,43 @@ def analyze_tokens(self):
...
-                # but will be needed for recording variable references.
-                node_start_table[utf8_start] = (start, end)
-                node_type, node_start = 'name', (utf8_start, start)
+                # names, we start the queue for the current parenthesis level
+                # at the byte offset for the keyword token, but only start
+                # pushing character offsets once we're past the keyword.
+                paren_stack.setdefault(paren_level, (utf8_start, []))
+
+                if tok_name not in ('def', 'class'):
+                    paren_stack[paren_level][1].append((start, end))
```
erikrose: We could be a little more efficient (saving a lookup) by saving the result of the `setdefault` just above and then saying `saved_thingy[1].append(...)` here.

dxr/plugins/python/indexers.py
```diff
@@ -249,44 +257,43 @@ def analyze_tokens(self):
...
+                    # Attribute access.  Don't reset, stay at the same level.
+                    pass
+                else:
+                    node_start, node_queue = paren_stack.pop(paren_level, (None, None))
+                    if node_start is not None:
+                        node_start_table[node_start] = node_queue
 
-                node_type, node_start = None, None
+            elif tok_type == token.NEWLINE:
+                node_start, node_queue = paren_stack.pop(paren_level, (None, None))
```
erikrose: This is the 3rd time this is repeated. I wonder if there's a clever way to factor it out.

dxr/plugins/python/indexers.py
```diff
@@ -233,13 +241,13 @@ def analyze_tokens(self):
         utf8_token_gen = tokenize.generate_tokens(
             StringIO(self.contents.encode('utf-8')).readline)
 
-        # These are a mapping from the utf-8 byte starting points provided by
-        # the ast nodes, to the unicode character offset tuples for both the
-        # start and the end points.
+        # This is a mapping from the utf-8 byte starting points provided by
```
erikrose: I'm having trouble understanding what a unicode character offset tuple is. What are the elements of such a tuple? Just start and end? It sounds like start and end are each themselves tuples. I think an example would help me a lot.

erikrose: Also, since this table is what we return, it would be great to update the docstring, making it specific about what comes out of this function. Right now, it sounds like it does a lot more analysis than it does.

dxr/plugins/python/indexers.py
```diff
@@ -249,44 +257,43 @@ def analyze_tokens(self):
...
             elif tok_type == token.OP:
-                # In order to properly capture the start and end of function
-                # calls, we need to keep track of the parens.  Put the
-                # starting positions on a stack (here implemented with a dict
-                # so that it can be sparse), but only if the previous node was
-                # a name.
-                if tok_name == '(':
-                    if node_type == 'name':
-                        paren_stack[paren_level] = node_start
+                # Container delimiters (parens, brackets, and braces) start a
```
erikrose: Really, I don't see anything functionally wrong with the code. The tests are very encouraging; it looks like this now gets a bunch more stuff right! I'm just having trouble understanding it. In particular, I could use some explanatory comments about the purpose of paren_stack. Are nodes inside parens coming out of the tokenizer with positions relative to those parens or something? Once I can get that through my thick Friday-afternoon skull, we can merge this. Thanks so much for the brain-tremor-inducing-but-apparently-correct patch, Jeff! :-)

jbradberry:
@erikrose Ok, you're up again.

dxr/plugins/python/indexers.py
```diff
@@ -230,27 +230,35 @@ def refs(self):
...
+        # bytestring, which of course won't match up when we are indexing into a
+        # unicode string if there are multi-byte characters involved, so we need to
+        # provide a conversion table from utf-8 to unicode indexes.  Unfortunately,
+        # attribute ast nodes wind up with the same lineno and col_offset as the node
+        # they are attributes of, making this a bit tricky, so we need to group
+        # together offsets that are part of the same chain into a list.  This list
+        # can then be used to yield successive real offsets for the attributes as we
+        # get ast nodes that repeat the same lineno and col_offset.
+
+        # This table will contain, e.g. {(42, 4): [((42, 8), (42, 14)), ...], ...}
```
erikrose: Now I get it! Actually, your original comment makes sense now in light of the new one. The example helps enormously. Thanks!

dxr/plugins/python/indexers.py
```diff
@@ -276,22 +277,16 @@ def analyze_tokens(self):
                 if tok_name in '([{':
                     paren_level += 1
                 elif tok_name in '}])':
-                    node_start, node_queue = paren_stack.pop(paren_level, (None, None))
-                    if node_start is not None:
-                        node_start_table[node_start] = node_queue
+                    paren_stack.pop(paren_level, (None, None))
```
erikrose: It doesn't matter what the 2nd param is here, since we're throwing it away, right? I'd rather in that case have something clearly unused, like `'dummy'`. Or catch and ignore the exception.

dxr/plugins/python/indexers.py
erikrose: I'd love to have a similar example for `paren_stack`, whose structure and meaning are hard to suss out of the code.

dxr/plugins/python/indexers.py
jbradberry: Yep, I am a silly person and went through with a hastily applied hatchet.

dxr/plugins/python/indexers.py
```diff
@@ -260,10 +260,11 @@ def analyze_tokens(self):
                 # names, we start the queue for the current parenthesis level
                 # at the byte offset for the keyword token, but only start
                 # pushing character offsets once we're past the keyword.
-                paren_stack.setdefault(paren_level, (utf8_start, []))
+                queue_start, node_queue = paren_stack.setdefault(paren_level, (utf8_start, []))
+                node_start_table.setdefault(queue_start, node_queue)
```
erikrose: Do I deduce correctly that, if there's already a `node_start_table[queue_start]` key, we can count on it pointing to the same `node_queue` value, so it doesn't matter if this does nothing?

jbradberry: Yep.  queue_start is the index of the starting point some chain of attributes, e.g. `foo.bar.baz`, and the node queue needs to be a list of the starting locations of each of the attributes in that chain.  If we've already started capturing the indexes of such a chain, we want to keep on with the same list.

jbradberry:
@erikrose Sorry to take so long to get back to this.  I've added clarifying comments and an example for `paren_stack`, which I think is all that still needed to be done.

erikrose:
Lgtm! Thanks, Jeff!

PR #599, opened 2016-07-22
Upgrade clang to 3.6 to match production.

erikrose:


pelmers:
r+, I'm very happy to see this change.

PR #601, opened 2016-07-25
Stop using lazy regex in description plugin, add test.

pelmers:
Also anchor the comment at start of line.


pelmers:
The caveat here is that the regex stops looking at the comment at the first /, but / only ends a comment if preceded by a *.
Should I add some logic to the re.finditer() loop later on to consider this case?

erikrose:
I'm wondering if we can do better within the regex itself. Having a look now.

erikrose:
I've built on yours: #603. Now you get to review it. :-)

PR #604, opened 2016-07-26
Allow per-tree specification of worker count. Fixes bug 1288242.

erikrose:
This way, we can keep RAM use low for hungry plugins but still aggressively parallelize other trees.


erikrose:
@pelmers r?

pelmers:
r+!

PR #606, opened 2016-07-28
Don't ignore failures when running "node".

abbeyj:
The return value of post_build is ignored so the current code throws
away the exit code of the process.  Any failure is just ignored.
Using subprocess.check_call instead will throw an exception if
node exits with a non-zero status so we'll be sure to notice.

### DIFF
--- a/dxr/plugins/js/indexers.py
+++ b/dxr/plugins/js/indexers.py
@@ -34,12 +34,11 @@ class TreeToIndex(dxr.indexers.TreeToIndex):
     def post_build(self):
         # Execute the esprima to dump metadata, by running node from here and
         # passing in the tree location
-        retcode = subprocess.call(['node', 'analyze_tree.js',
-                                   self.tree.source_folder,
-                                   join(self.tree.temp_folder, 'plugins/js')] +
-                                   self.tree.ignore_filenames,
-                                  cwd=join(self.plugin_folder, 'analyze_js'))
-        return retcode
+        subprocess.check_call(['node', 'analyze_tree.js',
+                               self.tree.source_folder,
+                               join(self.tree.temp_folder, 'plugins/js')] +
+                               self.tree.ignore_filenames,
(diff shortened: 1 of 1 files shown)


erikrose:
@pelmers Is this going to blow up any of our trees? I remember talking about swallowed JS failures before; we sure get a lot of warnings.

dxr/plugins/js/indexers.py
```diff
@@ -34,12 +34,11 @@ def __init__(self, plugin_name, tree, vcs_cache):
     def post_build(self):
         # Execute the esprima to dump metadata, by running node from here and
         # passing in the tree location
-        retcode = subprocess.call(['node', 'analyze_tree.js',
-                                   self.tree.source_folder,
-                                   join(self.tree.temp_folder, 'plugins/js')] +
-                                   self.tree.ignore_filenames,
-                                  cwd=join(self.plugin_folder, 'analyze_js'))
-        return retcode
```
erikrose: Good catch, of course. The post_build return value is totally ignored.

pelmers:
Yes, many warnings appear but none of those are fatal. But I have seen esprima hit an out of memory error which crashes the node process. Currently we just continue as if nothing happened, and any .js files that we would have visited later will have no analysis.

erikrose:
So, obviously the best thing would be to solve the OOM problems, but that might be a major endeavor. (You're welcome to have a go, abbeyj. I suspect spinning up esprima separately for each file was prohibitively slow, but I'm not sure.) Failing that, we can at least, for now, stop returning a value that's ignored.

pelmers:
My hunch was that the reason for OOM was that analyze_tree.js maintains `dir_cache` on every file segment it encounters so it doesn't need to re-run all the ignore patterns every time, but this cache doesn't have an upper limit and kept growing, and the addons tree has a lot of files.

pelmers:
Stack trace of this happening: https://bugzilla.mozilla.org/show_bug.cgi?id=1280948#c12

PR #608, opened 2016-07-28
Repair broken click on table row.

pelmers:
Fixes bug 1290286.

It would be cleaner to move to a nicer markup that lets us just use <a> directly, but it looks like this fixes the bug by attaching the click listener to the container of the tree switch menu, which re-renders by nunjucks when a search occurs.


erikrose:
Thanks for the quick fix! :-D

PR #610, opened 2016-07-31
Fix bug 1289576, prevent direct-search crashes given a unicode search string

ertugrulkeremoglu:
`test_non_ascii_search_string` test files are copied from `test_basic`.

And the "hello world" example over in `code/main.c` is translated to Chinese for testing search giving unicode string.

Index the sample code and run `dxr` typing following commands:
```
cd tests/test_non_ascii_search_string
dxr index
dxr serve -a
```

And search the `世界!`

I hope that this is a welcome contribution. : )


ertugrulkeremoglu:
I realized that it was a bad idea to open a Pull Request from master branch. So, should I close this PR and create a new one from a sub branch or is it OK either way?

erikrose:
Don't worry about for my sake; I can merge from anything. But it will be confusing on your end, so you might want to clean up your local copy by.

tests/test_non_ascii_search_string/code/dot_c.c
```diff
@@ -0,0 +1 @@
+// Just a .c file to test find-by-extension
```
erikrose: Instead of copying all of test_basic, can you cut it down to the minimum required to test the fix?

ertugrulkeremoglu: Of course. I'll work on it. And thanks for the information. So let's close this PR. We meet in the next one. : )

PR #612, opened 2016-08-02
Fixes #1289576 - Direct-search crashes given a unicode search string

ertugrulkeremoglu:
This PR is relevant to #610 .

In addition:
- Test files are reduced to minimum required to test.


erikrose:
Note to self: remember to vouch https://mozillians.org/en-US/u/ertugrulkeremoglu/ when we're done.

erikrose:
It looks like you now have a test environment but no actual tests. Generally, there should be a test which fails before your bugfix and passes afterward.

ertugrulkeremoglu:
Sorry for delay.

I'm having issue testing the bugfix via `TestCase.search_response(query)` method. It always returns `200 OK` status, even before the changes. It doesn't return `500` status code.

Also I've tried a few more test methods but results are the same, though on browser it returns `500` before changes.

I think, it seems, there is a layer which encodes the query before searching. What should I/we do at the point?

erikrose:
This reproduces the stacktrace:

```python
# -*- coding: utf-8 -*-

from dxr.testing import SingleFileTestCase


class DirectSearchTests(SingleFileTestCase):
    source = u"""
        readmé
        """

    def test_unicode(self):
        """Typing some Unicode and hitting Return shouldn't crash."""
        self.direct_result_eq(u'readmé', 6)
```

tests/test_non_ascii_search_string/test_non_ascii_search_string.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: UTF-8 -*-
+
+from dxr.testing import SingleFileTestCase
+
+
+class NonAsciiSearchStringTests(SingleFileTestCase):
```
erikrose: Since you're no longer DxrInstanceTestCase, all those other files (the dxr.config and the whole code folder) are unnecessary. In fact, this testcase could even merge into `test_non_ascii_paths` (which could be renamed `test_unicode` or something).

erikrose:
I took the liberty of rebasing and removing the unnecessary 'c' binary. Thanks for working so hard on this patch!

PR #613, opened 2016-08-05
Run apt with DEBIAN_FRONTEND=noninteractive

na-g:
Sets the debconf frontend to noninteractive, silencing "debconf: unable to initialize frontend: FOO" errors emitted during dev container build.


erikrose:
Thanks for the patch! :-)

PR #617, opened 2016-09-08
Bug/1289597

NimaK:
Fixed errors/behavior when shift-clicking a line.  Changes made:

- Added a case for no previously selected line.
- Updated case shift-click on single previous selected line.
- Moved current clicked line actions into each case (it was causing the
toggle off case to be re-highlighted).


erikrose:
Looks good! Thanks so much for the fix!

PR #628, opened 2017-01-04
When pulling a file out of a VCS revision, don't assume its containin…

erikrose:
…g folder still exists. Fixes bug 1324026.

It may well have moved in the checkout's revision. So stop cd-ing to its containing folder. Instead, go into the deepest existing folder on its containing path, and run the VCS commands against the remainder of the relative path from there. Going into the deepest existing folder is a way to get the correct VCS invoked, even if there are nested repositories in the tree.

Refactor to create GenerativeTestCase, which should let us write these VCS test cases without tarring up an .hg or .git dir, which is hard to modify or examine and tempts us to couple too many tests together.

### DIFF
--- a/dxr/app.py
+++ b/dxr/app.py
@@ -276,8 +276,7 @@ def raw_rev(tree, revision, path):
 
     config = current_app.dxr_config
     tree_config = config.trees[tree]
-    abs_path = join(tree_config.source_folder, path)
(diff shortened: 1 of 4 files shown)


erikrose:
Landed in f3cc71a98975c99c34483fdc827489245b9387b9 with a docstring change.

PR #631, opened 2017-01-09
Don't treat vertical tabs or form feeds as line ending characters.

abbeyj:
This matches the behavior of clang and most text editors.

Otherwise our line number's don't match up with clang's and we end up trying to put links on the wrong lines.

### DIFF
--- a/dxr/utils.py
+++ b/dxr/utils.py
@@ -295,7 +295,16 @@ def split_content_lines(unicode):
     resulting lines.
 
     """
-    return unicode.splitlines(True)
+    lines = unicode.splitlines(True)
+    # Vertical Tabs and Form Feeds are treated as end-of-lines by splitlines
+    # but we don't want them to be.
+    def unsplit_some_lines(accum, x):
+        if accum and (accum[-1].endswith(u"\v") or accum[-1].endswith(u"\f")):
+            accum[-1] += x
+        else:
+            accum.append(x)
+        return accum
+    return reduce(unsplit_some_lines, lines, [])
 
 
 def unicode_for_display(str):
--- a/dxr/utils.py
+++ b/dxr/utils.py
@@ -297,8 +297,15 @@ def split_content_lines(unicode):
     """
     lines = unicode.splitlines(True)
     # Vertical Tabs, Form Feeds and some other characters are treated as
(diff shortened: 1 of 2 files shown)


dxr/utils.py
```diff
@@ -295,7 +295,16 @@ def split_content_lines(unicode):
     resulting lines.
 
     """
-    return unicode.splitlines(True)
+    lines = unicode.splitlines(True)
+    # Vertical Tabs and Form Feeds are treated as end-of-lines by splitlines
```
erikrose: Wow, that is *not* what the docs say. :-o

erikrose: Ah, correction: it's what the `unicode` docs say. `str.split()` actually does what we want.

erikrose: And \v and \f were added in Python 2.7.

erikrose: Now, there are a whole scad of other chars that it also considers linebreaks: \x1c, \x1d, \x85, etc.: https://docs.python.org/2.7/library/stdtypes.html#unicode.splitlines. Do you think we ought to handle all of them while we're at it? The bother and non-future-proofness of that makes me wonder whether we'd be better off encoding to utf-8, calling str.splitlines, and then re-decoding. I wonder what's fastest.

pelmers:
Will this pr fix [bug 1224516](https://bugzilla.mozilla.org/show_bug.cgi?id=1224516)?

abbeyj:
This does seem to be the same issue.  I see form feeds in that source file
as well.

dxr/utils.py
abbeyj: I ran the contents of Debugger.cpp (from the below linked bug) through the original version of my function from this PR as well as a version that does `[line.decode('utf-8') for line in unicode.encode('utf-8').splitlines(True)]`.  The original version takes ~5ms, the encode-and-decode version takes ~8ms.

I can also modify the original version to use `... in (u"\v", u"\f", u"\x1c", ...)` to cover all the oddball whitespace characters.  This actually runs just a bit faster than my original version at ~4.6ms.

Tell me what you'd prefer.

erikrose: Thanks for doing that! Let's go with the `in (u"\v", u"\f", u"\x1c", ...)` version, and easy choice since it both does the right thing and is fastest.

PR #637, opened 2017-02-09
Unregister ourselves as a DiagnosticsConsumer before we are destroyed.

abbeyj:
This prevents use-after-free bugs in clang 3.9.  It appears that in this new
version of clang, the DiagnosticsConsumer can be used even after the source
file has finished being processed and the ASTConsumer has been destroyed.

### DIFF
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -207,6 +207,14 @@ public:
     printPolicy.SuppressTagKeyword = true;
   }
 
+  ~IndexConsumer() {
+#if CLANG_AT_LEAST(3, 6)
+    ci.getDiagnostics().setClient(inner.release());
+#else
+    ci.getDiagnostics().setClient(inner);
+#endif
+  }
+
 #if CLANG_AT_LEAST(3, 3)
   // `clone` was removed from the DiagnosticConsumer interface in version 3.3,
   // so this can all be deleted once we're no longer supporting 3.2.


erikrose:
Fantastic! Just curious: how did you manage to track that down? I'm trying to build my C++ debugging arsenal.

abbeyj:
I ran clang under valgrind.  First I found the clang command that was being run (like `clang++ -Xclang -load -Xclang /.../libclang-index-plugin.so ...`) by looking in build.log.  Normally I would just run that directly under valgrind with `valgrind -- clang++ ...` but that wasn't working for me.  The clang++ binary acts as a "driver" and launches off another copy of clang to do the actual work of compiling.  When you run this under valgrind it breaks things somehow.  Maybe a bug in my version of valgrind.

Instead I got the command that the driver would normally run by adding the `-###` command line switch to the first clang command, resulting in `clang -### -Xclang ...`.  This then prints out the full command that would have been run which looks like `"clang-3.9" "-cc1" "-triple" ...`.  Then I ran that under valgrind with `valgrind -- "clang-3.9" "-cc1" "-triple" ...`.  This gave me an error report showing where the memory had been allocated, where it had been freed and where it was used incorrectly after being freed.

This would have probably been enough to figure out what was going on but I wanted to drop into the debugger to explore so I changed the command to `valgrind --vgdb=yes --vgdb-error=1 -- "clang-3.9" "-cc1" "-triple" ...`.  This will run until the first error occurs and then pause to let you attach a debugger.  Instructions for how to attach are printed on screen.  At this point I could see that it was trying to access something to do with Diagnostics and it had a pointer to our ASTConsumer which I could see had already been freed by the unique_ptr's destructor.  I just had to figure out how how it ended up with a dangling pointer which I found by looking through the dxr-index.cpp source for places where we might be handing a pointer to the Diagnostics engine.

If you need more details or a walkthrough, hit me up on IRC.

erikrose:
That's brilliantly helpful. Thanks for the walkthrough! :-D

PR #643, opened 2017-02-28
Fix anonymous namespaces in headers.

abbeyj:
Previously these namespaces would get a name assigned using the name of the
.cpp file that they were included from.  This caused problems when the header
was included in more than one file.  Searching for references to things
defined in such an anonymous namespace would show only results from a single
.cpp file and not all results.

The new behavior is to assign these namespaces a name based on the header
that they are defined in.  This allows the search to correctly find all references.

For an example, go to https://dxr.mozilla.org/mozilla-central/source/ipc/glue/ProtocolUtils.h#49
and try to find references for `CHANNEL_OPENED_MESSAGE_TYPE`.  The search
will not find any results and the search query shows
`(anonymous namespace in obj-x86_64-pc-linux-gnu/xpcom/threads/Unified_cpp_xpcom_threads1.cpp)`.
That .cpp file is one that includes this header but it doesn't have any references to
this identifier.  Effectively the search is limited to only that .cpp file.
There are actually several references to this identifier, for example:
https://dxr.mozilla.org/mozilla-central/source/ipc/glue/ProtocolUtils.cpp#77

I didn't properly account for the effect of having an anonymous namespace in
a header when I originally wrote this code.  This should do the correct thing now.

I separated the logic out into a helper function because I have a followup change
for static functions that's also going to need this.

### DIFF
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -300,6 +300,22 @@ public:
       recordValue("locend", locationToString(afterToken(endLoc)));
   }
 
+  // Given a declaration, get the name of the file containing the corresponding
+  // definition or the name of the file containing the declaration if no
+  // definition can be found.
+  std::string getRealFilenameForDefinition(const NamedDecl &d) {
+    const NamedDecl *decl = &d;
+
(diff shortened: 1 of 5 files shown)


erikrose:
lgtm!

PR #645, opened 2017-03-29
Pick the source directory specified last in the command line.

vkalintiris:


erikrose:
Thanks for the patch!

PR #648, opened 2017-05-22
Disabmiguate static functions that share the same name but appear in …

abbeyj:
…different files.

Before this change if you had a function like "static void foo()" in
one file and another, different function also called "static void foo()"
in a different file, DXR would not be able to entirely distinguish
between the two.  For instance, if you clicked on one of the
definitions and selected "Find references" then the list that was
produced would include references not only to the selected function but
also to the other function.

The way I mangled the name is to add "(static in filename.cpp)::" to the beginning.
This is similar to the "(anonymous namespace in filename.cpp)::" that we already
use for anonymous namespaces.  If somebody has a suggestion for something
better or different I'm more than willing to change this.

The `test_static` directory is mostly a copy of `test_anon_ns` but with static
functions instead of functions inside of anonymous namespaces.

### DIFF
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -395,6 +395,9 @@ public:
       const std::string &realname = getRealFilenameForDefinition(d);
       ret = "(" + ret.substr(1, anon_ns.size() - 2) + " in " + realname + ")" +
         ret.substr(anon_ns.size());
+    } else if (d.getLinkageInternal() == InternalLinkage) {
+      const std::string &realname = getRealFilenameForDefinition(d);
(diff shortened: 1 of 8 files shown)


erikrose:
Doubtless an improvement, but do I conclude correctly that we can still be ambiguous if we have several files of the same name, in different folders?

erikrose:
LGTM. It may change the behavior of some saved searches, but they were overbroad anyway, and it greatly improves the behavior of the context menus. Good show! Merge when ready.

abbeyj:
The filename that is used is the relative path from the source directory so even if we have two files with the same name, the identifiers generated here should still be unique.

erikrose:
Ah, perfect. Thanks!

PR #652, opened 2017-08-22
Modern es

scmikes:
This pull contains the first set of code to upgrade dxr to a modern elasticsearch.  The final target is elasticsearsearch 5.x latest.

This pull target is elasticsearch 2.4.1.  elasticsearch 2.4.1 still supports fieldnames as the default. 5.5.5 and later requires the create index command to specify filednames in the index create statement.

Main Changes:
   - index property name changes.    payload property appears in 2 places.  Under region and ref. Change names to region_payload and ref_payload.   Various py code changes to support this change.

 

dxr index and multiple test cases failed if an index with the same name as the old index tried to be created.

I am not certain if the desired behavior for dxr index , reusing index name is:  delete old index, create new.   I explicitly coded this in es.py.  Please review, if not correct, will fix to correct flow.

3 unit tests fail dealing with url redirect.   I am not sure what path to take to address these three unit tests.

added environment variable to define default elasticsearch host used by dxr DXR_DEFAULT_ELASTICSEARCH_HOST.  This the host to use if no host is defined in any config file or passed in on the command line.  This can be set in docker-compose.yml or on the command line.  The default value is 127.0.0.1.

Note: this code functions on both 1.7.5 (current docker elasticsearch version).   To test with elasticsearch 2.4.1 and this code
    1) setup elasticsearch 2.4.1 server
    2) edit docker-compose.yml   DXR_DEFUALT_ELASTICSEARCH_HOST to 2.4.1 server
    3) make shell
    4) make test

Note 2:   this code has only been validiated the following way so far.  Unit tests are the key validation so far.
     1) run unit tests - 3 fail with redirect LOCATION requrest property value missing
     2) dxr index    vim code base
     3) dxr index our products csv files

Remaining Targets:
    1) switch to scmikes/dxr   modern-es-555 branch to work on 555 work
             - define index create statement to activate fieldnames (in theory, should behave the same as 1.7.5 and 2.4.1 default behavior)
             - change docker env setup to use elasticsearch 5.5.5 (or later if 5.5.6 released).  Target is 5.x
     2) fix 3 unit test cases that fail
     3) very light validation of dxr serve.  This has not been validated at all so far.  Unit tests are key validation tests with the 2.4.1

### DIFF
--- a/dxr/cli/serve.py
+++ b/dxr/cli/serve.py
@@ -1,8 +1,10 @@
 from click import command, option
 
 from dxr.app import make_app
+from dxr.config import DXR_DEFAULT_ELASTICSEARCH_HOST
 from dxr.cli.utils import config_option
 
+#MLS FIXME change localhost default for test cases
 
 @command()
 @config_option
@@ -12,7 +14,7 @@ from dxr.cli.utils import config_option
         flag_value='0.0.0.0',
         help='Serve on all interfaces.  Equivalent to --host 0.0.0.0')
 @option('--host', '-h',
-        default='localhost',
+        default=DXR_DEFAULT_ELASTICSEARCH_HOST,
         show_default=True,
         help='The host address to serve on')
 @option('--workers', '-w',
--- a/dxr/cli/serve.py
+++ b/dxr/cli/serve.py
@@ -1,7 +1,6 @@
 from click import command, option
 
 from dxr.app import make_app
-from dxr.config import DXR_DEFAULT_ELASTICSEARCH_HOST
 from dxr.cli.utils import config_option
 
 
@@ -13,7 +12,7 @@ from dxr.cli.utils import config_option
         flag_value='0.0.0.0',
         help='Serve on all interfaces.  Equivalent to --host 0.0.0.0')
 @option('--host', '-h',
-        default=DXR_DEFAULT_ELASTICSEARCH_HOST,
+        default='localhost',
         show_default=True,
         help='The host address to serve on')
 @option('--workers', '-w',
--- a/dxr/cli/serve.py
+++ b/dxr/cli/serve.py
@@ -4,7 +4,6 @@ from dxr.app import make_app
 from dxr.config import DXR_DEFAULT_ELASTICSEARCH_HOST
 from dxr.cli.utils import config_option
 
-#MLS FIXME change localhost default for test cases
 
 @command()
 @config_option
--- a/dxr/config.py
+++ b/dxr/config.py
@@ -6,8 +6,9 @@ Please update docs/source/configuration.rst when you change this.
 from datetime import datetime
 from multiprocessing import cpu_count
 from ordereddict import OrderedDict
-from os import getcwd
+from os import getcwd, getenv
 from os.path import abspath
+
 
 from configobj import ConfigObj
 from funcy import merge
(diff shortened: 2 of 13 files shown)


dxr/lines.py
```diff
@@ -120,8 +120,12 @@ def ref_class(plugin, id):
                 warn('Ref subclass from plugin %s with ID %s was referenced '
                      'in the index but not found in the current '
                      'implementation. Ignored.' % (plugin, id))
-
-        payload = es_data['payload']
+
```
erikrose: Let's fix [linter errors like this one](https://travis-ci.org/mozilla/dxr/builds/267092335#L2659-L2666) as we go so Travis will at least try to run the tests.

dxr/config.py
```diff
@@ -19,6 +20,11 @@
 from dxr.plugins import all_plugins_but_core, core_plugin
 from dxr.utils import cd, if_raises
 
+#setup default elasticsearch host to use if not specifice in any config file
```
erikrose: "Set up" and "specified"

dxr/cli/serve.py
```diff
@@ -12,7 +13,7 @@
         flag_value='0.0.0.0',
         help='Serve on all interfaces.  Equivalent to --host 0.0.0.0')
 @option('--host', '-h',
-        default='localhost',
+        default=DXR_DEFAULT_ELASTICSEARCH_HOST,
```
erikrose: This is supposed to be the address the web serve itself listens on, not the ES host.

dxr/config.py
```diff
@@ -19,6 +20,11 @@
...
+DXR_DEFAULT_ELASTICSEARCH_HOST = getenv('DXR_DEFAULT_ELASTICSEARCH_HOST', '127.0.0.1')
```
erikrose: If we don't end up using this in serve.py, then we can collapse it into the next line.

dxr/es.py
```diff
@@ -107,10 +107,18 @@ def filtered_query_hits(index, doc_type, filter, sort=None, size=1, include=None
...
+    
+        es.create_index(index, settings=settings)
+    except ElasticHttpError:
+        es.delete_index(index)
+        es.create_index(index, settings=settings)
+        
     es.health(index=index,
               wait_for_status='yellow',
-              wait_for_relocating_shards=0,  # wait for all
+#              wait_for_no_relocating_shards='true',  # wait for all
```
erikrose: Did you mean to leave this commented out?

dxr/plugins/core.py
```diff
@@ -156,11 +154,10 @@
             },
 
             'refs': {
-                'type': 'object',
+                'properties': {
                 'start': UNINDEXED_INT,
```
erikrose: We should indent these lines inside the `properties` key.

dxr/plugins/core.py
```diff
@@ -174,17 +171,18 @@
                         'qualname_hash': UNINDEXED_LONG
                     }
                 }
+                }
             },
 
             'regions': {
-                'type': 'object',
+                'properties':{
                 'start': UNINDEXED_INT,
```
erikrose: These too.

erikrose (changes requested):
I just see some minor touch-ups to address so you can get this merged before embarking on the 5.5 stuff. Thanks again for your good work!

scmikes:
Thanks Erik,

I will fix that and push that up.  I am away from my dev machine today.  I should be able to fix and push up tomorrow.

Michael Schneider
Sr Software Engineering Consultant
Tech & Innov

Siemens Industry Sector
Siemens Product Lifecycle Management Software Inc.
2000 Eastman Drive
Milford, OH  45150 United States
Tel.      :+1 (513) 576-2935
Fax      :+1 (513) 576-2734
michael.l.schneider@siemens.com <michael.l.schneider@siemens.com%20>
www.siemens.com/plm

erikrose:
Looks good! There's just one missed spot in core.py where we should indent. Search for my "These too" comment. Thanks again!

scmikes:
Erik,



Will submit with next round of submits.  Working on 5.5.5.



Mike


----------------------------------------------------------
"The best performance improvement is the transition from the nonworking state to the working state." - John Ousterhout<http://www.answers.com/topic/john-ousterhout>

Michael Schneider - Senior Software Engineering Consultant
Siemens  PLM Software
________________________________

PR #653, opened 2017-08-23
Modern es 241

scmikes:
Erik,

Thanks for the code review.  I did not  know flakes8 was part of the process, I ran that and cleand up errors.  I had to undo one change because the py code was test case data :-)

1) server.py   very good catch on using elasticsearch default host for dxr web server.  This is fixed. Thank you, sorry I missed this.   

2) config.py fixed typos in comment
    Note: env var for default elastisearch host is used in testing, so even though it is dropped from server.py it is still used

3) es.py
     #wait_for_no_reloacing_shards='true' is commented out.   This is a 5.5.5 thing, but the pyelasticsearch.py lib did not support the new 5.5.5 option.  Not sure what you want to do about that.  The option "should" there in 5.5.5, but it is not available in pyelasticsearch.py

4) core.py  - corrected indentation in properties attributes

Note: The code for 2.4.1 should run in both 1.7.5 and 2.4.1.   I use an external 2.4.1 server for def and use the env var.   The current unit test code uses the 1.7.5 docker config, and nothing extra should be needed for unit tests.

3 unit tests are failing, and I am not sure which direction to take.   I was going to complete the 5.5.5 work before working the unit tests.   If you see a fix for the unit tests, please let me know I will apply fixes ASPAP.

Also I am happy to make any further changes to comply with any coding standards, just let me know.  My pyflake8 is slighly newer, so I fixed a few more spacing errors in old code.  That is part of these changes.

The unit tests should run and fail with 3 errors.   The Location attribute of the response dict is not set.  I added code to the test case to dump all dict elements to stdout out.  Please let me know if you see a direction to address these errors.

### DIFF
--- a/dxr/build.py
+++ b/dxr/build.py
@@ -524,7 +524,7 @@ def index_file(tree, tree_indexers, path, es, index):
                 # request time we want to be able to merge them individually
                 # with those from skimmers.
                 refs_and_regions = bucket(tags, lambda index_obj: "regions" if
-                                          isinstance(index_obj['payload'], basestring) else
+                                          isinstance(index_obj['region_payload'], basestring) else
                                           "refs")
                 if 'refs' in refs_and_regions:
                     total['refs'] = refs_and_regions['refs']
--- a/dxr/cli/serve.py
+++ b/dxr/cli/serve.py
@@ -1,8 +1,10 @@
 from click import command, option
 
 from dxr.app import make_app
+from dxr.config import DXR_DEFAULT_ELASTICSEARCH_HOST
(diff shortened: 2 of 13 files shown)


erikrose:
Thanks for the quick response! Say, if you merge your fixes to your existing modern-es branch and push that up to your GitHub repo, the workflow will be much better: it will just continue on with the pull request #652, and it'll automatically collapse my comments as they are addressed. I'll close this PR in the meantime.

erikrose:
Annnnnd I see you already did that. :-) Thanks!

PR #654, opened 2017-11-14
Modern es 555

scmikes:
current status of work porting dxr to elasticsearch 555 and later


scmikes:
Many tests fail.  I was able to index vim code base, but dxr server queries fail.   I am going to pause development work for a bit.

erikrose:
Thanks! It's good to have them here for safekeeping, in any case.

PR #656, opened 2018-01-23
Add support for clang 5.0.

abbeyj:


dxr/plugins/clang/dxr-index.cpp
```diff
@@ -1299,9 +1304,16 @@ void PreprocThunk::MacroDefined(const Token &tok, const MacroDirective *md) {
                                   SourceRange range, const MacroArgs *ma) {
     real->MacroExpands(tok, md.getMacroInfo(), range);
   }
+    #if CLANG_AT_LEAST(5, 0)
+  void PreprocThunk::MacroUndefined(const Token &tok, const MacroDefinition &md,
+                                    const MacroDirective */*Undef*/) {
```
erikrose: What does Undef mean? Is it a special pragma or something, or am I just missing your drift?

abbeyj: It is the name of that parameter: https://clang.llvm.org/doxygen/classclang_1_1PPCallbacks.html#a83ec9e0187f46a037f16f00450c09367

We're not using that parameter in this function so to avoid an unused variable warning I'm not giving it a name.  Instead I commented the name out so people will have some idea of what it is for and can easily restore the name if needed.

I don't know why GitHub highlights this so strangely.  I could add a space before the comment if you wanted.  Or take the comment out entirely.

erikrose: Strange. I actually downloaded it and looked at it, because I thought the strange highlighting meant a control char or something got in. I'm fine with it; push away!

PR #657, opened 2018-01-23
Fix the context menu not appearing when clicking on a call to operator()

abbeyj:
We treat the call as covering just the token `(`.  Since this has no letters or numbers the code decided that a substring search was not appropriate and then didn't display the context menu at all.  But items like "Jump to definition" and "Find references" are still appropriate and should be shown.

This is on top of #656 since I can no longer build without that patch.

### DIFF
--- a/dxr/static_unhashed/js/context_menu.js
+++ b/dxr/static_unhashed/js/context_menu.js
@@ -145,18 +145,16 @@ $(function() {
 
             word = selectedTxtString.substr(startIndex, endIndex - startIndex);
 
-            // No word was clicked on, nothing to search.
-            if (word === '') {
-                return;
-            }
+            var menuItems = [];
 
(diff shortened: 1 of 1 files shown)


dxr/static_unhashed/js/context_menu.js
```diff
@@ -168,7 +166,12 @@ $(function() {
                 menuItems = menuItems.concat(currentNodeData);
             }
 
-            contextMenu.menuItems = menuItems;
+            if (menuItems.length === 0) {
+                // Nothing to display.
+                return;
+            }
+
+            var contextMenu = { menuItems: menuItems };
```
erikrose: LGTM. If you want, you could even merge this into the next line; I don't really see a clarity win for this one-use `contextMenu` symbol given that it's being passed to `setContextMenu()`.

PR #659, opened 2018-02-07
Add support for references to overloaded operators.

abbeyj:
clang doesn't provide an end point for these but since they are a single token
we can use the start point as the end point.


erikrose:
I don't know this deals with the multi-char operators like `++`, but your tests prove it does, so lgtm!

PR #660, opened 2018-02-07
Fix user-defined literals.

abbeyj:
C++ introduced a new thing called "user-defined literals".  These have double
quotes in their name.  This was causing problems when writing out the
temporary .csv files.  If there is a double quote in a CSV field then the
quote should be escaped by doubling it.  There was already code to do this
but it was only enabled for certain fields.  Identifiers were assumed to
never contain quotes and thus not need escaping.

Since using the proper quoting algorithm will always produce the correct
result for any type of field, turn it on unconditionally for all fields.

I'm not sure if the query syntax is what we want to use here or not.  Having unescaped literal double quotes inside of a double-quoted string feels odd.  I was a bit surprised that it worked as-is.
Is this guaranteed to do the right thing or is it just a quirk of the way the query parser works?

### DIFF
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -419,17 +419,15 @@ public:
     *out << name;
   }
 
-  void recordValue(const char *key, std::string value, bool needQuotes=false) {
+  void recordValue(const char *key, std::string value) {
     *out << "," << key << ",\"";
     int start = 0;
-    if (needQuotes) {
-      int quote = value.find('"');
-      while (quote != -1) {
-        // Need to repeat the "
-        *out << value.substr(start, quote - start + 1) << "\"";
-        start = quote + 1;
-        quote = value.find('"', start);
-      }
+    int quote = value.find('"');
+    while (quote != -1) {
+      // Need to repeat the "
+      *out << value.substr(start, quote - start + 1) << "\"";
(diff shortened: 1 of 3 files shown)


dxr/plugins/clang/dxr-index.cpp
```diff
@@ -419,17 +419,15 @@ class IndexConsumer : public ASTConsumer,
     *out << name;
   }
 
-  void recordValue(const char *key, std::string value, bool needQuotes=false) {
+  void recordValue(const char *key, std::string value) {
```
erikrose: This also cuts down on anonymous `true`s at call sites. I like it.

dxr/plugins/clang/tests/test_user_defined_literals.py
```diff
@@ -0,0 +1,23 @@
...
+        int operator "" _i(const char *) { return 0; }
+
+        int main() {
+          42_i;
+          return 0;
+        }
+        """
+
+    def test_def(self):
+        self.found_line_eq('+function:"operator""_i(const char *)"',
```
erikrose: Yes, this is fine. It turns out I even wrote out my rationale: https://github.com/mozilla/dxr/blob/c7f124a0cf1c68c7b7a8941c33b66a4883154f9e/dxr/query.py#L255-L263

PR #661, opened 2018-02-07
Don't crash when an include file does not exist.

abbeyj:
In this case `file` will be `NULL`.  We have a check for that but we
were dereferencing `file` before the check.

### DIFF
--- a/dxr/plugins/clang/dxr-index.cpp
+++ b/dxr/plugins/clang/dxr-index.cpp
@@ -1257,29 +1257,32 @@ public:
       StringRef relativePath,
       const Module *imported) {
     PresumedLoc presumedHashLoc = sm.getPresumedLoc(hashLoc);
-    const FileInfoPtr &source = getFileInfo(presumedHashLoc.getFilename());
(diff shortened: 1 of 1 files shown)


dxr/plugins/clang/dxr-index.cpp
```diff
@@ -1259,29 +1259,32 @@ class IndexConsumer : public ASTConsumer,
...
-        // you include a nonexistent file.
-        !file ||
+    // Don't record inclusions of files that are outside the source tree,
+    // like stdlibs. file is NULL if an #include can't be resolved, like if
+    // you include a nonexistent file.
+    if (!file)
+      return;
 
-        !(target->interesting) ||
+    const FileInfoPtr &source = getFileInfo(presumedHashLoc.getFilename());
```
erikrose: I guess we could even move this down a few more lines to nearer where it's used.

erikrose:
LGTM. Do what you like with my comment; perhaps you like having source and target defined near each other for clarity.

PR #662, opened 2018-02-08
rewrite docker to reduce layer usage and batch run tasks

theonlydoo:


erikrose:
I think we'd like to keep the logical separation of files we have right now, and I want to preserve the comments. But I'd entertain a version of this that mashed more of the commands within the files together with `&&`. Closing for now, but please reopen if you're interested in that. Thanks!

PR #663, opened 2018-02-09
Don't clean the object folder if the user has explicitly requested no…

abbeyj:
… cleaning.


erikrose:
Seems reasonable. r+.

PR #665, opened 2018-02-11
Allow searching for references to things defined in third-party code

abbeyj:
This is split into 3 commits for easier review.

Until now functions, classes and variables that were not defined within the source tree were "invisible".  Uses of these things were not hyperlinked and searching for references to them found no results.

With this change we no longer ignore these things and they become searchable just like other identifiers.  You won't be able to find the definition for such items but other searches like "Find references" will work.

This works for things defined in third-party libraries and even libc.  So now you can search for all uses of "strcpy" or "gets" in your code.

### DIFF
--- a/dxr/plugins/clang/tests/__init__.py
+++ b/dxr/plugins/clang/tests/__init__.py
@@ -9,7 +9,7 @@ class CSingleFileTestCase(SingleFileTestCase):
     def config_input(cls, config_dir_path):
         input = super(CSingleFileTestCase, cls).config_input(config_dir_path)
         input['DXR']['enabled_plugins'] = 'pygmentize clang'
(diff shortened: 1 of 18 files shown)


dxr/plugins/clang/tests/__init__.py
```diff
@@ -12,11 +12,3 @@ def config_input(cls, config_dir_path):
         input['code']['build_command'] = '$CXX %s -c main.cpp' % cls.cflags
         return input
 
-
-# Tests that don't otherwise need a main() can append this one just to get
-# their code to compile:
-MINIMAL_MAIN = """
```
erikrose: Wow, nice simplification!

erikrose:
Looks great. Merge when ready!

PR #666, opened 2018-02-13
Run tests in parallel for a nice speedup

abbeyj:
Lucky #666

This ended up bigger than I originally anticipated.  If you want me to split it up into smaller pieces or something, please tell me.

I get impatient waiting for tests to run.  Many of them could run in parallel except for a few problems:
- All the tests use the same name for the ES index so if you run more than one at a time they interfere with each other.  We can fix this by giving each test a unique index name.
- Tests were inadvertently writing temp files into a `dxr-temp-code` directory in the root of the DXR directory.  One test would delete this when it was done running while another was still using it.  Making sure the temporary directories are created relative to the `dxr.config` file as originally intended helps.
- But there's still a problem when there are multiple `DxrInstanceTestCase` tests in the same directory since they will still all try to use the same temporary directory.  We can fix this by copying the entire test to a unique temporary directory in the setup for the class.  This actually simplifies some things since it makes `DxrInstanceTestCase` and `GenerativeTestCase` both run from a temporary directory.

Other speedups include:
- Performing the indexing in-process instead of shelling out to `dxr index`.
- Not having to call `dxr clean` at all since deleting the temporary directory will clean up everything.

Before:
Ran 550 tests in 116.558s

After:
Ran 550 tests in 44.232s

Tests run over 2.5x faster.


erikrose:
Sorry for the wait; this is a big one, as you said. Taking a swat at it now!

dxr/config.py
```diff
@@ -182,6 +184,9 @@ def __init__(self, input, relative_to=None):
                     except SchemaError as exc:
                         raise ConfigError(exc.code, [section])
 
+        self._section['es_catalog_index'] = self._section['es_catalog_index'].format(
+            config_path_hash=self.path_hash())
```
erikrose: And since you didn't change the default values of the es_alias or es_catalog_index options, we should be able to deploy this without breaking anything. Do you agree? (Also .format() doesn't care if you give it extra args for which there aren't tokens in the string, so that's good too.)

dxr/build.py
```diff
@@ -212,7 +214,8 @@ def delete_index_quietly(es, index):
             # Substitute the format, tree name, and uuid into the index identifier.
             index = tree.es_index.format(format=FORMAT,
                                          tree=tree.name,
-                                         unique=uuid1())
+                                         unique=uuid1(),
+                                         config_path_hash=config.path_hash())
```
erikrose: We shouldn't need this for uniqueness, since "a random 14-bit sequence number" is part of uuid1() when called with no params. Is it here to help humans correlate an alias and its corresponding index?

dxr/config.py
```diff
@@ -153,6 +154,7 @@ def __init__(self, input, relative_to=None):
 
         if not relative_to:
             relative_to = getcwd()
+        self.path = relative_to
```
erikrose: So we're assuming only 1 DXR config file per dir, then, since we don't include the file name (since we don't have access to it)?

dxr/plugins/rust/tests/test_cargo/dxr.config
```diff
@@ -1,8 +1,8 @@
 [DXR]
 enabled_plugins     = pygmentize rust
-es_index            = dxr_{format}_{tree}_{unique}
-es_alias            = dxr_{format}_{tree}
-es_catalog_index    = dxr_catalog
+es_index            = dxr_test_{format}_{tree}_{unique}
```
erikrose: Good catch on all of these!

dxr/testing.py
```diff
@@ -342,12 +334,12 @@ class DxrInstanceTestCaseMakeFirst(DxrInstanceTestCase):
...
     @classmethod
     def generate(cls):
-        check_call(['make'], cwd=join(cls.this_dir(), 'code'))
         super(DxrInstanceTestCaseMakeFirst, cls).generate()
+        check_call(['make'], cwd=join(cls._config_dir_path, 'code'))
 
     @classmethod
     def teardown_class(cls):
-        check_call(['make', 'clean'], cwd=join(cls.this_dir(), 'code'))
+        check_call(['make', 'clean'], cwd=join(cls._config_dir_path, 'code'))
```
erikrose: I guess it doesn't hurt to keep calling `make clean` for now. Who knows what people do in those makefiles? They could be ickily doing something outside the test dir.

erikrose:
Looks great to me. Have a quick read through my comments to be sure I didn't bring up anything you didn't think of. I don't think there's any need to change the [format version](http://dxr.readthedocs.io/en/latest/glossary.html?highlight=format) either. Great work!

dxr/config.py
abbeyj: Yes, it was intended to be entirely backward compatible.  The change should be transparent for existing users.

dxr/build.py
abbeyj: The problem with the UUID is that we don't know what UUID is going to be generated in advance and I don't have a good way to get the value actually generated here all the way to `_delete_es_indices`.  Guaranteeing that the index and the catalog both start with the same prefix and that this prefix depends only on the test being run make them easy to clean up with a wildcard.  If you can suggest a better way to deal with this, I'll be happy to adjust it.

dxr/testing.py
abbeyj: Yes.  I just didn't really see any need to change this.  It shouldn't cause any problem to keep it and if there is really some external cleanup that needs to be done we'll still perform it.

PR #667, opened 2018-02-15
Handle files with CRLF line endings.

abbeyj:
Pygments gets confused by these and does odd things like splitting up a
CRLF pair so that the CR is on the end of one token and the LF is a
separate token.

This leads to us eventually outputting HTML that renders as two lines
in the browser for every one line in the source file.

We can fix up the output from pygments so that tokens do not contain
these unexpected line ending characters.  This guarantees that we
never output a <span> tag containing a CR or an LF.


dxr/plugins/tests/test_pygmentize_crlf.py
```diff
@@ -0,0 +1,28 @@
+from nose.tools import ok_
```
erikrose: A nit: we should probably turn pygmentize into a package (folder) and put the tests inside it. See `plugins/descriptor` for a precedent.

PR #669, opened 2018-04-17
Remove deprecated peep.py and use `pip install --require-hashes` inst…

abbeyj:
…ead.


dxr/cli/deploy.py
```diff
@@ -190,7 +190,7 @@ def build(self, rev):
                     # Does this recurse into submodules?
                     run('git fsck --no-dangling')
 
-                    # Install stuff, using the new copy of peep from the checkout:
+                    # Install stuff, using the copy of pip from the checkout:
```
erikrose: pip isn't from the checkout.

makefile
```diff
@@ -120,8 +120,9 @@ dxr/static_unhashed/js/templates.js: dxr/templates/nunjucks/*.html \
 	touch $@
 
 # Install requirements in current virtualenv:
-.peep_installed: $(VIRTUAL_ENV)/bin/activate requirements.txt
-	$(VIRTUAL_ENV)/bin/python tooling/peep.py install -r requirements.txt
+.requirements_installed: $(VIRTUAL_ENV)/bin/activate requirements.txt
+	$(VIRTUAL_ENV)/bin/pip install 'pip>=8.0.0'  # need 8.0 or better for --require-hashes
```
erikrose: I don't like the unauthenticated installation of pip here. I would prefer to either (1) just require pip >= 8 to pre-exist on the machine (my favorite, given that pip 8 came out over 2 years ago) or (2) bundle and use [pipstrap](https://github.com/erikrose/pipstrap). I do need to make sure we have pip 8 on our own servers.

abbeyj: pip 8 is not present in the environment on Travis (it has 1.5.4).  The CI run failed the first time I tried to do this PR (without this line present).  Is there a way to fix things on Travis?  If so, I'll happily remove this line.

erikrose: (Continued on IRC)

abbeyj: I couldn't figure out how to get a newer version of pip installed in the Docker Ubuntu 14.04 image.  I went with the pipstrap approach instead.  This does have the advantage that it should work for anybody, even if they're not using Docker.

PR #673, opened 2018-06-29
Fix missing "Jump to definition" link on function declarations.

abbeyj:
I believe I broke this in 94882fe.  There is now, in some circumstances,
the possibility of having multiple decldefs for a single function.  Only one
will have a defloc and the rest will be without it.

When we're building the page we sort them (in essentially random order) and
pick the first one.  If that one happens to be one without a defloc then the
"Jump to definition" menu item will not be shown.

The fix here makes sure that if we have a decldef with a defloc then it is
chosen in preference to others without a defloc.  If all the decldefs that
we have are without deflocs then the one that we pick is essentially chosen at
random like before.  But in this case all the decldefs should be identical so
this should not cause any issues.


dxr/plugins/clang/tests/test_func_decl_jump/code/shared.h
```diff
@@ -0,0 +1,26 @@
...
+ * When compiling second.cpp the output for this file will look like:
+ *     decldef,name,"function_in_main",...
+ *     decldef,name,"function_in_second",...,defloc,"second.cpp:3:5",...
+ * (where the "function_in_main" line has no defloc).
+ *
+ * We can't count on which of these two files will be processed first.  If we
+ * process main.cpp first then function_in_second will serve as our test
+ * function.  If we process second.cpp first then function_in_main will serve
+ * as our test case.  No matter which order we process the files in we are
+ * guaranteed to have one function that will work as an effective test.
```
erikrose: Clever!

dxr/plugins/clang/menus.py
```diff
@@ -205,10 +205,15 @@ def from_condensed(cls, tree, prop):
...
-                    'has_overriddens' in prop, 'has_overrides' in prop),
-                   qualname=prop['qualname'])
+        ret = cls(tree,
+                  (search_for_def, prop['qualname'],
+                   'has_overriddens' in prop, 'has_overrides' in prop),
+                  qualname=prop['qualname'])
+        if not search_for_def:
+            # Prefer FunctionRefs with definitions over FunctionRefs without
+            # definitions.
+            ret.sort_order += 0.1
```
erikrose: A nice quick fix. We just need to keep in mind (or in a comment somewhere conspicuous) that we're essentially reserving the tenths place for this purpose, lest somebody else start adding or subtracting also and nullify this intermittently.

Issue #688, opened 2019-08-05
Installation fails due to conflicting click version


NeolithEra:
Hi, users are unable to run dxr due to dependency conflict with _**click**_ package. As shown in the following full dependency graph of dxr, dxr requires _**click>=3.3,<5.0**_，while **_Flask==1.1.1_** requires _**click>=5.1**_.

According to pip’s “first found wins” installation strategy, _**click==4.1*_ is the actually installed version. However, _**click==4.1**_ does not satisfy _**click>=5.1**_.

### Dependency tree
```
dxr-master
| +-click(version range:>=3.3,<5.0)
| +-configobj(version range:>=5.0.6)
| | +-six(version range:*)
| +-Flask(version range:>=0.9)
| | +-click(version range:>=5.1)
| | +-itsdangerous(version range:>=0.24)
| | +-Jinja2(version range:>=2.10.1)
| | | +-MarkupSafe(version range:>=0.23)
| | +-Werkzeug(version range:>=0.15)
| +-funcy(version range:>=1.0,<2.0)
| +-futures(version range:>=2.1.1,<4.0)
| +-Jinja2(version range:>=2.6,<3.0)
| | +-MarkupSafe(version range:>=0.23)
| +-Mercurial(version range:>=3.4)
| +-more-itertools(version range:>=2.0,<3.0)
| +-ordereddict(version range:>=1.1,<2.0)
| +-parsimonious(version range:>=0.6,<0.7)
| | +-six(version range:>=1.9.0)
| +-pyelasticsearch(version range:>=1.1,<2.0)
| | +-certifi(version range:*)
| | +-elasticsearch(version range:>=1.3.0,<2.0.0)
| | | +-urllib3(version range:>=1.21.1)
| | +-simplejson(version range:>=3.0)
| | +-six(version range:<2.0,>=1.4.0)
| | +-urllib3(version range:>=1.8,<2.0)
| +-Pygments(version range:>=2.1.3,<3.0)
| | +-coverage(version range:*)
| | +-nose(version range:*)
| | +-pyflakes(version range:*)
| | +-pylint(version range:*)
| | +-tox(version range:*)
| +-python-hglib(version range:>=1.7,<3.0)
| +-requests(version range:>=1.0,<2.0)
| | +-certifi(version range:>=2017.4.17)
| | +-chardet(version range:<3.1.0,>=3.0.2)
| | +-idna(version range:>=2.5,<2.9)
[...]

PR #689, opened 2019-08-09
Fix dependency conflict for issue

NeolithEra:
#688


erikrose:
Looks like Travis somehow got disabled. Should be fixed now. Can you push something else to it to kick off a run of the tests?

abbeyj:
I think #690 will fix the rust test failures seen here.  If you rebase on top of that commit or merge that commit here then I think you'll be all set.

PR #690, opened 2019-08-18
Update for new version of rustup

abbeyj:
The new version is hosted at a different location, takes different
command line arguments, and installs to the user's home directory
instead of /usr/local.

I'm hopeful that this will fix the CI failure seen in #689


erikrose:
lgtm!

